PR #165, opened 2015-05-26
Fixed URL for FormattingConventions.doc

amazingant:
The given URL is dead, changed it to a wayback machine URL for the file.


dungpa:
The URL is dead even before I wrote the doc. Don't you want it to Rest In Peace :-)?

Thanks for the PR.

amazingant:
I don't mind one way or the other; it can stay dead if it wants to. :stuck_out_tongue_winking_eye:

Someone else was giving me "slack" on the F# slack channel about the dead URL and some of the formatting suggestions that they didn't like. Figured I'd fix the URL since it's not hard to do.

PR #166, opened 2015-06-08
Add minimal FAKE build script

dungpa:
Tasks:

 - [x] Auto-update versions
 - [x] Publish NuGet from script
 - [x] Support single-command build


dungpa:
Ok, it's much less painful to work on the code base now.

The build script is a bit convoluted due to non-standard organization of the code base.

There is a failure with FsCheck tests to be fixed.

PR #167, opened 2015-06-09
Fix desugar patterns' bug exposed by FsCheck

dungpa:
Let's see what FsCheck says.


dungpa:
Desugared patterns should be fine now.

There is a potential case with https://github.com/fsharp/FSharp.Compiler.Service/blob/2368210cd0cca11a46a6fa37f005c68a16af64e5/src/fsharp/ast.fs#L897; not sure when they're going to pop up.

PR #175, opened 2015-08-24
Added monadic bind operator pipelining and unit test

### LINKED ISSUE #173: Monadic Bind (>>=) pipelining
The Bind should probably be treated identically to the way the forward pipe operator `|>` is treated, as it's commonly used as part of pipelined expressions.

Having this included would greatly help those of us that are writing code that utilizes it.

I doubt this would be hard to implement, so if you can provide an outline of what exactly would need to be changed to do this, I'll be happy to give it a try.

I've included an example showing where the formatter trips up - this is done using the version included in Fsharp Power Tools.

```fsharp

open System

// taken from http://fsharpforfunandprofit.com/posts/computation-expressions-intro/
let strToInt = 
    Int32.TryParse >> function 
    | (true, v) -> Some v
    | (false, _) -> None

let strAdd str i = 
    match strToInt str with
    | Some v -> Some(v + i)
    | None -> None

let strAddLong _ = strAdd

let (>>=) result f = 
    match result with
    | Some v -> f v
    | None -> None
let okayToInline = strToInt "1" >>= strAdd "2" >>= strAdd "3" >>= strAdd "4"

let expectedResult =
    strToInt "1"
    >>= strAddLong "A long argument that is ignored" "2"
    >>= strAddLong "A long argument that is ignored" "2" 
    >>= strAddLong "A long argument that is ignored" "2"
    >>= strAddLong "A long argument that is ignored" "2"
    >>= strAddLong "A long argument that is ignored" "2"
    >>= strAddLong "A long argument that is ignored" "2" 


let actualResult = 
    strToInt "1" >>= strAddLong "A long argument that is ignored" "2" >>= strAddLong "A long argument that is ignored" "2" 
    >>= strAddLong "A long argument that is ignored" "2" >>= strAddLong "A long argument that is ignored" "2" 
    >>= strAddLong "A long argument that is ignored" "2" >>= strAddLong "A long argument that is ignored" "2" 
```

varon:
Fantomas master isn't building correctly, so I can't run the test.

If it works, this should resolve #173.

### DIFF
--- a/src/Fantomas.Tests/OperatorTests.fs
+++ b/src/Fantomas.Tests/OperatorTests.fs
@@ -155,3 +155,26 @@ let ``should not mess up ?<- operator``() =
     formatSourceString false """x?v <- 2""" config
     |> should equal """x?v <- 2
 """
+
+
+[<Test>]
+let ``should pipeline monadic bind``() =
+    formatSourceString false """
+strToInt "1"
+>>= strAddLong "A long argument that is ignored" "2"
+>>= strAddLong "A long argument that is ignored" "2"
+>>= strAddLong "A long argument that is ignored" "2"
+>>= strAddLong "A long argument that is ignored" "2"
+>>= strAddLong "A long argument that is ignored" "2"
+>>= strAddLong "A long argument that is ignored" "2"
+"""
+        config
+    |> should equal  """
+strToInt "1"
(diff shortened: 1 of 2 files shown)


varon:
Is there anything else this needs to be accepted?

dungpa:
No. It's great.

I will make sure to get this into VFPT 2.1.0 release.

PR #182, opened 2015-11-11
Setup AppVeyor build

dungpa:
Try to reproduce test errors when only VS 2013 is available.


dungpa:
@FeodorFitsner Why does the 'Windows Server 2012 R2' image have VS 2015 assemblies installed? I try to reproduce test errors if only VS 2013 is available. The tests pass on 'Windows Server 2012 R2' image but fail on 'Previous Windows Server 2012 R2' image (as expected).

Is 'Previous Windows Server 2012 R2' image stable for use?

FeodorFitsner:
For OSS accounts default `Windows Server 2012 R2` image has always been with VS 2015 installed. Why would you need running it on "Previous"?

dungpa:
I need an isolated environment of VS 2013 without VS 2015 installed. It would help reproduce the errors of the project because it has (wrong) dependencies on VS 2015. Thanks.

FeodorFitsner:
Fair enough. I've added a new `Visual Studio 2013` image which allows you explicitly target VS 2013.

dungpa:
@FeodorFitsner It works. Thanks for the great help.

PR #197, opened 2016-04-22
Records with implementation visibility attribute

### LINKED ISSUE #194: Formatter doesn't work on Opaque Record Type
The formatter seems to crash, or completely refuse to format any `.fs` file with an opaque record type.

Input:
```fsharp
module MinExample
type Example = private {Value:int; Value2:int}
```
Output:
```fsharp
module MinExample
type Example = private {Value:int; Value2:int}
```

I would expect it to work like it does without the private flag, as evidenced below:

Input:
```fsharp
module MinExample

type Example = {Value:int; Value2:int}
```
Output:
```fsharp
module MinExample

type Example = 
    { Value : int
      Value2 : int }
```

OkayX6:
Fixes #194

Time to fix it: ~2h (indicative - for reference).
- Time to rebase, reunderstand the project, make a fix

### DIFF
--- a/src/Fantomas.Tests/RecordTests.fs
+++ b/src/Fantomas.Tests/RecordTests.fs
@@ -16,6 +16,7 @@ type AParameters =
 """
 
 [<Test>]
+let ``record declaration with implementation visibility attribute``() =
     formatSourceString false "type AParameters = private { a : int; b: float }" config
     |> prepend newline
     |> should equal """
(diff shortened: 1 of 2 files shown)


dungpa:
Thanks.

> Time to fix it: ~2h (indicative - for reference).

I'll buy you coffee next time :).

OkayX6:
I haven't had this coffee! 😆

PR #212, opened 2017-06-17
README: update FantomasCLI package link

ForNeVeR:
Because only that package contains both the executable and the library.


dungpa:
@ForNeVeR Thanks.

PR #217, opened 2017-11-04
Initial Attempt to Incorrect Formatting

odytrice:
This PR is the result of my attempt to fix Issue #204 

I've tried to fix this issue but Its been really complicated. I've spent the last 2 weeks on this issue and it's not been fun. I tried changing the merge algorithm to merge the tokens based on the position of the comments relative to the last significant token (non-whitespace) before it. However it ended up breaking way more than it fixed.

### What I've done
- Added a test for the issue.
- Narrowed it down to the `integrateComments` method

### Hypothesis of the problem
The opening braces throws the matching algorithm off balance and then somewhere down the line, the remaining tokens get dropped.
as seen here

```
emitting whitespace ' ' in new tokens
ADDING ' '
emitting matching delimiter '(' in new tokens
ADDING '('
injecting block comment '(*comment*)'
ADDING '(*comment*)'
ADDING ' '
dropping ')' from original text
dropping newline from orig tokens
dropping whitespace from orig tokens
dropping 'printf' from original text
dropping whitespace from orig tokens
```
Because of this, it drops all tokens, prints out all the comments and then because the tokens haven't been transferred across, a "default" clause in the match expression just  emits the rest of the code.

I tried changing the order of the match cases but they were too many and it broke more than it fixed. Which I think is due to the fact that I don't fully understand all the cases. I also tried rewriting it but it was way complicated that I thought it would be because of the previous reason.

Here is the full log output
```
Content: ((*comment*)) at start = 13, finish = 25
Content: "a" at start = 38, finish = 40
Content: "a" at start = 38, finish = 40
Content: "b" at start = 78, finish = 80
Content: "b" at start = 78, finish = 80
Content: "c" at start = 118, finish = 120
Content: "c" at start = 118, finish = 120
sticky information:
-- unknown
sticky information: let -- unknown
sticky information: -- unknown
sticky information: x -- unknown
sticky information: -- unknown
sticky information: = -- unknown
sticky information:
-- unknown
sticky information: -- unknown
sticky information: ( -- unknown
sticky information: (* -- unknown
sticky information: comment -- unknown
sticky information: *) -- unknown
sticky information: ) -- unknown
sticky information:
-- unknown
sticky information: -- unknown
sticky information: printf -- unknown
sticky information: -- unknown
sticky information: " -- unknown
sticky information: a -- unknown
sticky information: " -- unknown
sticky information:
-- unknown
sticky information: -- unknown
sticky information: // -- right
sticky information: -- unknown
sticky information: another -- unknown
[...]

PR #218, opened 2017-11-21
Fix #204

### LINKED ISSUE #204: Incorrect reformatting of comments
Formatting this (comment inside `()`):

``` F#
let x =
    ((*comment*))
    printf "a"
    // another comment 1
    printf "b"
    // another comment 2
    printf "c"
```

results in the following (comment text doubled and another comments in the file are placed between two comment copies):

``` F#
let x = 
    ((*comment*) // another comment 1
                 // another comment 2
                 (*comment*))
    printf "a"
    printf "b"
    printf "c"
```

Repeated reformatting will double all text in the comment.

jindraivanek:
Attempt to fix #204.

Problem in `integrateComments` was that block comment was also part of newTokens.

See also #217.

### DIFF
--- a/src/Fantomas.Tests/CommentTests.fs
+++ b/src/Fantomas.Tests/CommentTests.fs
@@ -363,3 +363,26 @@ let hello() = "hello world"
 let hello() = "hello world"
 (* This is a comment. *)
 """
+
+[<Test>]
+let ``should keep comments inside unit``() =
+    formatSourceString false """
+let x =
+    ((*comment*))
+    printf "a"
+    // another comment 1
+    printf "b"
+    // another comment 2
+    printf "c"
+
+"""  config
+    |> prepend newline
+    |> should equal """
+let x =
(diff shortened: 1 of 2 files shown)


odytrice:
Please could you explain a bit more about how you discovered and solved the problem? I've gone through your PR. What exactly do you mean by "Raw Comment"? 
Thanks in advance

jindraivanek:
I put breakpoint in place where first "(" occurs (https://github.com/dungpa/fantomas/blob/master/src/Fantomas/TokenMatcher.fs#L589) and then follows what happens next. Turns out that alg got out of balance because on top of `newTokens` is `Comment` tokens, which is case that is not considered at all.

What I did, is in case of comments in `newTokens` just output it, while throwing out matching token from origTokens (assuming there is same comment token).

`RawComment` naming just follows used naming of things in newTokens, for example `RawDelimeter` or `RawAttribute`.

PR #221, opened 2017-11-26
 Use lookup for return type in extern definition. (Fix #189).

### LINKED ISSUE #189: should be `extern void` instead of `extern unit`
Similar to #186, I'm just reading in a file, having Fantomas format it, then writing it back out. The file this time is:

https://github.com/ctaggart/FsAst/blob/pinvoke/PInvoke.fs

``` fsharp
// https://msdn.microsoft.com/library/hh304361(v=vs.100).aspx
module PInvoke

open System.Runtime.InteropServices

[<DllImport("blas.dll",EntryPoint="dgemm_")>]
extern void dgemm_
  ( char *transa, char *transb, 
    int *m, int *n, int *k, 
    double *alpha, double *a, int *lda, 
    double *b, int *ldb, double *beta, 
    double *c, int *ldc );
```

When written back out, it has `unit` for the return type instead of `void`:

``` fsharp
module PInvoke

open System.Runtime.InteropServices

[<DllImport("blas.dll", EntryPoint = "dgemm_")>]
extern unit dgemm_(char nativeptr transa, char nativeptr transb, int nativeptr m, int nativeptr n, int nativeptr k, double nativeptr alpha, double nativeptr a, int nativeptr lda, double nativeptr b, int nativeptr ldb, double nativeptr beta, double nativeptr c, int nativeptr ldc)
```

jindraivanek:
Fix #189.

+ some DRY refactoring around lookup functions.

### DIFF
--- a/src/Fantomas.Tests/FunctionDefinitionTests.fs
+++ b/src/Fantomas.Tests/FunctionDefinitionTests.fs
@@ -214,6 +214,18 @@ module InteropWithNative =
 """
 
 [<Test>]
+let ``should handle external functions with void return type``() =
+    formatSourceString false """module InteropWithNative =
+        [<DllImport(@"__Internal", CallingConvention = CallingConvention.Cdecl)>]
+        extern void setCallbridgeSupportTarget(IntPtr newTarget)""" config
+    |> prepend newline
+    |> should equal """
+module InteropWithNative =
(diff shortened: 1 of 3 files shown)


ctaggart:
I love that my bugs are being fixed here. Thanks @jindraivanek.

PR #223, opened 2017-11-26
Don't remove parentheses around lambda "complex" parameter with type.

jindraivanek:
Fix issue from this comment: https://github.com/Microsoft/visualfsharp/pull/3542#issuecomment-337543936.


dungpa:
Thanks for this. I don't think we've added explicit support for struct tuples; that case is probably not working yet.

jindraivanek:
Yes, struct tuples are not supported. I create issue about that.

PR #237, opened 2018-05-02
Hotfix for runtime issue when executing dotnet-fantomas

nojaf:
https://www.nuget.org/packages/FSharp.Compiler.Service/ has a dependency on FSharp.Core higher or equal than 4.1.8, this caused for a mismatch in the generated `dotnet-fantomas.deps.json` located at `C:\Users\name\.nuget\packages\.tools\dotnet-fantomas\2.7.1\netcoreapp2.0`.

Long story short, the cli tool needs a direct dependency on FSharp.Core in order for this file to be generated correctly. I already pushed the fix to https://www.nuget.org/packages/dotnet-fantomas/2.7.1, it works now.

### DIFF
--- a/.paket/Paket.Restore.targets
+++ b/.paket/Paket.Restore.targets
@@ -48,16 +48,16 @@
 
     <!-- Because ReadAllText is slow on osx/linux, try to find shasum and awk -->
     <PropertyGroup>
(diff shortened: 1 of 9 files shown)


ErikSchierboom:
I can confirm that with this fix, I can run `dotnet fantomas` successfully as a CLI tool.

Issue #253, opened 2018-05-12
Wrong indention of nested else block


danyx23:
When testing Fantomas 2.7.0 against the [Argu](https://github.com/fsprojects/Argu) project, I noticed this bug in nested if blocks (notice the last line (nested else block) is aligned with the outer if block, not the inner)

Before:
```fsharp
    member __.AppendResult caseInfo context arguments =
        let result = mkUnionCase caseInfo resultCount ParseSource.CommandLine context arguments
        if result.CaseInfo.Depth = argInfo.Depth then
            __.AppendResultInner(result)
        else
            // this parse result corresponds to an inherited parameter
            // from a parent syntax. Use the ResultAggregator stack to
            // re-route the result to its matching aggregator
            if stack.TryDispatchResult result then ()
            else unrecognizedParseResults.Add result.Value
```

After:
```fsharp
    member __.AppendResult caseInfo context arguments =
        let result =
            mkUnionCase caseInfo resultCount ParseSource.CommandLine context 
                arguments
        if result.CaseInfo.Depth = argInfo.Depth then 
            __.AppendResultInner(result)
        else // this parse result corresponds to an inherited parameter
             // from a parent syntax. Use the ResultAggregator stack to
             // re-route the result to its matching aggregator
             if stack.TryDispatchResult result then ()
        else unrecognizedParseResults.Add result.Value
```

jindraivanek:
Possible duplicate of #241. Please try it on latest master.

danyx23:
@jindraivanek yes you are right, fixed on master, thanks!

Issue #233, opened 2018-03-05
Automate building of selected open source repositories


danyx23:
I think it would be valuable to automate a setup that takes a list of open source F# projects (ideally more complex projects, maybe even the F# compiler itself), downloads each from github, tries to build them and if that worked format everything with fantomas and see if it still builds. This would be a nice guard against regressions.

Has anything like this been done yet?

nojaf:
Infrastructure is now in place to build open source projects.
Thanks to @danyx23 , we have a fake target [TestExternalProjects](https://github.com/fsprojects/fantomas/blob/master/build.fsx#L185) that build projects, formats them and tries to build them again.

This is runnig on the [test-external-projects branch](https://github.com/fsprojects/fantomas/blob/test-external-projects/.travis.yml#L13) on Travis.

The only thing left here is adding some other projects. I open to suggestions. Something with Fable might be a good one.

jindraivanek:
I made my suggestions in form of PR: #338 :)

jindraivanek:
Closing as it is already done.

Issue #231, opened 2017-12-06
Missing newline, wrong indent in #if block.


jindraivanek:
1)
```
#if SOME_FLAG
printfn "Hello with flag"
#else
printfn "Hello without flag"    
#endif
```

Result

```
#if SOME_FLAGprintfn "Hello with flag"
#else
printfn "Hello without flag"
#endif
```

2)
```
do
#if SOME_FLAG
    printfn "Hello with flag"
#else
    printfn "Hello without flag"    
#endif
```

Result

```
do
#if SOME_FLAG printfn "Hello with flag"
#else
printfn "Hello without flag"
#endif
```

But

```
let x =
#if SOME_FLAG
    printfn "Hello with flag"
#else
    printfn "Hello without flag"    
#endif
```
works as expected.

SteveGilham:
Under the same umbrella, in processing a class to place `member` declarations starting in column 9, a method inside an #if was not consistently updated.

jindraivanek:
Example for `member` use:

```
type T = { x: int} with
#if SOME_FLAG
    static member X = ()
#else
    static member X = ()
#endif
```

output

```
type T =
    { x : int }

#if SOME_FLAGstatic member X = ()
#else
static member X = ()
#endif
```

nojaf:
I believe this is fixed in 2.8.1. Could you verify this please @SteveGilham?

SteveGilham:
I believe that the issues I marked in [#245](https://github.com/fsprojects/fantomas/issues/245) are indeed fixed in 2.8.1

nojaf:
Hmm you are right, some indentation info is lost. Could you create a new issue for this. I'd like to follow it up there.

PR #254, opened 2018-05-12
Allow easy build/format/build cycle of external projects (#233)

danyx23:
This PR is work in progress but I would appreciated some input on how to proceed, if the maintainers think this is a good approach etc. (thanks for your work btw, this is a really useful tool!)

I think a very important end to end test would be to check out some open source F# projects, build them, format them and try to build them again. If any of these fail, then fantomas introduced a syntax error.

The current approach is to add a new build step that can be called explicitly (i.e. because it is long running it would not be part of the normal build chain) that collects a list of external project references (only the Argu command line parser project at the moment) and then for each one:
* clones (or reset/pull) the project into a subdirectory
* builds the project (hardcodec to fake style build.cmd at the moment, needs to be more flexible)
* runs fantomas recursively on the subdirectory to reformat all source files
* rebuilds the project

In theory this could even be extended to compare the generated assemblies, but I am not sure if repeated builds are guaranteed to lead to binary identical assemblies in .NET... This would guard against errors like dropping indention one level too far for an if block if both indentions are syntactically valid.

Testing against just Argu already revealed #253 and another case of #231. The ticket related to this PR itself is #233


build.fsx
```diff
@@ -133,25 +145,61 @@ Target "Pack" (fun _ ->
...
 
+
+
+Target "TestExternalProjects" (fun _ ->
+    for project in externalProjectsToTest do
+        let relativeProjectDir = sprintf "external-project-tests/%s" project.DirectoryName
+        let buildCommandScript = sprintf "%s/%s/build.cmd" __SOURCE_DIRECTORY__ relativeProjectDir
+        if Fake.FileSystemHelper.directoryExists relativeProjectDir then
+            Fake.Git.Reset.ResetHard relativeProjectDir
+            Fake.Git.Branches.pull relativeProjectDir "origin" "master"
```
nojaf: Should we not build a stable release/tag instead? Then it is always reproducible.
Perhaps add this to the configuration.

nojaf:
We've noticed the target didn't ran on both CI systems?
How do you envision this additional test? Separate build? Part of CI anyway?

build.fsx
dsyme: Yes we should.  I'll mark this as ready for review and then please make this adjsutment

build.fsx
```diff
@@ -133,25 +145,61 @@ Target "Pack" (fun _ ->
...
 
-"Clean"
+"Root"
+  ==> "Clean"
   ==> "AssemblyInfo"
   ==> "ProjectVersion"
   ==> "Build"
   ==> "UnitTests"
+
+"UnitTests"
```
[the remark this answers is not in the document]
dsyme: It's harmless however.  I'll mark this as approved and then please submit a PR to cleanup

nojaf:
@danyx23 there is a new release out. This feature could really help us in testing.
Can we continue on this?

danyx23:
@nojaf yes sorry for the radio silence, have been travelling a lot lately but have more time again now

You know the project a lot more than I do, what do you think would be the desired behaviour for CI builds? 

As a very rough estimate let's assume that cloning, building, running fantomas and building again would take around 10 seconds on a CI server per project - how many projects would you think would be bearable to add? One way we could go is that we set up a special branch that has these tests and we only run an extensive suite of external projects at special occasions and only a small set (or none) at every CI invocation. What are your opinions?

nojaf:
I see two scenarios here:

1. We add the fake target to the existing CI and when including 1 project everything will take a couple of more minutes. Early feedback but it doesn't scale well when adding more projects.

2. A new separate build is started once a week and before a release. This could include a lot of various projects (web, desktop, cli, Fable, WebSharper, ...) and is ok to *take a while*. 
With this approach it be nice to just run the `fantomas.exe` from the [already built artifacts on appveyor](https://ci.appveyor.com/project/nojaf/fantomas/build/artifacts). It will better mimic the end user process.

I'm more in favor of the second approach. People write unit tests for every new PR, in theory we don't really have to test end to end for every delta. But to have that level of confidence when releasing it be nice to know that everything does indeed still work.

Every release I've done I've manually tested the next version with some of my projects/files. I'd like to see that part automated.

danyx23:
Option 2 sounds good. I'll look into this now and add the ability to specify a tag, do a shallow clone for faster checkouts etc.

As for which version to run - is there an easy way to retrieve the latest artifact from app veyor? They have an api but authenticating against that form a CI job could be involved so maybe it is easier to just build off of current master and run that fantomas.exe?

Should we integrate this into build.fsx or create a separate fake file for the external tests?

nojaf:
Maybe fetching the artifact from app veyor is a bridge too far. If we go with option two I would suggest a second script to keep things clean.

danyx23:
Just a quick note that I have not forgotten about this ticket; I have schedule some time for this in the week of October 8th, hope to provide a PR that is CI compatible then

nojaf:
Thanks for letting us know Daniel!

7sharp9:
I use Fantomas in Falanx, and catching any regressions would be nice such as #321, not sure how you would integrate this though.

danyx23:
Ok I worked some more on this, including some cleanup, optimizations of git checkout etc. To prepare for running this as part of CI I added a configuration function that allows platform dependent invocation commands for builds to be constructed (e.g. `build.cmd` fs `sh build.sh`)

I started testing this in a custom linux container with both dotnet and mono installed but ran into weird problems where neither Argu nor FSharp.Json nor FsPickler would build correctly. I noticed that xbuild is used instead of msbuild in this case and I am not sure why that is. Is there an official docker container that can be used to test the environment Travis would use to run?

I barely know Travis, so for actually running this as regularly as part of CI I need some help. Am I right in assuming that there can be only one .travis.yml file per repository? If so, did you envision to specify a conditional build step that would run build.fsx with the new TestExternalProjects target and this would be triggered periodically instead of at every commit?

Finally, the current external example project I use (FsPickler) does not build correctly after reformatting. It might be nice to use another project that successfully builds on windows and linux for finishing this PR :).

I would be happy for some feedback and if others could give this PR a try. I have some more time now so would be happy to steer this to become useful for all project contributors and for this PR to finally be merged and run on CI periodically.

nojaf:
I know https://github.com/ninjarobot/docker-fsharp-mono-netcore could be useful as container.
To be completely honest I don't know Travis that well either, I was hoping the same file could provide multiple builds. I'll test your PR when I find some time.

danyx23:
Ok I think this is gettings somewhere. Testing external projects works now on windows and on linux and I created a travis config that runs the external projects tests as a parallel stage.

Please try it and add some more projects you think are valuable for guarding against regressions.

Before merging into master we should finalize the travis config. My thoughts were that the best thing might be to create a new branch (e.g. "prerelease") that is used to trigger running this extensive test suite. We can also add cron based builds and add this as a condition to also run the extensive tests but I think it would be good if there were an easy way to explicitly signal to ci that you would like this tested and merging this into a prerelease branch and pushing would be the nicest way for this prior to a release I think.

Let me know if you have any comments or suggestions to further clean up the repo. I can also rebase the entire changes so that it will show up as a single commit for the merge.

nojaf:
Hi @danyx23, first things first: thanks a lot for all your work on this. It looks absolutely amazing.
Just for my understanding, if we have a branch `prerelease`. Would the plan then be to merge in master every once in a while? If this then triggers the tests then I think a cron expression might not be needed.

As for projects, we should add more but perhaps in a different PR. I have tested what you have now and I believe it is a good start.

nojaf:
Hi @danyx23, first things first: thanks a lot for all your work on this. It looks absolutely amazing.
Just for my understanding, if we have a branch `prerelease`. Would the plan then be to merge in master every once in a while? If this then triggers the tests then I think a cron expression might not be needed.

As for projects, we should add more but perhaps in a different PR. I have tested what you have now and I believe it is a good start.

nojaf:
Hi @danyx23, first things first: thanks a lot for all your work on this. It looks absolutely amazing.
Just for my understanding, if we have a branch `prerelease`. Would the plan then be to merge in master every once in a while? If this then triggers the tests then I think a cron expression might not be needed.

As for projects, we should add more but perhaps in a different PR. I have tested what you have now and I believe it is a good start.

danyx23:
Discussed this with nojaf in slack - I just pushed changes to enable building of external projects only on a specific branch (test-external-projects instead of prerelease) and added some comments. I had a look at doing an interactive rebase but with the many merge changesets this looked tricky so maybe you can just do a squash merge here in github when merging this PR instead?

danyx23:
Discussed this with nojaf in slack - I just pushed changes to enable building of external projects only on a specific branch (test-external-projects instead of prerelease) and added some comments. I had a look at doing an interactive rebase but with the many merge changesets this looked tricky so maybe you can just do a squash merge here in github when merging this PR instead?

Issue #252, opened 2018-05-08
dotnet global tool support


JonCanning:
See https://www.natemcmaster.com/blog/2018/02/02/dotnet-global-tool/

nojaf:
You can try an early build by downloading [the appveyor artifact](https://ci.appveyor.com/api/buildjobs/bqqsl9xtk3ew7gnj/artifacts/bin%2Ffantomas-tool.2.7.1.nupkg).

And install with 
`PS ~> dotnet tool install fantomas-tool -g --add-source ./YourDownloadsFolder`

PR #260, opened 2018-06-06
Adding support for fantomas as Global Dotnet Core tool. #252

nojaf:
Fantomas as a global tool would be a great addition (see #252). We have created a new `fsproj` and package for this since [Global tool only work for `netappcore2.1`](https://github.com/dotnet/cli/issues/9082) and we don't want to break support for the full framework in `Fantomas.Cmd`.


dungpa:
@nojaf Can you resolve conflicts? Thanks.

nojaf:
@dungpa resolved.

PR #268, opened 2018-07-03
Fix 267

### LINKED ISSUE #267: an exception erases source file
if target buffer is the same as source and the formatter throws an error, then the target and the source are empty and the original source will be lost.
to mitigate this problem there are a `--force` argument but this is by default `false`.

1.  it is counter intuitive that per default the files are erased. 
    could` --force` be set to `true` per default?
2.  why an extra argument at all? i don't see any user case that should allows formatter to destroy source files.
    could `--force` argument be removed and the original source be kept in the exception case?

P.S: i will make PR for case 1. but honestly i would prefer 2.

s-trooper:
This closes #267.

### DIFF
--- a/src/Fantomas.Cmd/Program.fs
+++ b/src/Fantomas.Cmd/Program.fs
@@ -96,7 +96,7 @@ let processSourceFile inFile (tw : TextWriter) config =
 [<EntryPoint>]
 let main _args =
     let recurse = ref false
-    let force = ref false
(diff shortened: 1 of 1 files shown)


nojaf:
@s-trooper AppVeyor is working again, could you add an additional commit to trigger it?

vasily-kirichenko:
@s-trooper Could you please merge with master?

Issue #251, opened 2018-05-07
dotnet fantomas does not work with > 1 .fsproj file in a directory


SteveGilham:
I add the `DotNetCliToolReference` to one project and `dotnet restore my.fsproj` it.
```
>dotnet fantomas --help
Specify which project file to use because this 'path/to/folder' contains more than one project file.
```
The same error message is given no matter where in the line I place the project 
```
dotnet my.fsproj fantomas --help
dotnet fantomas my.fsproj --help
dotnet fantomas --help my.fsproj
```

nojaf:
We don't support the older dotnet tool model anymore.
The problem can't occur with the global tool.

PR #271, opened 2018-07-05
Fix 270

### LINKED ISSUE #270: formatting compiler directives with inactive code is incorrect
There are couple of problems when formatting compiler directives with an inactive code. E.g. in this case the code is removed:
```fsharp
#if NOT_DEFINED
let x = 1
#endif
```
will be formatted to
```fsharp
#if NOT_DEFINED
#endif
```

s-trooper:
This closes #270.

### DIFF
--- a/src/Fantomas/CodeFormatterImpl.fs
+++ b/src/Fantomas/CodeFormatterImpl.fs
@@ -44,6 +44,8 @@ let createFormatContext fileName source projectOptions checker =
 
 let parse { FileName = fileName; Source = source; ProjectOptions = checkOptions; Checker = checker } = 
     async {
+        // This defines are needed for the Tokenizer
+        TokenMatcher.CompilationDefines <- checkOptions.ConditionalCompilationDefines
         // Run the first phase (untyped parsing) of the compiler
         let! untypedRes = checker.ParseFile(fileName, source, checkOptions)
         if untypedRes.ParseHadErrors then
--- a/src/Fantomas/CodeFormatterImpl.fs
+++ b/src/Fantomas/CodeFormatterImpl.fs
@@ -1,5 +1,5 @@
 ﻿[<RequireQualifiedAccess>]
-module internal Fantomas.CodeFormatterImpl
+module Fantomas.CodeFormatterImpl
 
 open System
 open System.Diagnostics
@@ -44,8 +44,6 @@ let createFormatContext fileName source projectOptions checker =
 
 let parse { FileName = fileName; Source = source; ProjectOptions = checkOptions; Checker = checker } = 
     async {
-        // This defines are needed for the Tokenizer
-        TokenMatcher.CompilationDefines <- checkOptions.ConditionalCompilationDefines
         // Run the first phase (untyped parsing) of the compiler
         let! untypedRes = checker.ParseFile(fileName, source, checkOptions)
         if untypedRes.ParseHadErrors then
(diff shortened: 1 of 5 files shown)


nojaf:
@s-trooper AppVeyor is working again, could you add an additional commit to trigger it?

src/Fantomas/CodeFormatterImpl.fs
```diff
@@ -44,6 +44,8 @@ let createFormatContext fileName source projectOptions checker =
 
 let parse { FileName = fileName; Source = source; ProjectOptions = checkOptions; Checker = checker } = 
     async {
+        // This defines are needed for the Tokenizer
+        TokenMatcher.CompilationDefines <- checkOptions.ConditionalCompilationDefines
```
nojaf: This seems a bit strange, what if we are parsing multiple files? Will the `CompilationDefines` be reset for each file or will they contains everything needed for all files?

Should we not return another type from the [parse](https://github.com/fsprojects/fantomas/blob/master/src/Fantomas/CodeFormatterImpl.fs#L45) functions that has the `ParsedInput` and the defines?

s-trooper: ``ConditionalCompilationDefines`` are the same for a whole Project, F# doesn't support ``#define``, does it? So formatting of multiple files in parallel is not a problem. A problem were to format multi projects in parallel (threads) but is it possible at all?

That said, i have nothing vs a thread safe solution but as far I understand it changing ``parse`` signature would require public API changes. I am not sure what sort of consensus needed here.

nojaf: Multiple projects does not work today, but is requested in https://github.com/fsprojects/fantomas/issues/251. I guess it could be a problem some day.

`parse` is indeed public api, maybe [format](https://github.com/fsprojects/fantomas/blob/master/src/Fantomas/CodeFormatterImpl.fs#L369) isn't? What if we get the info needed inside `parse` and collect it in `format` instead and pass it to `formatWith` down the line.

s-trooper: but ``formatAST`` is a part of public API and would need a signature change. I could wrap (change signature) ``parse`` like you suggested previously and unwrap in the API, that would work as long no one use ``parse`` directly outside there. 

The cleanest approach seems to me to change API but i would say #251 is better place for a API change as probably more things are needed than ``Defines`` forwarding.

nojaf: I'm a bit in the dark if changing the API is really a bad thing. The only editor I know that is using a recent version of Fantomas is Rider, perhaps @auduchinok could tell us if the change would have a lot of impact.

Most people are using Fantomas from the command line, or that is my assumption.
@dsyme do you have any feedback?

auduchinok: We're fine with changing signature of `FormatAST` (or anything else in the API) as long as it's possible to fix our use for the change so it works the same way.

s-trooper: I've removed the global variable and changed API as needed.

src/Fantomas/CodeFormatterImpl.fs
```diff
@@ -1,5 +1,5 @@
 ﻿[<RequireQualifiedAccess>]
-module internal Fantomas.CodeFormatterImpl
```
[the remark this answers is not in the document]
s-trooper: The data structure that i need to pass around is 'FormatContext' from the 'CodeFormatterImpl' module, for that reason i make this module public.

The alternative would be to move that structure to other (public) module, but that would require changes to the module dependencies graph at least. Or introduce a new public proxy structure, but that would harm the maintenance as people have to understand why a proxy structure there in the first place.

At the end, for good or for bad, i prefer small code changes as it is easier to reason about so i just make module public.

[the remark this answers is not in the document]
nojaf: Hmm I see, I can live with this. Btw, are you on fsharp slack? Ping me (also @nojaf) if you are.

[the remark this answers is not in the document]
s-trooper: No i am not.

PR #273, opened 2018-07-08
Fix "Fantômas" mistranslation in README

ThibaultLemaire:
Hello, French Grammar Nazi here :) I would like to point out that _Fantômas_ is a proper noun and has no meaning of its own in French. The translation for _ghost_ is _fantôme_.


nojaf:
@dungpa I'm guessing AppVeyor isn't working anymore since the move to `fsprojects`.
Could you check the configuration?

nojaf:
@ThibaultLemaire AppVeyor is working again, could you add an additional commit to trigger it again?

PR #275, opened 2018-07-12
Fix for preserve EOL feature

s-trooper:
the code formatted with ```preserveEOL``` feature will be broken and cannot compile, that PR should fix related issues.

P.S: PR #271  have to be accepted before, because a part of changes is based upon.

### DIFF
--- a/src/Fantomas.Tests/PreserveEOLTests.fs
+++ b/src/Fantomas.Tests/PreserveEOLTests.fs
@@ -455,3 +455,19 @@ let ``keep multiple pipes with pEOL``() =
     | 1 -> 2
     | _ -> 3
 """
+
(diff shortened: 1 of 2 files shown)


nojaf:
@s-trooper  AppVeyor is working again, could you add an additional commit to trigger it?

nojaf:
Hey @s-trooper, could you resolve the conflict? And I'm guessing this needs a bit of rework since we are not going with `let mutable CompilationDefines: string list = []` in #271

s-trooper:
Hi @nojaf, i've looked and updated this RP. The `CompilationDefines` rework has not effected this PR features.

nojaf:
Thanks @s-trooper

PR #279, opened 2018-07-19
Fix for recursive modules

### LINKED ISSUE #274: rec keyword is removed from recursive modules
`module rec Foo` becomes `module Foo`

7sharp9:
Fixes #274


nojaf:
Thanks @7sharp9

Issue #280, opened 2018-07-19
#if FOO || BAR => #if FOO


vasily-kirichenko:
```fsharp
#if DEBUG || RELEASE
#endif
```
=>
```
#if DEBUG
#endif
```

vasily-kirichenko:
Fixed via https://github.com/fsprojects/fantomas/pull/286

PR #286, opened 2018-07-21
Fix compiler directive on first line & compound directives

vasily-kirichenko:
The bug was in that 
```fsharp
#if FOO
let x = 1
#endif
```
was formatted as
```fsharp
#if FOOlet x = 1
#endif
```
This also fixes https://github.com/fsprojects/fantomas/issues/280


src/Fantomas.Tests/FormattingPropertyTests.fs
```diff
@@ -450,7 +450,8 @@ let rec shrinkSynExpr = function
     | SynExpr.MatchLambda(_, _, clauses, _, _) -> 
         seq { yield! Seq.collect collectSynMatchClause clauses }
     | SynExpr.TryWith(expr, _, clauses, _, _, _, _)
-    | SynExpr.Match(_, expr, clauses, _, _) -> 
+    | SynExpr.Match(_, expr, clauses, _, _) 
+    | SynExpr.MatchBang(_, expr, clauses, _, _) ->
```
vasily-kirichenko: I also fix incomplete pattern match by adding a branch for `match!`.

nojaf: Perhaps add this in another PR? Added the pattern match does not mean that `match!` is supported right?

Issue #243, opened 2018-05-06
Update ReadMe


xperiandri:
Please update readme according to a new version and @nojaf's [post](https://blog.nojaf.com/2018/05/03/fantomas-rejuvenated/)

PR #298, opened 2018-09-13
Fix CLI tool install instructions

alfonsogarciacaro:
It took me a while to install the tool because the README contained an outdated link (I guess FantomasCLI is not supported anymore) and I had to find Florian's blog post to remember the name of the package.

Other information from the README seems to be outdated too. If the preferred way now is to install `fantomas-tool`, should mentions to VS be deleted?


nojaf:
Hi Alfonso, thanks for this improvement. The readme is indeed outdated and should be tackled in https://github.com/fsprojects/fantomas/issues/243.

I'm also in favor of removing all references to Visual Studio and that "try online thing" and tsunami support. Guess I'll just need to bite the bullet and delete all that stuff.

PR #322, opened 2018-10-10
Fix for 321

7sharp9:
Only wrap formatContext.Source with an option when there is a string present.
Also added a test for this and refactored some test code to be reusable.


jindraivanek:
Thanks!

Issue #306, opened 2018-09-19
--preserveEOL adds newline at end of file if not already present


SteveGilham:
I would not have expected this when using this switch, issue #250 notwithstanding

nojaf:
Hi @SteveGilham , fantomas will always add a newline at the end of a file.
This behavior should not be determined by `--preserveEOL`.

Could you try this without the `--preserveEOL`? I would expect that the newline is added as well.

SteveGilham:
Yes, this behaviour also happens w/o --preserveEOL, but then in that case I'm not asking it to preserve what's at the end of the line, so that's allowable.

This issue is that `--preserveEOL` isn't preserving everything it would reasonably be expected to.

SteveGilham:
See also #250

nojaf:
Hi @SteveGilham , could you perhaps provide a code sample on [the online tool](http://ratatosk.dynu.net/fantomas) to highlight what the remaining issue is here.

SteveGilham:
Simply this

![screen shot 2018-11-27 at 08 27 55-fullpage](https://user-images.githubusercontent.com/7460854/49068723-5e039080-f21f-11e8-9477-bdca334d0632.png)

nojaf:
Ok, I'm afraid I'm going to close this one as `won't fix`. We discussed this and would like to remain opinionated on this.

Having a newline no matter what settings was in the product from the very beginning and has [its benefits](https://robots.thoughtbot.com/no-newline-at-end-of-file).

PR #324, opened 2018-10-12
[WIP] fix for 306

nojaf:
https://github.com/fsprojects/fantomas/issues/306 is indeed a problem.
I'll try to come up with a fix.

### DIFF
--- a/src/Fantomas.Tests/PreserveEOLTests.fs
+++ b/src/Fantomas.Tests/PreserveEOLTests.fs
@@ -471,3 +471,8 @@ let x = 1
 let y = 2
 #endif
 """
+
+[<Test>]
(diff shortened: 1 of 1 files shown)


nojaf:
Hmm ok, this test is a bit silly.
```fsharp
/// Format a source string using given config
let formatDocument config formatContext =
    async {
        let! formattedSourceCode = format config formatContext
        
        // When formatting the whole document, an EOL is required
        if formattedSourceCode.EndsWith(Environment.NewLine) then 
            return formattedSourceCode 
        else 
            return formattedSourceCode + Environment.NewLine
    }
```
So it makes sense that there is an newline at the end.
Starting to think this isn't really a bug.

PR #325, opened 2018-10-13
Upgrade to 452

nojaf:
I'm on a fresh Windows machine and it is no longer possible to download the developer pack for .NET 4.5.
[See list](https://www.microsoft.com/net/download/visual-studio-sdks).

4.5 is also end of life for a while now.


jindraivanek:
Why was `netcoreapp2.0` removed?

nojaf:
Good catch, that was not done on purpose, I'll restore this.

PR #330, opened 2018-10-18
Restored netcoreapp2.0

nojaf:
Not sure if I will publish 2.9.1 immediately. Might wait until someone really needs it.


nojaf:
Perhaps for @7sharp9 or @auduchinok, please let me know.

PR #348, opened 2018-10-31
Add braces when dealing with a lazy keyword.

### LINKED ISSUE #335: Lazy<'T> is incorrectly rewritten
```fsharp
let v = // <- Lazy "1"
    lazy
        1 |> string
```

gets rewritten as

```fsharp
let v : string Lazy = lazy 1 |> string

// v is now the string "Value is not created."
```

So precedence isn't being considered by the rewrite.

nojaf:
Should fix #335.
Does always wrap lazy values with `(` `)`. I am ok with this, not sure if always correct.


jindraivanek:
Added handling cases of multiline expr after lazy
```
let v = // <- Lazy "1"
    lazy
        1
        |> id
        |> string
```

Issue #331, opened 2018-10-19
Fluent API with comments breaks code


isaacabraham:
This code (taken from the `Program.fs` file in the Server project of https://github.com/Zaid-Ajaj/tabula-rasa)

```fsharp
Log.Logger <- 
  LoggerConfiguration() 
    // Suave.SerilogExtensions has native destructuring mechanism
    // this helps Serilog deserialize the fsharp types like unions/records
    .Destructure.FSharpTypes()
    // use package Serilog.Sinks.Console  
    // https://github.com/serilog/serilog-sinks-console
    .WriteTo.Console() 
    // add more sinks etc.
    .CreateLogger()
```

gets rewritten as:

```fsharp
Log.Logger <- LoggerConfiguration().// Suave.SerilogExtensions has native destructuring mechanism
                                    // this helps Serilog deserialize the fsharp types like unions/records
                                    Destructure.FSharpTypes().// use package Serilog.Sinks.Console  
                                                              // https://github.com/serilog/serilog-sinks-console
                                                              WriteTo.Console().// add more sinks etc.
                                                                                CreateLogger()
```

which is actually invalid F# and does not build.

Using standard fantomas settings with 120 char line length.

In fact, fluent APIs in general don't get very well formatted e.g.:

```fsharp
Log.Logger <-
    LoggerConfiguration()
        .Destructure.FSharpTypes()
        .WriteTo.Console()
        .CreateLogger()
```

becomes

```fsharp
Log.Logger <- LoggerConfiguration().Destructure.FSharpTypes().WriteTo.Console().CreateLogger()
```

Although this is more aesthetic.

nojaf:
[...]

PR #349, opened 2018-11-02
better formatting for multiline fluent api.

nojaf:
Should improve #331, not sure why `should break on . operator and keep indentation` is currently failing.


src/Fantomas/CodePrinter.fs
```diff
@@ -605,7 +607,17 @@ and genExpr astContext = function
         expr
         +> indent 
         +> (col sepNone es (fun (s, e) -> 
-                autoNln (!- (sprintf ".%s" s) 
+                let currentExprRange = e.Range
+                let addNewlineIfNeeded ctx = 
+                    let willAddAutoNewline:bool = 
+                        autoNlnCheck (!- (sprintf ".%s" s) +> ifElse (hasParenthesis e) sepNone sepSpace +> genExpr astContext e) sepNone ctx
```
jindraivanek: DRY (line 621)

src/Fantomas/CodePrinter.fs
```diff
@@ -667,10 +679,16 @@ and genExpr astContext = function
...
+    | LongIdentSet(s, e, r) -> 
+        let addNewLineIfNeeded = 
+            let necessary = e.Range.StartLine > r.StartLine
+            let spaces = [1..e.Range.StartColumn] |> List.fold (fun acc curr -> acc +> sepSpace) id
+            ifElse necessary (sepNln +> spaces) id
+        !- (sprintf "%s <- " s) +> addNewLineIfNeeded +> genExpr astContext e
     | DotIndexedGet(e, es) -> addParenIfAutoNln e (genExpr astContext) -- "." +> sepOpenLFixed +> genIndexers astContext es +> sepCloseLFixed
     | DotIndexedSet(e1, es, e2) -> addParenIfAutoNln e1 (genExpr astContext) -- ".[" +> genIndexers astContext es -- "] <- " +> genExpr astContext e2
     | DotGet(e, s) -> 
+        let isMultiline = e.Range.StartLine <> e.Range.EndLine
```
jindraivanek: not used

src/Fantomas/SourceTransformer.fs
```diff
@@ -56,6 +56,9 @@ let rec multiline synExpr =
 
     | Tuple es ->
         List.exists multiline es
+        
+    | App(DotGet(App(DotGet(dg),_), _),_) -> 
+        true
```
jindraivanek: This leads to change of behavior of things like `let foo = X.F().G().H()`. Better would be to use `autoNlnCheck` in `genExpr`.

nojaf: Not sure where to add autoNlnCheck here.
`should break on . operator and keep indentation` fails again by removing this.

nojaf:
Not happy with that last fix.
```

**Comment by jindraivanek** [comment]
I have working alternative:

```
and genExprSepEqPrependType astContext prefix e ctx =
    let multilineCheck = futureNlnCheck (genExpr astContext e) sepNone ctx
    match e with
    | TypedExpr(Typed, e, t) -> (prefix +> sepColon +> genType astContext false t +> sepEq
                                +> breakNlnOrAddSpace astContext (multilineCheck || checkPreserveBreakForExpr e ctx) e) ctx
    | e -> 
        (prefix +> sepEq +> breakNlnOrAddSpace astContext (multilineCheck || checkPreserveBreakForExpr e ctx) e) ctx
```

```
let internal futureNlnCheck f sep (ctx : Context) =
    if not ctx.BreakLines then false else
    // Create a dummy context to evaluate length of current operation
    use colWriter = new ColumnIndentedTextWriter(new StringWriter())
    let dummyCtx = ctx.With(colWriter)
    let writer = (dummyCtx |> sep |> f).Writer
    let lines = writer.InnerWriter.ToString().Split([|'\r';'\n'|]) |> Seq.filter (String.IsNullOrWhiteSpace >> not) 
    (lines |> Seq.length) > 2
```

(`autoNlnCheck` is not enough because it check only for too big column)

This basically force newline after `let foo =` if following expr will be formatted multiline (for any reason).

But this break some other tests for multiline strings:

https://github.com/fsprojects/fantomas/blob/571970871c8493d870e103f6f26163ec36077cf3/src/Fantomas.Tests/StringTests.fs#L28-L48

https://github.com/fsprojects/fantomas/blob/571970871c8493d870e103f6f26163ec36077cf3/src/Fantomas.Tests/StringTests.fs#L128-L155

I personally quite like rule "binding is one-liner or there is new line after `=`", but it is question what behavior is correct, and if muliline string should behave differently.

nojaf:
I like your proposal. The rule

> "binding is one-liner or there is new line after =",

makes sense to me, perhaps we should consider changing the behavior of multi line strings.

PR #366, opened 2018-11-25
WIP Fix 362

nojaf:
@jindraivanek couple of problems with this solution:

I think `integrateComments` should receive the full configuration object.
Currently not possible due to FormatConfig and TokenMatcher being somewhat coupled.

The fix also looks a bit like firefighting, not sure if the problem is structurally fixed.


jindraivanek:
> I think integrateComments should receive the full configuration object.

Agree.

> Currently not possible due to FormatConfig and TokenMatcher being somewhat coupled.

Please move `FormatConfig` type to separate file. This would also help us synchronizing config with Fantomas-UI.

src/Fantomas/FormatConfig.fs
```diff
@@ -116,7 +116,7 @@ type internal ColumnIndentedTextWriter(tw : TextWriter) =
         member __.Dispose() =
             indentWriter.Dispose()    
 
-type internal Context = 
+type Context =
```
jindraivanek: It is removing `internal` really needed?

nojaf: No, my bad.

nojaf:
@jindraivanek I think the extraction of Context.fs is not really that breaking to the outside world.
Would keep the new version as a revision, what do you think?

jindraivanek:
Yeah, extraction of `Context` is better.

Issue #363, opened 2018-11-21
Extra newline is introduced when file ends with multiline comment


nojaf:
### Description

Fantomas adds another extra newline where not needed.

Before:

```fsharp
#r "System.Xml.Linq"

open System.Xml.Linq
open System.Xml.XPath

let xml = "<a>1</a>"

(*
Material number: /BGE_ALISE/GROUPE_UM
Shipment date: /BGE_ALISE/DAT_EXP
Supplier code: /BGE_ALISE/COD_AFFRET
Transport order reference: /BGE_ALISE/NO_UE
Start location: /BGE_ALISE/SIT_EXP
End location: /BGE_ALISE/CCLI_ADRLIV
*)

```

After

```fsharp
#r "System.Xml.Linq"

open System.Xml.Linq
open System.Xml.XPath

let xml = "<a>1</a>"


(*
Material number: /BGE_ALISE/GROUPE_UM
Shipment date: /BGE_ALISE/DAT_EXP
Supplier code: /BGE_ALISE/COD_AFFRET
Transport order reference: /BGE_ALISE/NO_UE
Start location: /BGE_ALISE/SIT_EXP
End location: /BGE_ALISE/CCLI_ADRLIV
*)

```

### Repro code

[...]

PR #367, opened 2018-11-26
Fix for #363

nojaf:


src/Fantomas.Tests/PreserveEOLTests.fs
```diff
@@ -516,4 +516,53 @@ let config =
 let config =
     [("n", "1")
      ("d", "2")]
+"""
+
+[<Test>]
+let ``ending with multiline comment should not introduce additional newline`` () =
+    let config: Fantomas.FormatConfig.FormatConfig =
```
jindraivanek:
```suggestion
    let config = { Fantomas.FormatConfig.FormatConfig.Default with PreserveEndOfLine = true }
```

jindraivanek: Other options than `PreserveEndOfLine` is not important

nojaf: `config` actually already ok on line 7.

PR #373, opened 2018-11-28
Add or remove () when necessary. Should fix #369

### LINKED ISSUE #369: Fantomas shouldn't remove parens when using the dynamic operator (?)
I just applied Fantomas to a project with great results, thanks a lot for your work!

I found however a tricky issue that took me a while to find out. I had a piece of code like this:

```fsharp
let memoEquals (x:'a) (y:'a) =
  (true, JS.Object.keys(x)) ||> Seq.fold (fun eq k ->
    eq && (isFunction x?(k) || eqByRef x?(k) y?(k)))
```

After applying Fantomas, it became:

```fsharp
let memoEquals (x: 'a) (y: 'a) =
    (true, JS.Object.keys(x))
    ||> Seq.fold(fun eq k -> eq && (isFunction x?k || eqByRef x?k y?k))
```

Note that parens are removed from `x?k` and `y?k`. This is changing the meaning because in the first version the generated code is `x[k]` while in the second is `x.k`. Ideally Fantomas should distinguish when the dynamic operator is being applied and keep parens in that case.

nojaf:


src/Fantomas/CodePrinter.fs
```diff
@@ -574,7 +574,14 @@ and genExpr astContext synExpr =
         !- s +> sepSpace +> sepOpenS +> genExpr { astContext with IsNakedRange = true } e 
         +> ifElse (checkBreakForExpr e) (sepNln +> sepCloseSFixed) sepCloseS
     // This supposes to be an infix function, but for some reason it isn't picked up by InfixApps
-    | App(Var "?", e::es) -> genExpr astContext e -- "?" +> col sepSpace es (genExpr astContext)
+    | App(Var "?", e::es) ->
+        match es with
+        | SynExpr.Ident(_)::_
```
jindraivanek: We should also use parens in this case - this means user used `x?(k)` and we shouldn't  remove parens in that case, see https://github.com/fsprojects/fantomas/issues/369#issue-384812957

src/Fantomas.Tests/DynamicOperatorTests.fs
```diff
@@ -0,0 +1,17 @@
...
+
+[<Test>]
+let ``Keep () when dynamic operator is used``() =
+    formatSourceString false "let memoEquals x = x?(k + 1)" config
+    |> should equal """let memoEquals x = x?(k + 1)
+"""
+
+[<Test>]
+let ``Remove () when dynamic operator is string``() =
+    formatSourceString false "let memoEquals x = x?(k)" config
```
jindraivanek:
```suggestion
    formatSourceString false "let memoEquals x = x?k" config
```

PR #375, opened 2018-11-29
Nightly nuget feed

nojaf:
Adding appveyor build number to nuget package version.


nojaf:
@jindraivanek this is currently working for me. Please try [instructions](https://github.com/nojaf/fantomas/tree/myget#early-builds).

Only thing missing now is a way to prevent the push when the build is against a PR.
We should only push when merged to master.

jindraivanek:
Looks like it is working

```
fantomas --version
Fantomas 2.9.1.190
```

It is `--version 2.9.1-ci190` part needed? Can we get something like `--version 2.9.1-latest`? If not, where user find latest number?

nojaf:
I've added some text to the readme about that. However I like your suggestion to just go with latest. 

Only thing left is now not to push on PR.

jindraivanek:
> Only thing left is now not to push on PR.

Isn't that ensured by this line?
https://github.com/fsprojects/fantomas/blob/8567c031f1a857f548780ca303d07d1641b61e1b/appveyor.yml#L8

nojaf:
No, now it is also doing it on the current PR, see https://ci.appveyor.com/project/nojaf/fantomas/builds/20676461.

I guess because it compares it against master.

nojaf:
@jindraivanek I believe I found a way not to push when PR.
Please review change (build.fsx in particular) and consider merging.

jindraivanek:
LGTM.

`Fake.EnvironmentHelper.getBuildParam "myget-key"`

This is getting MyGet key from some secret variable in AppVeyor?

nojaf:
Yes, myget-key is configured in appveyor.

PR #376, opened 2018-11-30
Use qualified name for inputPath and outputPath

tathanhdinh:
Sorry to bother you for such a trivial PR.

It just tries to add qualified name in matching values of `InputPath` and `OutputPath`, for example:

```
let outputPath = ref Notknown
```

becomes

```
let outputPath = ref OutputPath.Notknown
```

because I've learned from [here](https://www.ikriv.com/dev/misc/fsharp/fs_enums) that the compiler will not warn if there are ambiguities (it simply uses the most recent suitable value), so there would possibly exist ambiguities (in future, not now).

I'm still (re)learning F# and do not know much, so feel free to reject if you find anything inappropriate.

Many thanks.

### DIFF
--- a/src/Fantomas.Cmd/Program.fs
+++ b/src/Fantomas.Cmd/Program.fs
@@ -58,12 +58,14 @@ let time f =
     printfn "Time taken: %O s" sw.Elapsed
     res
 
+[<RequireQualifiedAccess>]
 type InputPath = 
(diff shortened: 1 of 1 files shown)


tathanhdinh:
@jindraivanek thanks for the kind review. I've updated the PR.

PR #377, opened 2018-12-01
Moved solution file to root folder.

nojaf:
Moved solution file up one level to see all files in Rider in `FileSystem` mode.
image

Also removed `tests` folder, contained old files.

### DIFF
--- a/tests/languageshootout/binarytrees.fs
+++ b/tests/languageshootout/binarytrees.fs
@@ -1,0 +1,48 @@
+﻿/// The Computer Language Benchmarks Game
+/// http://shootout.alioth.debian.org/
+///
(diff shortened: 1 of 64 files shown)


jindraivanek:
Why removing test sample files?

nojaf:
I believe they are not being used and we do a similar check in https://github.com/fsprojects/fantomas/tree/test-external-projects. I want to revert it if I think these files are still relevant.

jindraivanek:
Can we do it in another PR? Output dirs is also referenced in README https://github.com/fsprojects/fantomas#use-cases But probably we should remove it or transform it into some test case or at least update outputs.

Issue #283, opened 2018-07-20
pattern matched unions are formatted badly


7sharp9:
Using `sourceTransformer.fs` from this repo 

```fsharp
let rec multiline synExpr = 
    match synExpr with
    | ConstExpr _
    | NullExpr
    | OptVar _
    | SequentialSimple _ ->
        false

    | ObjExpr _
    | While _
    | For _
    | ForEach _
    | TryWith _
    | TryFinally _
    | Sequentials _
    | IfThenElse _ ->
        true

    | Paren e
    | SingleExpr(_, e)
    | TypedExpr(_, e, _)
    | CompExpr(_, e)
    | ArrayOrListOfSeqExpr(_, e)
    | DesugaredLambda(_, e)
    | Lambda(e, _)
    | TypeApp(e, _)
    | LongIdentSet(_, e)
    | DotGet(e, _)
    | TraitCall(_, _, e) ->
        multiline e
...
```

is formatted into this:

```fsharp
let rec multiline synExpr =
    match synExpr with
    | ConstExpr _ | NullExpr | OptVar _ | SequentialSimple _ -> false
    | ObjExpr _ | While _ | For _ | ForEach _ | TryWith _ | TryFinally _ | Sequentials _ | IfThenElse _ -> true
    | Paren e | SingleExpr(_, e) | TypedExpr(_, e, _) | CompExpr(_, e) | ArrayOrListOfSeqExpr(_, e) | DesugaredLambda(_, 
                                                                                                                      e) | Lambda(e, 
                                                                                                                                  _) | TypeApp(e, 
                                                                                                                                               _) | LongIdentSet(_, 
                                                                                                                                                                 e) | DotGet(e, 
                                                                                                                                                                             _) | TraitCall(_, 
                                                                                                                                                                                            _, 
                                                                                                                                                                                            e) -> 
        multiline e
```
Another piece of code in this file:

```fsharp
let (|OneLinerBinding|MultilineBinding|) b =
    match b with
    | LetBinding([], PreXmlDoc [||], _, _, _, _, OneLinerExpr _)
    | DoBinding([], PreXmlDoc [||], OneLinerExpr _)
    | MemberBinding([], PreXmlDoc [||], _, _, _, _, OneLinerExpr _)
    | PropertyBinding([], PreXmlDoc [||], _, _, _, _, OneLinerExpr _) 
    | ExplicitCtor([], PreXmlDoc [||], _, _, OneLinerExpr _, _) -> 
[...]

PR #391, opened 2019-01-16
Each pattern is on a newline.

nojaf:
Improves #283.


nojaf:
Not sure, 
>you can consider using ...

not the strongest of wording that it should be on the same line if it is short.
Now it always obviously consistent, I personally prefer that. However feel free to make an issue if you disagree.

Issue #354, opened 2018-11-09
Allow for saving arguments in a config file


baronfel:
Fantomas should allow for inputting formatting arguments via a config file, like `.fantomas-config`. This config file could be located by convention or by an explicit CLI flag.  Having such a file would allow for settings to be tracked in source control, applied without the user having to remember many cli flags, and make it easier to run formatting via shell utilities, FAKE, various editor toolings, etc.

prosconi:
@nojaf do we want to do this or not? If so, I wouldn't mind submitting a PR to get my feet wet into the repo

nojaf:
@prosconi I would like to see a more fleshed out suggested approach for this feature first.
What file format do you have in mind? Where would the config file reside? Project-level? User folder?

prosconi:
I was thinking the format would be JSON. As far as where the file lives, that is a tricky one since there are a lot of ways to invoke fantomas

* CLI Explicit -> Look for command line parameter
* CLI Folder -> Look for ``inputPath\.fantomas-config``
* CLI File -> Look for ``Path.GetDirectoryName(inputPath)\.fantomas-config``
* CLI StdIn -> ???
* API -> Require user to load config manually?

jindraivanek:
I am in favor of this.

Agree with format JSON, maybe some auto (de)serializing library will be best. Anyway, it should be fail-safe, missing field in config file should be replaced by default option, and extra field (option that not exist in Fantomas) should not break loading of config (but it should emit some warning on stdout).

Regarding location of config file, I would prefer unified behavior, where config file is applied to whole subdirectory - so we will look for config file in target dir and in all parent dirs.

API should not look for config file, but we could make API for load config from file. AFAIK all API calls take config as explicit parameter.

MangelMaxime:
Hello,

I saw this issue poping on twitter and read the thread.

> Agree with format JSON, maybe some auto (de)serializing library will be best. Anyway, it should be fail-safe, missing field in config file should be replaced by default option, and extra field (option that not exist in Fantomas) should not break loading of config (but it should emit some warning on stdout).

This description looks really similar to what I am doing for Nacara config file where I am using `Thoth.Json`. 

> it should be fail-safe, missing field in config file should be replaced by default option

It's easy to map missing field via `Option.defaultValue`

> extra field (option that not exist in Fantomas) should not break loading of config (but it should emit some warning on stdout).

`Thoth.Json(.Net)` only decode the field you ask it to do. So extra fields are not checked.

[...]

PR #425, opened 2019-03-04
PR for issue #354

prosconi:
This pull request has automatic searching for new JSON file ``.fantomas-config`` from the inputPath in the CLI. It is intended that loading of this file will *only* crash the application when specified in the CLI. The new CLI parameter name is ``--configFile``.

Please let me know what you think.


src/Fantomas/FantomasConfig.fs
```diff
@@ -0,0 +1,146 @@
...
+        Error(ex.Message)
+
+/// Unwraps tryLoad and throws an exception if there is an error
+let load configFilename =
+    match tryLoad configFilename with
+    | Ok x -> x
+    | Error x -> failwith x
+
+/// Saves the specified config to the specified file name
+let save configFilename config =
```
nojaf: Please do not introduce functions that are not being used.

src/Fantomas/FantomasConfig.fs
```diff
@@ -0,0 +1,146 @@
...
+            |> fun (cfg, warnings) -> 
+                { FileName = fileName
+                  Warnings = warnings
+                  FantomasConfig = cfg
+                  FormatConfig = Dto.toFormatConfig cfg }
+            |> Ok
+        | false -> 
+            Error(sprintf "File does not exist '%s'" fileName)
+    with ex -> 
+        Error(ex.Message)
```
nojaf: Why catch this if you are going to throw it anyway a few lines later?

src/Fantomas.Cmd/Program.fs
```diff
@@ -241,8 +247,35 @@ let main _args =
...
+        | true ->
+            match !inputPath with
+            | InputPath.Unspecified -> FormatConfig.Default
+            | InputPath.StdIn _     -> FormatConfig.Default
+            | InputPath.Folder f    -> loadConfig f
+            | InputPath.File f      -> loadConfig f
+        | false ->
+            let config = FantomasConfig.load !configFile // this will throw an exception if there is something wrong
+            printfn "Loaded config '%s'" config.FileName
+            config.Warnings |> Seq.iter (printfn "Warning! %s")
```
nojaf: When will this ever be useful? What warnings could you possibly have here?

src/Fantomas.Cmd/Program.fs
```diff
@@ -241,8 +247,35 @@ let main _args =
 
     ArgParser.Parse(options, handleInput, "Fantomas <input_path>")
 
+    let loadConfig path =
```
nojaf: Is a Result option really the correct type here? This seems like a good case for a DU with ConfigNotLoad, ConfigInvalid and Config.

src/Fantomas.Cmd/Program.fs
```diff
@@ -241,8 +247,35 @@ let main _args =
...
+    let baseConfig =
+        match String.IsNullOrWhiteSpace !configFile with
+        | true ->
+            match !inputPath with
+            | InputPath.Unspecified -> FormatConfig.Default
+            | InputPath.StdIn _     -> FormatConfig.Default
+            | InputPath.Folder f    -> loadConfig f
+            | InputPath.File f      -> loadConfig f
+        | false ->
+            let config = FantomasConfig.load !configFile // this will throw an exception if there is something wrong
```
jindraivanek: IMO we should never throw exception from CLI, emit error and exit with non-zero exit code.

src/Fantomas/FantomasConfig.fs
```diff
@@ -0,0 +1,146 @@
...
+open System
+open System.IO
+open Fantomas.FormatConfig
+open Newtonsoft.Json
+
+module Dto =
+
+    /// DTO for FormatConfig (we want to keep this separated from FormatConfig in case we change the FormatConfig internal type)
+    [<CLIMutable>]
+    type FantomasConfig =
```
jindraivanek: Do we really need separate type? I think it is OK to have warnings from loading JSON if we remove some config option.

jindraivanek:
Overall it looks good 👍 Thanks!

src/Fantomas/FantomasConfig.fs
jindraivanek: Both `load` and `tryLoad` have its place here, but I agree it would be better to have main body in `load` and call it from `tryLoad`.

src/Fantomas.Cmd/Program.fs
prosconi: Missing members are still warnings, but things like malformed JSON will throw

src/Fantomas.Cmd/Program.fs
prosconi: I like this

nojaf:
Closing in favour of the implementation in https://github.com/fsprojects/fantomas/pull/606

PR #434, opened 2019-04-21
Refactor preserve eol with trivia

nojaf:
Suggestion to tackle https://github.com/fsprojects/fantomas/issues/390
In essence, we collect information from the F# tokens that are not present in the AST (comments, newlines, keywords, ...) and we use to information in CodePrinter to restore the original behaviour.


nojaf:
@jindraivanek I've encountered something tricky while inserting newlines that have been found.

F.ex
```fsharp
type C () = 
    let rec g x = h x
    and h x = g x

    member x.P = g 3
```

We print this in [CodePrinter ](https://github.com/fsprojects/fantomas/blob/d5f7ad93f89013481dd8c1af25637e6748aa1bd5/src/Fantomas/CodePrinter.fs#L988)with 2 newlines after the `and`. 

However, those newlines are also found in the tokens. So the test fails like:
```
  Expected string length 75 but was 77. Strings differ at index 54.
  Expected: "... let rec g x = h x\n    and h x = g x\n\n    member x.P = g 3\n"
  But was:  "... let rec g x = h x\n    and h x = g x\n\n\n\n    member x.P = g 3\n"
```
This all makes sense but how will be ignore found newlines in the trivia when we are already printing them?

jindraivanek:
@nojaf I think we need to change `sepNln` to also get node range as parameter and remove `NewLine` trivia before (maybe we would also need version of `sepNln` for after node) that node. That way we only print "extra" newlines from trivia.

nojaf:
Yeah thought of that as well. However, that could lead to a tricky situation where you have a mix of newlines and comments. Worth investigating though.

nojaf:
Hi, in https://github.com/fsprojects/fantomas/pull/434/commits/abbc74c0f2ca7ad61c948240ee0a90aa1bfe184c I've upgraded to use FCS 30.

And change the public api https://github.com/fsprojects/fantomas/pull/434/commits/abbc74c0f2ca7ad61c948240ee0a90aa1bfe184c#diff-a08be5b4e8318802307cade3ba44981d. Mainly to support [ISourceText](https://github.com/fsprojects/fantomas/pull/434/commits/abbc74c0f2ca7ad61c948240ee0a90aa1bfe184c#diff-81310b74b3c9e1eb9b2289e05fe0809b) and remove some synchronous functions.

This is just an attempt, nothing is written in stone yet. Let me know if this works for you or if you see room for improvement.

//cc @auduchinok @7sharp9 @nosami @jindraivanek

nosami:
LGTM

nojaf:
Should we change the output of the test or handle array/list of primitives in a different way?

jindraivanek:
> handle array/list of primitives in a different way?

Dunno. For me personally current behavior is OK even in this case.

nojaf:
```
Starting test execution, please wait...
  ! running formatting twice should produce the same results [1ms]
  X long array sequence [996ms]
  Error Message:
     Expected string length 6340 but was 12324. Strings differ at index 21.
  Expected: "\nlet input =\n    [| 2; 0; -1; 2; -2; -3; 1; -2; 0; -8; -1; -7..."
  But was:  "\nlet input =\n    [| 2\n       0\n       -1\n       2\n       -2\n ..."
  ----------------------------------^
```
well there would be a lot of newlines in this case.
I think at some point we should group them and try to fit as many one line to save space.
I'll change the test for now and we can revisit this later.

PR #455, opened 2019-07-31
Fix for issue 452

JackMatusiewicz:
There was no spacing applied between the attribute type name and its arguments. This is probably because it was assumed that they'd always be between parentheses. Now Fantomas will check to see if the arg expression for attributes is between parentheses. If it isn't then it will add a space to ensure the code generated still compiles.


src/Fantomas.Tests/AttributeTests.fs
```diff
@@ -30,6 +30,17 @@ let ``attributes on expressions``() =
 do ()
 """
 
+[<Test>]
+let ``attributes without parentheses on expressions``() =
+    formatSourceString false """
+    [<MyValue 55>]
+    do ()""" config
```
Smaug123: Could you add a test for what happens if there are too many spaces in the attribute? (presumably they get stripped down to 1)

JackMatusiewicz: Yes, will also add similar tests for args in parens

jindraivanek:
Thanks for your contribution! ❤️ 

Nowadays all new changes are done in `trivia` branch (for upcoming Fantomas 3.0). Could you please [retarget](https://github.blog/2016-08-15-change-the-base-branch-of-a-pull-request/) this PR to `trivia` branch?

jindraivanek:
Is this ready?

src/Fantomas/Utils.fs
```diff
@@ -2,6 +2,13 @@
 
 open System
 open System.Text.RegularExpressions
+open FSharp.Compiler.Ast
+
+[<RequireQualifiedAccess>]
+module SynExpr =
```
nojaf: Does it not make more sense to move this to SourceParser or SourceTransformer?
@jindraivanek this feels inconsistent.

src/Fantomas/CodePrinter.fs
```diff
@@ -377,8 +379,10 @@ and genAttributesCore astContext (ats: SynAttribute seq) =
         match e with
         | ConstExpr(Const "()", _) -> 
             opt sepColonFixed target (!-) -- s
-        | e -> 
-            opt sepColonFixed target (!-) -- s +> genExpr astContext e
+        | e ->
+            let argSpacing =
+                if SynExpr.isInParens e then id else (!- " ")
```
nojaf: Use `sepSpace` instead of `!- " "`

src/Fantomas/Utils.fs
jindraivanek: Yes, SourceTransformer looks like good place for this.

src/Fantomas/CodePrinter.fs
JackMatusiewicz: Done

src/Fantomas/Utils.fs
JackMatusiewicz: Deleted entirely, found SourceTransformer.hasParenthesis that does exactly what I need.

PR #463, opened 2019-08-13
update to fcs 31

baronfel:
This updates FCS to version 31, which included some breaking changes to the Attribute and SimplePat return types. To prevent having to rework _all_ the logics, I created helper patterns that break down the new collection-of-collections structures into simple lists.

The new collection-of-collections structures contain the enveloping ranges, though, so that might be interesting for formatting purposes if you want to look into that more.


nojaf:
Please target the `trivia` branch as all the latest development is being done over there.
We are already on FCS 30 there.

baronfel:
Ah, thanks for the pointer. I'll retarget it and push the change.

baronfel:
Ok, change fixed and retargeted to that branch.

PR #468, opened 2019-08-22
Fix for abstract member declaration with optional param

sasmithjr:
Fixes an issue in an abstract member declaration where the only parameter is a non-tuple, non-function, optional param. Formatting was dropping the `?` on the parameter.

For example:
```fsharp
type A =
    abstract B: ?p:float -> unit
```
would format to
```fsharp
type A =
    abstract B: p:float -> unit
```

The issue was in the printer: the `_` wildcard case was not account for the `isOpt` boolean. When making the wildcard account for `isOpt`, I noticed all three matches were the same except for the `hasBracket` bool, so I moved only that bool to the matching statement and sent to the same printing code underneath.

Added some test cases that account for both the fix and the changed code. Members `B` and `C` exercise the `TTuple` case for optional parameters, and member `D` exercises the `TFun` case for an optional parameter. Member `E` is the specific test for this fix, and the rest are additional tests as I didn't see anything that matched them in the existing tests.

Let me know if there is anything that needs to be changed!

### DIFF
--- a/.paket/Paket.Restore.targets
+++ b/.paket/Paket.Restore.targets
@@ -116,7 +116,7 @@
     </PropertyGroup>
 
     <PropertyGroup Condition=" '$(MSBuildSupportsHashing)' == 'true' And '$(CacheFilesExist)' == 'true' ">
-      <!-- If the restire file doesn't exist we need to restore, otherwise only if hashes don't match -->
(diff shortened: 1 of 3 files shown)


nojaf:
Thanks!

Issue #502, opened 2019-10-10
Sufficiently indent match case bodies for other indentation lengths than 4


cmeeren:
The following is my own style when using 2-space indentation. Note that I indent the case bodies extra to push them past the expressions I match on.

```f#
match x with
| Some y ->
    let z = 1
    Some (y + z)
| None -> None
```

The following is [Fantomas](https://jindraivanek.gitlab.io/fantomas-ui/#?fantomas=preview-trivia&code=LYQwLgxgFgBAHjA7gSzFAsAKAD4wMoD2wApjAJ4wC0AfFjPTADbFgwBeMAvDAIx0OESMABQUA1OwCUWXADkCAO1I0Y8pUA&config=N4IgkgdgJgphAuBlADgQwMYwHIFcC2IAXAEwA0IACqgOYwDqAllPABZEAcADAL5A) with 2-space indentation:

```f#
match x with
| Some y ->
  let z = 1
  Some(y + z)
| None -> None
```

I suggest that Fantomas makes sure to indent the case bodies past the match expressions (e.g. by indenting relative to the start of the case match expression instead of the pipe symbol).

Fantomas [already does](https://jindraivanek.gitlab.io/fantomas-ui/#?fantomas=preview-trivia&code=DYUwLgBAhhC8ECMIB8B8EAyBLAzmAdALZQAOEAFAGYCuAdhQMYA0EAFgJQQC0qAsAFAQhw0JAAecCACYBw4WNkR2QA&config=N4IgkgdgJgphAuBlADgQwMYwHIFcC2IAXAEwA0IACqgOYwDqAllPABZEAcADAL5A) something similar for lambda bodies:

```f#
let a =
  b
  |> List.map (fun (c, h) ->
       let x = 2
       x)
```

Note how the body is indented relative to `List.map`, not the pipe symbol.

PR #511, opened 2019-10-12
Fix for #502: Sufficiently indent match case bodies for other indentation lengths than 4

jindraivanek:


src/Fantomas/CodePrinter.fs
```diff
@@ -1660,10 +1660,10 @@ and genClause astContext hasBar (Clause(p, e, eo) as node) =
         else
             preserveBreakNln astContext e ctx
 
+    let pat = genPat astContext p
+    let body = optPre (!- " when ") sepNone eo (genExpr astContext) +> sepArrow +> clauseBody e
     genTriviaBeforeClausePipe p.Range +>
-    ifElse hasBar sepBar sepNone +> genPat astContext p
-    +> optPre (!- " when ") sepNone eo (genExpr astContext) +> sepArrow
-    +> clauseBody e
+    ifElse hasBar (sepBar +> atCurrentColumnWithPrepend pat body) (pat +> body)
```
nojaf: I'm a bit lost what `atCurrentColumnWithPrepend ` does at this point.
In the example of 
```fsharp
| Some y ->
    let z = 1
    Some(y + z)
```
It will print `Some y` first? And what happens then? Because the arrow `->` is still right after `pat` so how does the `atIndentLevel` work then?

jindraivanek: Its about that we store current column (at start of `Some`), then do `prependF` without setting indentation column, then set indentation column to stored value and then do rest.

This is needed for multiline pattern like
```
 | P1 y 
 | P2 y ->
    let z = 1
    Some(y + z)
```

I am all in for suggestion for better name of better comment for `atCurrentColumnWithPrepend` :)

PR #539, opened 2019-10-25
Fix: print line comment after arrow in lamda between paranthesis.

### LINKED ISSUE #534: Comments like `(fun arg -> // comment` are lost
### Description

As per the title

### Repro code

https://jindraivanek.gitlab.io/fantomas-ui/#?code=PoFQhgTg5gpgLgAgEQDEDKALSAHEBPbGAZwBEB7OAOXiQQAoAzAVwDsFgEBaAPgQHo+CMgCMiZADbwYAWABQ9AJQKgA&config=N4IgkgdgJgphAuBlADgQwMYwHIFcC2IAXAEwA0IACqgOYwDqAllPABZECcADAL5A

nojaf:
Fixes #534 

Currently, two tests fail but I'm a bit confused about those tests in the first place.

```fsharp
[<Test>]
let ``line comment after "if"``() =
    formatSourceString false """
if //comment
    true then 1
else 0""" config
    |> prepend newline
    |> should equal """
if true then 1 //comment
else 0
"""

[<Test>]
let ``line comment after "else"``() =
    formatSourceString false """
if true then 1
else //comment
    0""" config
    |> prepend newline
    |> should equal """
if true then 1
else 0 //comment
"""
```

Why can't the comments not be after the keyword like in the original code?

### DIFF
--- a/src/Fantomas.Tests/Fantomas.Tests.fsproj
+++ b/src/Fantomas.Tests/Fantomas.Tests.fsproj
@@ -1,4 +1,4 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
   <Import Project="..\netfx.props" />
   <PropertyGroup>
     <Version>3.0.0</Version>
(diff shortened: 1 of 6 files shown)


jindraivanek:
> Why can't the comments not be after the keyword like in the original code?

I don't remember exact details, but I think it was complicated to have correct indentation. Certainly it would be better to respect position of comment.

nojaf:
Closing in favour of https://github.com/fsprojects/fantomas/pull/557

PR #546, opened 2019-10-28
fix function application being formatted toward invalid syntax

### LINKED ISSUE #545: invalid code generated after multiline string when other expressions exist on same line
Issue created from [fantomas-ui](https://jindraivanek.gitlab.io/fantomas-ui/#?code=DYUwLgBAhhDOEF4CwAoC6IEsBmcIB4A+CAIhIjAAsQA7VDBiABwCdMaxsbSzsB7AagCksMuQywgA)

the spacing before last usage of `s` is important, it has to be greater than left column of the `printfn` function call:

#### Code

```fsharp
let a s =
    if s <> "" then
        printfn """fooo
%s"""    s
```        

#### Error

```fsharp
let a s =
    if s <> "" then
        printfn """fooo
%s""" s

```

#### Options

Fantomas 3.0.0

| Name | Value |
| ---- | ----- |
| `IndentSpaceNum` | `4` |
| `PageWidth` | `120` |
| `StrictMode` | `false` |
        
# More information

Related issue: #513

smoothdeveloper:
The indentation needs to be checked against the function expression start column in which case, prepend the "to be printed" expression with enough leading space.

fixes #545

Please let me know if the changes in implementation needs to be handled in a better way.

### DIFF
--- a/src/Fantomas/CodePrinter.fs
+++ b/src/Fantomas/CodePrinter.fs
@@ -1031,10 +1031,29 @@ and genExpr astContext synExpr =
             +> indent +> (ifElse (not hasPar && addSpaceBefore) sepSpace sepNone) +> appNlnFun e2 (genExpr astContext e2) +> unindent)
 
     // Always spacing in multiple arguments
-    | App(e, es) -> 
-        atCurrentColumn (genExpr astContext e +> 
+    | App(e, es) ->
+        // https://github.com/fsprojects/fantomas/issues/545
+        // we need to make sure each expression in the function application has offset at least greater than
+        // identation of the function expression itself
+        // we replace sepSpace in such case
+        let mutable savedColumn = 0
+        let retainColumn ctx =
+            savedColumn <- ctx.Writer.Column
+            ctx
+        let indentIfNeeded ctx =
+            let savedColumn = savedColumn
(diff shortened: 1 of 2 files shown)


src/Fantomas/CodePrinter.fs
```diff
@@ -1029,10 +1029,29 @@ and genExpr astContext synExpr =
...
+        // https://github.com/fsprojects/fantomas/issues/545
+        // we need to make sure each expression in the function application has offset at least greater than
+        // identation of the function expression itself
+        // we replace sepSpace in such case
+        let mutable savedColumn = 0
+        let retainColumn ctx =
+            savedColumn <- ctx.Writer.Column
+            ctx
+        let indentIfNeeded ctx =
+            let savedColumn = savedColumn
```
jindraivanek:
```suggestion
            let savedColumn = ctx.Writer.AtColumn
```

`AtColumn` can be used here to same effect, removing need for mutable variable.

smoothdeveloper: Thanks, I’ll give it a go!

smoothdeveloper: It works as you mentionned, thanks for the review.

PR #550, opened 2019-11-01
codeprinter: Fix SpaceBeforeArgument

### LINKED ISSUE #549: No way or no documentation on how to disable the `default = true` command-line options
### Description

The command-line flags that default to true, such as 	`--noSpaceBeforeArgument`, `--noSpaceAfterComma`, `--noSpaceAfterSemiColon`, `--noSpaceAroundDelimiter`, don't seem to accept a "false" string as a second argument, either by `--fooBar=false` or `--fooBar false`.

So it's not clear how they can be disabled.

VelocityRa:
`addSpaceBeforeParentsInFunCall` would always return false for
expressions that included an ident and parens. This fix makes it so
it returns true whenever the second expression (`arg`) is parens.

Fixes #549.

### DIFF
--- a/src/Fantomas/CodePrinter.fs
+++ b/src/Fantomas/CodePrinter.fs
@@ -43,9 +43,9 @@ type ASTContext =
 
 let rec addSpaceBeforeParensInFunCall functionOrMethod arg = 
     match functionOrMethod, arg with
-    | SynExpr.LongIdent(_, LongIdentWithDots s, _, _), _ ->
-        let parts = s.Split '.'
-        not <| Char.IsUpper parts.[parts.Length - 1].[0]
+    | SynExpr.LongIdent(_, _, _, _), ConstExpr(Const "()", _) ->
+        true
+    | SynExpr.Ident(_), ConstExpr(Const "()", _)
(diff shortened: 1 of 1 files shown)


VelocityRa:
It looks like all tests are written with the assumption that this option is `false` (no spaces) but the default is `true`.
Should I change the tests to have spaces or the option to `false`?

VelocityRa:
Hm, it looks like some tests expect spaces, and some don't?

Also in some cases my fix trims it without it being a function call, for example:
```
  Expected: "...   else Odd\n\nlet (|Integer|_|) (str: string) =\n    let mut..."
  But was:  "...   else Odd\n\nlet (|Integer|_|)(str: string) =\n    let muta..."
```
so I'll take a look at that. Probably need to specialize the matching.

nojaf:
I don't see how the configuration option is taking to account here?
All you did is swap some booleans right?

VelocityRa:
@nojaf It's taken into account, it's just not obvious why. You can see it from the call site.

`addSpaceBeforeParensInFunCall` should IMO have been called `canAddSpaceBeforeParensInFunCall `, since it's called/checked before the code actually checks for the configuration option.

Issue #555, opened 2019-11-06
SpaceBeforeArguments inconsistency: should apply to functions that receive `unit` arg too


knocte:
### Description

The SpaceBeforeArguments option adds a space when the function declaration has arguments, e.g. `someFunc(foo: Bar): Baz` gets expanded to `someFunc (foo: Bar): Baz`; however, if you remove the arguments of the function at some point in your code, the space disappears too (e.g. `let someFunc(): Baz`), which is not consistent.

### Repro code

https://jindraivanek.gitlab.io/fantomas-ui/#?code=PYBwpgdgBAygngZwC5gLYFgBQWsBsxJTARgCCATgOYAUAZsMAFxQBCAhuQJTPsBeUAXixQRUapxyZiZKtQAibFABUAlqjAA6AHLAA7hOyZ8hAK4QVSCjW5QzFwcNHjJdy7M5A

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 120.0 DAI (120.0 USD @ $1.0/DAI)  attached to it.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/fsprojects/fantomas/555/3664).
* Want to chip in? Add your own contribution [here](https://gitcoin.co/issue/fsprojects/fantomas/555/3664).
* Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
* $127,153.56 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

gitcoinbot:
Issue Status: 1. Open 2. **Started** 3. Submitted 4. Done 

<hr>

__Work has been started__.


These users each claimed they can complete the work by 1 month from now.
Please review their action plans below:


**1) [velocityra](https://gitcoin.co/velocityra) has started work.**

I'll familiarize myself with SpaceBeforeArgument's code more, figure out why the specific edge cases (camel case + no args) are broken and fix it.

Learn more [on the Gitcoin Issue Details page](https://gitcoin.co/issue/fsprojects/fantomas/555/3664).

gitcoinbot:
Issue Status: 1. Open 2. **Cancelled** 

<hr>

[...]

PR #556, opened 2019-11-07
Fix SpaceBeforeArgument edge cases

### LINKED ISSUE #554: SpaceBeforeArgument inconsistency: should apply to both camelCase and PascalCase funcs
### Description

The `SpaceBeforeArgument` option doesn't have any effect to functions with PascalCase names, as opposed to functions with camelCase names.

### Repro code

https://jindraivanek.gitlab.io/fantomas-ui/#?code=PYBwpgdgBAygngZwC5gLYCh0BsxKgYwENUwsBhQhMACgDNhgAuKAIUICcBKZtgLygC8Uap0w48ABUpFylGvSasO3JfyEigA

```
open System

let camelCase(foo: Bar): Baz = ()

let PascalCase(foo: Bar): Baz = ()
```

Current results:
```
open System

let camelCase (foo: Bar): Baz = ()

let PascalCase(foo: Bar): Baz = ()
```

Expected results:
```
open System

let camelCase (foo: Bar): Baz = ()

let PascalCase (foo: Bar): Baz = ()
```

VelocityRa:
Fixes #554 and #555.

105 tests need adjustment, I suppose I should do that in a separate commit? They're a lot fewer if I change the default.

I looked at a bunch of them and most seem alright, but match patterns are changed too, ie. this: https://github.com/fsprojects/fantomas/blob/cf56baefe5a971fd8726eb65e7234e91d36e346e/src/Fantomas.Tests/ComputationExpressionTests.fs#L62
becomes
```fsharp
          | Tree (x, left, right) ->
```
Is that acceptable?

### DIFF
--- a/src/Fantomas.Tests/ActivePatternTests.fs
+++ b/src/Fantomas.Tests/ActivePatternTests.fs
@@ -36,7 +36,7 @@ let (|Even|Odd|) input = if input % 2 = 0 then Even else Odd
 
 let (|Integer|_|) (str: string) =
    let mutable intvalue = 0
-   if System.Int32.TryParse(str, &intvalue) then Some(intvalue)
+   if System.Int32.TryParse (str, &intvalue) then Some(intvalue)
    else None
 
 let (|ParseRegex|_|) regex str =
@@ -52,7 +52,7 @@ let (|Even|Odd|) input =
 
 let (|Integer|_|) (str: string) =
     let mutable intvalue = 0
-    if System.Int32.TryParse(str, &intvalue) then Some(intvalue)
+    if System.Int32.TryParse (str, &intvalue) then Some(intvalue)
     else None
 
 let (|ParseRegex|_|) regex str =
--- a/src/Fantomas.Tests/AttributeTests.fs
+++ b/src/Fantomas.Tests/AttributeTests.fs
@@ -16,7 +16,7 @@ type Funcs =
     |> should equal """[<Extension>]
 type Funcs =
     [<Extension>]
-    static member ToFunc(f: Action<_, _, _>) = Func<_, _, _, _>(fun a b c -> f.Invoke(a, b, c))
+    static member ToFunc (f: Action<_, _, _>) = Func<_, _, _, _>(fun a b c -> f.Invoke (a, b, c))
 """
 
 [<Test>]
@@ -112,9 +112,9 @@ and [<Test>] b () = 10""" config
     |> prepend newline
     |> should equal """
 [<Test>]
-let rec a() = 10
+let rec a () = 10
 
-and [<Test>] b() = 10
+and [<Test>] b () = 10
 """
 
 [<Test>]
@@ -190,10 +190,10 @@ let printInStyle (style: string) (msg): unit = jsNative
 let printModel model: unit = jsNative
 
 [<Emit("console.trace()")>]
-let printStackTrace(): unit = jsNative
+let printStackTrace (): unit = jsNative
 #endif
 
-let e2e value = Props.Data("e2e", value)
+let e2e value = Props.Data ("e2e", value)
 """
 
 [<Test>]
--- a/src/Fantomas.Tests/ClassTests.fs
+++ b/src/Fantomas.Tests/ClassTests.fs
@@ -91,7 +91,7 @@ type Shape2D(x0 : float, y0 : float) =
        y <- y + dy
 
     abstract member Rotate: float -> unit
-    default this.Rotate(angle) = rotAngle <- rotAngle + angle
+    default this.Rotate (angle) = rotAngle <- rotAngle + angle
     """ config
     |> prepend newline
     |> should equal """
@@ -117,7 +117,7 @@ type Shape2D(x0: float, y0: float) =
         y <- y + dy
 
     abstract Rotate: float -> unit
-    default this.Rotate(angle) = rotAngle <- rotAngle + angle
+    default this.Rotate (angle) = rotAngle <- rotAngle + angle
 """
 
 [<Test>]
@@ -186,8 +186,8 @@ let ``classes and implicit constructors``() =
     |> should equal """
 type MyClass2(dataIn) as self =
     let data = dataIn
-    do self.PrintMessage()
-    member this.PrintMessage() = printf "Creating MyClass2 with Data %d" data
+    do self.PrintMessage ()
+    member this.PrintMessage () = printf "Creating MyClass2 with Data %d" data
 """
 
 [<Test>]
@@ -202,8 +202,8 @@ let ``classes and private implicit constructors``() =
     |> should equal """
 type MyClass2 private (dataIn) as self =
     let data = dataIn
-    do self.PrintMessage()
-    member this.PrintMessage() = printf "Creating MyClass2 with Data %d" data
+    do self.PrintMessage ()
+    member this.PrintMessage () = printf "Creating MyClass2 with Data %d" data
 """
 
 [<Test>]
@@ -221,7 +221,7 @@ and File(filename: string, containingFolder: Folder) =
     |> should equal """
 type Folder(pathIn: string) =
     let path = pathIn
-    let filenameArray: string array = System.IO.Directory.GetFiles(path)
+    let filenameArray: string array = System.IO.Directory.GetFiles (path)
     member this.FileArray = Array.map (fun elem -> new File(elem, this)) filenameArray
 
 and File(filename: string, containingFolder: Folder) =
@@ -278,7 +278,7 @@ let ``should keep parens in class inheritance in the right place``() =
     class
         inherit DGMLClass()
 
-        let functions = System.Collections.Generic.Dictionary<string, IState>()
+        let functions = System.Collections.Generic.Dictionary<string, IState> ()
     end
 """
 
@@ -362,7 +362,7 @@ System.String.Concat("a", "b" +
     member __.Property = "hello"
 
 let longNamedFunlongNamedFunlongNamedFunlongNamedFunlongNamedFun (x: T) = x
-let longNamedClasslongNamedClasslongNamedClasslongNamedClasslongNamedClasslongNamedClass = T()
+let longNamedClasslongNamedClasslongNamedClasslongNamedClasslongNamedClasslongNamedClass = T ()
 
 System.String.Concat
     ("a",
@@ -381,5 +381,5 @@ type Exception with
     |> should equal """open System
 
 type Exception with
-    member inline __.FirstLine = __.Message.Split([| Environment.NewLine |], StringSplitOptions.RemoveEmptyEntries).[0]
+    member inline __.FirstLine = __.Message.Split ([| Environment.NewLine |], StringSplitOptions.RemoveEmptyEntries).[0]
 """
--- a/src/Fantomas.Tests/CommentTests.fs
+++ b/src/Fantomas.Tests/CommentTests.fs
@@ -38,12 +38,12 @@ let print_30_permut() =
     """ config
     |> prepend newline
     |> should equal """
-let print_30_permut() =
+let print_30_permut () =
 
     /// declare and initialize
     let permutation: int array =
         Array.init n (fun i ->
-            Console.Write(i + 1)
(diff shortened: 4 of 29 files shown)


VelocityRa:
No feedback received, so I pushed adjustments to the tests.

Some edge cases that don't work are constructor calls via `new` (ie. `new WebClient()`),calls to exceptions raised with `raise` (ie. `raise (InnerError("inner")`) and method chains (ie. `WebHostBuilder().UseConfiguration(config)`).

These look like separate bugs, since neither `addSpaceBeforeParensInFunCall` nor `addSpaceBeforeParensInFunDef` are called for them, so I think I'm going to leave them aside for this PR.

Should I write tests for these and if so, where to put them?

This is what I was using for testing (formatted with the PR):
```fsharp
open System

let oneArg (foo: Bar): Baz = ()

oneArg (DateTime.Now)

let unitArg (): unit = ()

unitArg ()
unitArgLongIdent.Foo ()
UnitArgCamelCase.Foo ()

let camelCase (foo: Bar): Baz = ()
let PascalCase (foo: Bar): Baz = ()

let camelCaseUnitArg (): Baz = ()
let PascalCaseUnitArg (): Baz = ()
```

VelocityRa:
I tried 

```fsharp
[<Test>]
let ``SpaceBeforeArgument``() =
    formatSourceString false """
    let oneArg(foo: int): int = foo
    oneArg(2)
    """ config
    |> should equal """
    let oneArg (foo: int): int = foo
    oneArg (2)
    """
```
And I'm getting the following error with no more useful information:
```
Fantomas.FormatConfig+FormatException : Parsing failed with errors: [|/src.fsx (4,5)-(4,11) parse error Unexpected identifier in implementation file|]
```
Is that invalid F# or something? I'm new to the language.

nojaf:
Hello @VelocityRa , to see if your F# code is valid you can always use the [ast-viewer](https://jindraivanek.gitlab.io/ast-viewer/#?code=Q&defines=Q&filename=MoYwTglgDgLgdAMwM4A8g). If it doesn't parse there, you have a syntax error somewhere.

> No feedback received

Don't take this the wrong way but it was hard to see the impact of the PR without running the tests locally. So I indeed deliberately waited until the build was green to see the outcome.

When looking at the changes to the tests I'm afraid I can't approve these as too many things changed. This will appear as breaking behaviour and people will be upset about this.

To move forward I propose to introduce a new set of smaller settings that control one thing only and replace `SpaceBeforeArgument` with these.
This way we can proceed in a way that does not break the existing tests (so please revert those).

* **SpaceBeforeUnitParameterInLowercaseFunctionDefinition** *(default = false)*

changes
```fsharp
let a() = 10
```

to
```fsharp
let a () = 10
```

when true

* **SpaceBeforeUnitParameterInUppercaseFunctionDefinition** *(default = false)*

changes
```fsharp
let A() = 10

type Person() =
    member this.Walk() = meh
```

to
```fsharp
let A () = 10

type Person () =
    member this.Walk () = meh
```

when true

* **SpaceBeforeParenthesesParameterInLowercaseFunctionDefinition** *(default = true)*

changes
```fsharp
let bar(a: int) = a
```

to
```fsharp
let bar (a: int) = a
```

when true

* **SpaceBeforeParenthesesParameterInUppercaseFunctionDefinition** *(default = false)*

changes
```fsharp
let Bar(a: int) = a

type Person(x: string) =
    member this.Walk(a:int, b:string) = meh
```

to
```fsharp
let Bar (a: int) = a
type Person (x: string) =
    member this.Walk (a:int, b:string) = meh
```

when true

* **SpaceBeforeUnitArgumentInLowercaseFunctionCall** *(default = false)*

changes
```fsharp
foo()
```

to
```fsharp
foo ()
```

when true

* **SpaceBeforeUnitArgumentInUppercaseFunctionCall** *(default = false)*

changes
```fsharp
Foo()
person.ToString()
```

to
```fsharp
Foo ()
person.ToString ()
```

when true

* **SpaceBeforeParenthesesArgumentInLowercaseFunctionCall** *(default = true)*

changes
```fsharp
bar(7 + 8)
```

to
```fsharp
bar (7 + 8)
```

when true

* **SpaceBeforeParenthesesArgumentInUppercaseFunctionCall** *(default = false)*

changes
```fsharp
Bar(8+8)
```

to
```fsharp
Bar (8+8)
```

with these new settings, I would keep the existing behaviour of `SpaceBeforeArgument` intact by applying the suggested defaults.

All these new settings would obviously need an extensive set of unit tests to make sure that we covered all scenarios.

Make sure to cover:
- function definitions
- Object method calls
- Lamba
- Signature files
- Pattern matches on discriminated unions
- Class/Struct definitions
- ... <small>(things I can't come up with right now)</small>

> Not sure where to put tests for these, so I'm just going to make a new file.

You should add new tests in files where the code is related to the settings.
I think `LetBindings`, `LongIdentWithDotsTests`, `SignatureTests`, `ClassTests`, `StructTests`

VelocityRa:
Thanks for the extensive feedback @nojaf.

> to see if your F# code is valid you can always use the ast-viewer. 

I did use https://repl.it/languages/fsharp to test it and it worked fine there, but in the fantomas test it still threw the error, so I assumed I was doing something else wrong. So, still not sure what happened there.

> Don't take this the wrong way but it was hard to see the impact of the PR without running the tests locally. So I indeed deliberately waited until the build was green to see the outcome.

I see. I suspected that this might be the case, but the unfortunate consequence was that I had to waste at least an hour adjusting all these tests, knowing that it would likely get rejected - just to get some feedback.

In this case it was funded so I don't mind as much - but usually it won't be so here's some feedback for you too as fellow OSS maintainer: not a great way to welcome new contributors to your project.

---

As I mentioned, this work is funded and the agreed upon scope was much smaller than what's being requested.
If I change the default to `false`, I "only" need to adjust 37 tests instead of 105. That would not be acceptable either I imagine?

If not, and I implement all the options you asked for, would it be acceptable if the edge cases mentioned in https://github.com/fsprojects/fantomas/pull/556#issuecomment-552139260 weren't fixed? I have not looked at these at all and they involve parts of the codebase I'm not familiar with.

And what about the tests? I'm really new to F# so it'd take a while time for me to come up with and write an extensive collection, so how necessary would they be for my work to be merged?

knocte:
@nojaf when you say `Uppercase` or `Lowercase` I guess you mean `PascalCase` and `camelCase`, respectively.

But... don't get me wrong, are you seriously proposing 8 new settings? is this to deter us from making the contribution? Why not create a new `SpaceBeforeArgumentStrict` setting that covers all these cases, to avoid breaking backwards-compatibility but at the same time keep some sanity?

nojaf:
Changing the default settings and slightly changing some tests is not an option.
Either we solve this styling problem thoroughly or we don't. I want a solution where there is a lot of flexibility so that it is future proof and easy to maintain. Each new setting is so small that it represents only one thing and cannot invoke any ambiguity. To summarize please take the high road on this.

@VelocityRa I will create a contribution guideline document in the coming weeks. There I will elaborate on why unit tests are so crucial to this project and provide a set of guidelines to set the correct expectations when creating a PR.

@knocte I used the word `Lowercase` instead of camelCase to avoid confusion when someone would write weird constructs like `let foo_bar () = 42`. The Lowercase setting should still work here even though the identifier is snake case.

smoothdeveloper:
@nojaf, thanks for the review, I'd like to put minor notes for consideration:

For the sake of the codebase/implementation, I find it important to finely discriminate the usages of such settings (like the way you did), it doesn't mean all has to be surfaced yet as user visible settings; but it means the implementation adheres to a fine understanding of the formatting choices according to the language concepts, which I think is a good concern to have beside looking at particular issues individually.

If possible we should split the settings for function and member definition (+ constructors you mentionned in chat), would that make sense? probably remove the lower/upper case variants for definition sites unless such distinction remains useful on top of the split on the type of definition.

For the call sites, as is (based on untyped AST), using the StartsWith lower/upper case remains necessary.

Is there a motivation for Fantomas to evolve to also use the typed AST or there are already identified reasons (too complex, or others) to not go there eventually?

smoothdeveloper:
> I had to waste at least an hour adjusting all these tests

@VelocityRa the time spent on fixing the tests is not "wasted", it helps highlight all the cases that are impacted by the change to existing logic.

Not to sound too optimistic, but I'm sure people are going to be willing to help going with a fine adjustment of settings. Technically speaking your contribution address the funded issue, and initiates helpfull discussion for the project to keep moving in the right direction wrt to final implementation and elaborating more contributions guidelines, etc.

nojaf:
@smoothdeveloper lower/upper case variants for definition sites would be necessary to preserve the [current style](https://jindraivanek.gitlab.io/fantomas-ui/#?code=DYUwLgBAYg9jEAoCGAuAlgOzASggXgiQFgAoUSAIyQCdEL0tcCKg). Formatting using the default settings in the next version of Fantomas should lead to same result. You do have a point though that the problem lowercase/uppercase is more one on the calling side of functions. So I'm still a bit on the fence on this one.

I'm on board to even further split up functions, member definitions and constructors. This might be a correct argument to drop lower/upper case distinction and keep those strictly for the calling side.

I have never had any motivation to look at TAST, gut feeling this would bring in a whole new level of complexity and I don't think it is necessary to solve these cases.

knocte:
> @knocte I used the word Lowercase instead of camelCase to avoid confusion when someone would write weird constructs like let foo_bar () = 42. The Lowercase setting should still work here even though the identifier is snake case.

Fair enough.

> I'm on board to even further split up functions, member definitions and constructors. This might be a correct argument to drop lower/upper case distinction and keep those strictly for the calling side.

Would this translate in less settings? TBH I think the UX would be greatly diminished by having such an explosion of settings when typing `--help` for something that should really be very simple. I mean, myself I'm going to turn all this SpaceBeforeArgument subsettings to `true` for my project. Maybe wrap all of them under a single `--spaceBeforeArgumentAlways` flag?

nojaf:
I guess we could wrap all settings into one flag from the command line perspective only.
The internal configuration record could still work agnostic from the combine setting.
However, I'm not sure how that would play out if you set `--spaceBeforeArgumentAlways ` in combination with something else. A workaround could also be something like [this](https://github.com/fsprojects/fantomas/issues/354#issuecomment-440212775).

> Would this translate in fewer settings? 
Probably not, it might end up around the same amount.

VelocityRa:
What about
>If not, and I implement all the options you asked for, would it be acceptable if the edge cases mentioned in https://github.com/fsprojects/fantomas/pull/556#issuecomment-552139260 weren't fixed? I have not looked at these at all and they involve parts of the codebase I'm not familiar with.

nojaf:
Please implement this in a way that doesn't require any modifications in the existing tests.
My approach listed in this thread should be of help.

VelocityRa:
Got it; so it's fine if not all edge cases work, as long as no tests are changed.

knocte:
> so it's fine if not all edge cases work, as long as no tests are changed.

I don't interpret @nojaf's comment this way. What I get is that each edge case should be covered by a new setting (and a new setting deserves new tests, hence why all existing tests that run with default settings don't need changing). Am I right @nojaf ?

VelocityRa:
If that's the case, then he still hasn't  answered my question.

Above (https://github.com/fsprojects/fantomas/pull/556#issuecomment-552184750) he detailed 8 new options (which I already knew I will have to implement/test) and they are all irrelevant to the edge cases I've asked twice about, which will need no additional tests since in that scenario I would not be touching them at all.

nojaf:
Yes @knocte that is what I mean.
So don't change any code in the existing tests and proceed by adding new settings to implement the behavior.

You need to see it like this: someone formats code with the current version of Fantomas and the default settings.
When that person formats the same code with the next version of Fantomas (with again the default settings), the result should be the same.

VelocityRa:
Yes. I know. I got that part.

Please read https://github.com/fsprojects/fantomas/pull/556#issuecomment-552139260. This is what I have asked about 3 times by this point.
Do you care about these or not.

nojaf:
The new settings should have an impact on those things as well.

nojaf:
Closing in favor of https://github.com/fsprojects/fantomas/pull/649

PR #557, opened 2019-11-09
Fix: print line comment after arrow in lamda between paranthesis

nojaf:
In https://github.com/fsprojects/fantomas/pull/539 I tried to fix bug #534 and by I ended up with two broken tests that were somewhat strange and debatable.

This PR contains the same fix for #534 and also refactors the way `SynExpr.IfThenElse` is formatted.
As close to the [F# style guide](https://docs.microsoft.com/en-us/dotnet/fsharp/style-guide/formatting#formatting-if-expressions) as possible.

In the case of 
> If either cond, e1 or e2 are longer, but not multi-line

the guide leaves some room for interpretation. What is "longer" exactly?
Because of this question I introduced a new setting `MaxIfThenElseShortWidth` to let the user control this.

The main motivation to rewrite the formatting of `SynExpr.IfThenElse` is the support for comments around keywords (`if`, `elif`, `else if`), this have been tested extensively in `IfThenElseTests.fs`.


src/Fantomas.Tests/CommentTests.fs
```diff
@@ -326,12 +326,15 @@ else
...
     |> prepend newline
     |> should equal """
-if a then ()
+if a then
+    ()
 else
-    // Comment 1
-    if b then ()
-    // Comment 2
-    else ()
```
jindraivanek: IMO there should be indent after `else`.

src/Fantomas.Tests/IfThenElseTests.fs
```diff
@@ -0,0 +1,786 @@
...
+  e // c11
+"""  config
+    |> prepend newline
+    |> should equal """
+if // c1
+    a // c2
+    then // c3
+    b // c4
+else // c5
+if // c6
```
jindraivanek: no indent after `else`

src/Fantomas.Tests/IfThenElseTests.fs
```diff
@@ -0,0 +1,786 @@
...
+let ``comment after else keyword before if keyword`` () =
+    formatSourceString false """if  a then b else // meh
+    if c then d else e
+"""  config
+    |> prepend newline
+    |> should equal """
+if a then
+    b
+else // meh
+if c then
```
jindraivanek: no indent after `else`

nojaf:
The no indentation after the `else` part was intended after reading the [style guide](https://docs.microsoft.com/en-us/dotnet/fsharp/style-guide/formatting)

> Multiple conditionals with elif and else are indented at the same scope as the if:

I consider `else if` and `elif` to be the same thing.

The [fsharp lang ref](https://docs.microsoft.com/en-us/dotnet/fsharp/language-reference/conditional-expressions-if-then-else) seems to confirm this.

> When chaining if...then...else expressions together, you can use the keyword elif instead of else if; they are equivalent.

So that is why no indentation.

In an extreme example I wouldn't imagine you would write the following code like this right?
```fsharp
if aaaa then
    meh
else 
    if bbbb then
        fooo
    else
        if ccc then
            baaarrr
        else
            if dddd then
                meehhh
            else
                ()
```

jindraivanek:
> Multiple conditionals with elif and else are indented at the same scope as the if:

IMO this means only that `elif` and `else` is in same indentation as `if` not the block after `else`.

> The fsharp lang ref seems to confirm this.

I know it is valid code, I just always considered this as bad style.

But "it is same as `elif`" is quite good argument.

> In an extreme example I wouldn't imagine you would write the following code like this right?

Yeah, I would write it with `elif` ;)

----

To move forward, I think we can merge this, and raise an issue in style docs to clarify this.

nojaf:
I've created https://github.com/dotnet/docs/issues/15922, we can always later change the formatting to whatever the outcome ends up being.
@jindraivanek could you approve to move forward?

jindraivanek:
@nojaf Thanks!

knocte:
After this PR has been merged, is there a way to configure fantomas to make ifs always multi-line?

nojaf:
@knocte try setting `--maxIfThenElseShortWidth` to 0.

PR #578, opened 2019-11-28
Enum cases comments

theimowski:
Generate Trivia for enum cases

Also, changed the default pageWidth in docs which was wrong

### DIFF
--- a/src/Fantomas.Tests/CommentTests.fs
+++ b/src/Fantomas.Tests/CommentTests.fs
@@ -5,39 +5,6 @@ open FsUnit
 open Fantomas.Tests.TestHelper
 
 [<Test>]
-let ``comments on enums - https://github.com/fsprojects/fantomas/issues/572``() =
-    formatSourceString false """type A =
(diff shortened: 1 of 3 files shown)


src/Fantomas.Tests/CommentTests.fs
```diff
@@ -4,6 +4,39 @@ open NUnit.Framework
 open FsUnit
 open Fantomas.Tests.TestHelper
 
+[<Test>]
+let ``comments on enums - https://github.com/fsprojects/fantomas/issues/572``() =
```
nojaf: Minor nitpick, I write `, 100` to reference an issue.
See https://github.com/fsprojects/fantomas/blob/master/src/Fantomas.Tests/CompilerDirectivesTests.fs#L365
I should write these little things down somewhere. It doesn't really matter all that much and it will just improve some constancy.

theimowski: sure let me correct that

PR #587, opened 2019-12-03
Unwanted new line added - fixes #586

### LINKED ISSUE #586: Unwanted new line added
Issue created from [fantomas-ui](https://jindraivanek.gitlab.io/fantomas-ui/#?code=LYewJgrgNgpgBAQTgXgFBw3WAXOAPFOABlXUwG0AeAJRgEcIBLAJxgEUIBDKRgM0ZhgEAY2EwAzuIB8AXTIZQkWHABCKeZgw44AT0IBGVEA)

This issue contains 3 separate (probably) issues related to newline being inserted where it shouldn't.
First issue in this description, and 2 more will follow in comments.

extra new line before nested module with attribute:

#### Code

```fsharp
module A =
    let x = 0

    [<RequireQualifiedAccess>]
    module B =
        let y = 1

```
            

#### Result

```fsharp
module A =
    let x = 0


    [<RequireQualifiedAccess>]
    module B =
        let y = 1

```
            
#### Options

Fantomas 3.1.0

| Name | Value |
| ---- | ----- |
| `IndentSpaceNum` | `4` |
| `PageWidth` | `120` |
| `StrictMode` | `false` |

theimowski:
Fixes all cases from #586 except from the `elif` one - wasn't able to trace it down, and as a workaround for now I switched to using pattern matching in my project.

I'll extract the `elif` case from #586 and make it a separate issue

### DIFF
--- a/src/Fantomas.Tests/ClassTests.fs
+++ b/src/Fantomas.Tests/ClassTests.fs
@@ -386,7 +386,8 @@ type Exception with
 
 [<Test>]
 let ``no extra new lines between interface members, 569``() =
-    let original = """namespace Quartz.Fsharp
+    shouldNotChangeAfterFormat """
+namespace Quartz.Fsharp
 
 module Logging =
     open Quartz.Logging
@@ -417,11 +418,10 @@ module Logging =
     let SetQuartzLogger l = LogProvider.SetCurrentLogProvider(l)
 """
 
-    formatSourceString false original config |> should equal original
-
 [<Test>]
 let ``no extra new lines between type members, 569``() =
-    let original = """type A() =
+    shouldNotChangeAfterFormat """
+type A() =
 
     member this.MemberA =
         if true then 0 else 1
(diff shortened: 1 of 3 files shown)


src/Fantomas.Tests/ClassTests.fs
```diff
@@ -433,3 +433,41 @@ let ``no extra new lines between type members, 569``() =
 """
 
     formatSourceString false original config |> should equal original
+
+[<Test>]
+let ``no extra new line before nested module with attribute, 586``()=
+    let original = """module A =
```
nojaf: Maybe start using [shouldNotChangeAfterFormat](https://github.com/fsprojects/fantomas/blob/master/src/Fantomas.Tests/TestHelpers.fs#L145) helper function since all your tests are like this.
See https://github.com/fsprojects/fantomas/blob/818c5a956be2fdaf6f4f756c0bd8e5277cd99f38/src/Fantomas.Tests/RecordTests.fs#L205

theimowski: Yup makes sense - pushed a change

PR #607, opened 2019-12-21
Using Argu for commandline argument parsing

nojaf:
Proposal to use Argu for the command line argument parsing.
IMHO it looks nicer and we remove the ref variables.


nojaf:
Thanks for the review @jindraivanek , I'll try to resolve #612 and #613 with this PR as well.
So I'll add a couple more changes to this.

PR #623, opened 2020-01-03
Add braces to sequence expression inside computation expression.

### LINKED ISSUE #553: Sequence expression inside computation expression outputs uncompilable code
### Description
A sequence expression inside a computation expression is converted to uncompilable code by fantomas, as it removes one pair of curly braces.

### Repro code
See [this example](https://jindraivanek.gitlab.io/fantomas-ui/#?fantomas=v2-9-2&code=DYUwLgBAnhC8EEMDOUB2BjCBvAsAKAkIgCdwBXY1bCABgDo6IBGCAX31aA&config=N4IgkgdgJgphAuB5CAVATgTwOoEt4AsQAueNAVxgBoQAlGAezVjUQAc4ARGAYwBsBDNP3g56EYqQrUAyqRzd4AWXqwJ5GAF8gA) for the code
```fsharp
let y = async {
    return { 0.. 1 }
}
```
resulting in 
```fsharp
let y = async { return 0..1 }
```
Compiler complains: *This value is not a function and cannot be applied*

nojaf:
Fixes #553
Not at all sure if this is a good fix or not.
It doesn't break any other tests so might be ok 😅

### DIFF
--- a/src/Fantomas/CodePrinter.fs
+++ b/src/Fantomas/CodePrinter.fs
@@ -412,7 +412,7 @@ and genAttributes astContext (ats: SynAttributes) =
             let dontAddNewline =
                 TriviaHelpers.``has content after that ends with``
                     (fun t -> t.Range = a.Range)
-                    (function | Directive(_) -> true | _ -> false)
+                    (function | Directive(_) | Newline -> true | _ -> false)
                     ctx.Trivia
             let chain =
                 acc +>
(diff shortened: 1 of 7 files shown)


src/Fantomas/CodePrinter.fs
```diff
@@ -894,7 +894,7 @@ and genExpr astContext synExpr =
             +> ifElse isArrow (sepArrow +> preserveBreakNln astContext e2) (!- " do" +> indent +> sepNln +> genExpr astContext e2 +> unindent))
 
     | CompExpr(isArrayOrList, e) ->
-        let astContext = { astContext with IsNakedRange = true }
+        let astContext = { astContext with IsNakedRange = false }
```
jindraivanek: Maybe removing this line would be clearer and with same effect?

nojaf: Yep, it has the same effect.

Issue #642, opened 2020-01-14
Support check-style flag on fantomas-tool


lpedrosa:
### Summary

Add CLI support for check-style flag much like other code formatters out there. This would allow fantomas to run in a CI environment and check if the codebase is consistently formatted.

### Description

In a team setting, it is desirable to have a consistent codebase. Many other languages have formatters available, here are some examples:

* [gofmt](https://golang.org/cmd/gofmt/)
* [rustfmt](https://github.com/rust-lang/rustfmt)
* [black](https://github.com/psf/black)

All of these formatters have a way to specify that we are only interested in understanding if any of the files scanned requires formatting:

* rustfmt and black have a `--check` flag, which exits with a code other than 0, if any of the files requires formatting.
* gofmt has an `-l` flag, which outputs only the filenames of the files that require formatting.

Both these approaches can be used in a CI pipeline step, in order to keep formatting consistency in a codebase.

I believe the `FakeHelpers` module provides [a similar task](https://github.com/fsprojects/fantomas/blob/master/src/Fantomas/FakeHelpers.fs#L112-L129) if you are using FAKE.

Would it be possible to bring some of this logic to the CLI?

### Proposal

Consider a simple program with a folder structure:

```
MySolution.sln
src/
    MyProgram/
        Helpers.fs
        Program.fs
        MyProgram.fsproj
        
```

Assuming we forgot to run fantomas on our project, when we run it with the `--check` flag:

```
$ dotnet fantomas ./src/MyProgram/ --recurse --check
Files that need formatting:
src/MyProgram/Helpers.fs
src/MyProgram/Program.fs
```

The exit code could then be checked by any CI scripts and fail the build

```
$ echo $?
1
```

I am available to help with this, if you are interested.

nojaf:
Hello @lpedrosa, this sounds like a good idea but I'm not sure I want this in Fantomas just yet.
Maybe in the future when Fantomas is more stable, some projects now still have lots of issues after formatting. People might have false positives because the code doesn't format on their local machine anyway.

I would also like to see this in practice first to get a good feel of how this would work out in on a real project.

So my suggestion here it the following:
Could you create a proof of concept in a FAKE script? I would like to try this first in my own projects.
I believe all the building blocks to create something are there and the code for the POC could later maybe end up in Fantomas itself.

I hope you can understand my reasoning here, let me know what you think.

lpedrosa:
That makes sense, thanks for the detailed explanation!

[...]

PR #655, opened 2020-01-28
Check code after format

lpedrosa:
## Summary

Started adapting the CLI code so it uses the `FakeHelpers.formatFileAsync` as mentioned on #642 

I have tried to respect the previous functionality, e.g.

* When the file is unchanged it is still written to the provided `TextWriter`

A good side-effect of this changes is that we can now identify is a file is unchanged and opt not to do anything. 

I haven't yet included the `--check` flag, please see the **Notes** section below

Let me know what you think @nojaf.

## Notes

I also took some time to look through the project and gather some knowledge about the internals. Given that, I wrote down a couple of suggestions that require a discussion before I start slicing through the code:

* The `formatFileAsync` returns a DU which may create or not temp files in the background. I was wondering if it could just return the modified content string instead. Callers can then decide what to do with it, rather than having to read the modified contents from the temp file.

```fsharp
type FormatResult =
    | Formatted of file: string * content: string // as opposed to the temp file
    | Unchanged of file: string
    | Error of file:string * Exception
```

I also have a couple of questions:

* Have you considered testing the CLI? For example, I have noticed that the `--stdin` functionality doesn't work with this current version (I'll submit another PR to fix this).
* Is the CLI output stable, or can it be changed? For example:
  - When the CLI fails to format a file, it prints a massive stack trace. It could be more user friendly instead e.g. `error: Failed to format file <x>: <error message>`.
  - Unchanged files should probably still be printed to stdout


src/Fantomas/FakeHelpers.fs
```diff
@@ -67,6 +65,13 @@ let formatFileAsync config (file : string) =
         | ex -> return Error(file, ex)
     }
 
+let formatFileAsync config (file : string) =
+    let originalContent = File.ReadAllText file
+    async {
+        return! originalContent
```
nojaf: Nitpick: I've never seen this construct before, maybe write it out the long way.

nojaf:
Hey @lpedrosa , looking good so far!

I'm ok with changing `FormatResult`, however, the behaviour of the FAKE helper should stay the same. [formatCode](https://github.com/nojaf/trivia-tool/blob/master/build.fsx#L47) should not break.

Maybe mark `formatFilesAsync`,`formatFileAsync` and others internal so the public api of FakeHelpers is a bit more clear. You might need an `[<assembly:InternalsVisibleToAttribute("Fantomas.Tests")>]` for `Fantomas.CoreGlobalTool` as well.

> Have you considered testing the CLI? For example, I have noticed that the --stdin functionality doesn't work with this current version (I'll submit another PR to fix this).

No, I've never really considered this. But if there is an easy way (that can run in CI), by all means, go for it.

> Is the CLI output stable, or can it be changed? For example:
When the CLI fails to format a file, it prints a massive stack trace. It could be more user friendly instead e.g. error: Failed to format file <x>: <error message>.
Unchanged files should probably still be printed to stdout

There is room for improvement there, I'm ok with changes to that area.

> Unchanged files should probably still be printed to stdout

Yes, indeed.

src/Fantomas/FakeHelpers.fs
lpedrosa: Gotcha, I'll do it the long way around. 

`return!` just flattens an `Async<T>` result before exiting an `async` block, so you don't get an Async<Async<T>>.

lpedrosa:
Thanks for the review @nojaf. I will focus on wrapping up this PR, by eliminating the need for the temp files and adding the check functionality along with some tests for that.

With regards to changing the tool output, do you want this as a separate PR, since it will affect the release version number?

The CLI tests are something I would like to look into, but probably not in this PR.

I've submitted another PR (#656) that fixes the `--stdin` functionality.

nojaf:
> With regards to changing the tool output, do you want this as a separate PR, since it will affect the release version number?

It is ok to include in this PR.
To me, this is more of improvement rather than a breaking change. The next version would be 3.2 so that would cover this feature. Might be somewhat debatable but you shouldn't worry about that.

> I've submitted another PR (#656) that fixes the --stdin functionality.

Thanks!

Lastly, could you also update the [documentation](https://github.com/fsprojects/fantomas/blob/master/docs/Documentation.md) of the CLI?

nojaf:
Hello @lpedrosa , do you need any assistance to finish this PR?

lpedrosa:
I should be able to wrap this up today, thanks!

lpedrosa:
I have included the `--check` functionality in the CLI and have also updated the docs. 

The code can be cleaned up a bit e.g. move the check functionality to a separate module, test it, etc. 

I can do that afterwards, considering that you might want to release 3.2.x and I don't want to hold the release for that.

With regards to testing CLI modules or code, would you prefer another test project, or should I just reference the CLI project in the current test project.

src/Fantomas.CoreGlobalTool/Program.fs
```diff
@@ -96,24 +98,38 @@ let rec allFiles isRec path =
...
-                                                           FakeHelpers.createParsingOptionsFromFile inFile,
-                                                           FakeHelpers.sharedChecker.Value)
-        tw.Write(formatted)
+        let! formatted = inFile |> FakeHelpers.formatFileAsync config
+
+        match formatted with
+        | FakeHelpers.FormatResult.Formatted(_, formattedContent) ->
+            tw.Write(formattedContent)
+        | FakeHelpers.FormatResult.Unchanged(_) ->
+            inFile
```
nojaf: I know this is probably the current behaviour but does it make sense to write out the same file?

src/Fantomas.CoreGlobalTool/Program.fs
```diff
@@ -313,23 +331,90 @@ let main argv =
...
+                        function
+                        | FakeHelpers.FormatResult.Error(_) -> true
+                        | _ -> false
+
+                    let hasErrors =
+                        changes
+                        |> Seq.exists isError
+
+                    if hasErrors then
+                        return 99
```
nojaf: Any particular reason you went with 99? Just wondering though.

nojaf:
I released [3.2](https://github.com/fsprojects/fantomas/releases/tag/v3.2.0) yesterday actually.
The proposal is to release a first 3.3 beta once this PR is merged. That way you don't really have to wait that long to use `--check`. And we can gather some feedback before marking the API as stable.

If you are up for it I would prefer to clean up and write tests in this PR.

>  With regards to testing CLI modules or code, would you prefer another test project, or should I just reference the CLI project in the current test project.

I would go with a separate project. Keep using NUnit is that is possible. That makes the most sense to me.

Very happy with the progress you've made! Big up!

src/Fantomas.CoreGlobalTool/Program.fs
lpedrosa: Ahah. Not really. We could go with your lucky number. Black (the python formatter) chose 123.

src/Fantomas.CoreGlobalTool/Program.fs
lpedrosa: You are right. I think the reason it was done like this originally is because your `output file` could be different than your `input file` and they still wanted the unchanged files to be copied over.

I am okay with restricting this a bit, if you prefer, by identifying if the underlying TextWriter destination is the same as the source, to avoid unnecessary writes. 

Or I could give a pass to the CLI logic altogether and enumerate a bunch of use cases that it currently supports, but don't make much sense.

lpedrosa:
Thanks for the positive feedback @nojaf. I'm really enjoying helping out and getting my hands dirty with F#.

I will clean this up and add some tests then.

src/Fantomas.CoreGlobalTool/Program.fs
nojaf: I was secretly hoping it was [a Jay-Z reference](https://open.spotify.com/track/7IdFdRlCjUi6kkhbPoRfnw?si=Qb3yoi0zR2eAhQC2VDYpag) 😋

@jindraivanek do you have a lucky number? I'm ok with 99 or anything else that sense.

src/Fantomas.CoreGlobalTool/Program.fs
nojaf: Ah yes, indeed if the input and output path is different it should write the file even if no formatting was necessary.
In theory, you could also check if the formatted file differs from the output path and don't write if it isn't the case.

So if you have input path A and output path B, compare formatted A with B (if it exists) before writing to path B.

> Or I could give a pass to the CLI logic altogether and enumerate a bunch of use cases that it currently supports, but don't make much sense.

I would keep it simple: if no formatting was applied no file should be written.

lpedrosa:
I have added a test project for the CLI functionality. I have also refactored main logic of the check command to a separate module and tested that.

I'm just missing the check that @nojaf highlighted above where "if there are no changes and the output file is the same, it shouldn't write anything at all"

lpedrosa:
> I'm just missing the check that @nojaf highlighted above where "if there are no changes and the output file is the same, it shouldn't write anything at all"

Actually, I suggest we merge this PR in. I have an idea how to tackle the above and it will require refactoring some of the `processX` functions and friends. That will probably be very involved, and another PR will be easier to review.

src/Fantomas.CoreGlobalTool.Tests/CheckTests.fs
```diff
@@ -0,0 +1,75 @@
+namespace Fantomas.CoreGlobalTool.Tests
```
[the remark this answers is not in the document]
nojaf: Something like
```fsharp
let format () =
    use p = new Process()
    p.StartInfo.UseShellExecute <- false
    p.StartInfo.FileName <- @"C:\Users\nojaf\Projects\fantomas\src\Fantomas.CoreGlobalTool\bin\Debug\netcoreapp3.1\fantomas-tool.exe"
    p.StartInfo.Arguments <- "--check \"C:\\Users\\nojaf\\Projects\\fantomas\\src\\Fantomas\\CodeFormatter.fsx\""
    p.Start() |> ignore
    p.WaitForExit()
    printfn "exit was %i" p.ExitCode
```

nojaf:
Hey @lpedrosa, I've found some time to tackle some of the remarks I had.
Could you take a look at these latest changes I've made?

For the most part, I moved some things around so that the same functionality can be easily exposed via FAKEHelpers module. 
The main idea is that the caller of the code can decide what they do with the result.
In case the CLI is the caller, I've processed the output as you did. 

I hope you don't mind me tinkering with your PR, let me know what you think.

lpedrosa:
Hey @nojaf, I've been away this week. I'll have a look at these changes hopefully on Monday.

> I hope you don't mind me tinkering with your PR, let me know what you think

Feel free to thinker away :)

src/Fantomas.CoreGlobalTool.Tests/TestHelpers.fs
```diff
@@ -0,0 +1,39 @@
...
+
+    use p = new Process()
+    p.StartInfo.UseShellExecute <- false
+    p.StartInfo.FileName <- @"dotnet"
+    p.StartInfo.Arguments <- sprintf "%s %s" fantomasDll arguments
+    p.Start() |> ignore
+    p.WaitForExit()
+    p.ExitCode
+
+let checkCode file =
```
lpedrosa: What about moving `checkCode` to the corresponding test file where it is used and exposing the `runFantomasTool` (which looks pretty solid to expose as a test helper, kudos for that) for other tests?

src/Fantomas/FakeHelpers.fs
```diff
@@ -55,75 +53,94 @@ let formatFileAsync config (file : string) =
...
+                | Formatted(source, formatted) ->
+                    File.WriteAllText(source, formatted)
+                    Some source
+                | _ -> None)
 
+        return result
+    }
 
-let checkCode config files =
+type CheckResult =
```
lpedrosa: Haha! Nice one, it's like you read my mind with this type. This way it's easier to do reporting on errors/changes 😄

src/Fantomas.CoreGlobalTool.Tests/TestHelpers.fs
nojaf: You're right that makes more sense.
I'd typically do things like these when the next test suite would pop up.

nojaf:
Ok, I think we can merge this now.
Unless you think otherwise @lpedrosa?

lpedrosa:
Looks good to me! 🚀

PR #656, opened 2020-01-29
Restore --stdin flag functionality

lpedrosa:
## Summary

This PR fixes the `--stdin` functionality which seemed to not work when the new version started using `Argu` for parsing the CLI arguments.

I have also preserved the original functionality where stdin input can only be read if redirected through a pipe:

```
# this would work
$ echo 'open System;; let () = printf "Hello World"' | fantomas-tool.exe --stdin
open System

let () = printf "Hello World"

# this would not
fantomas-tool.exe --stdin
Input path is missing...
```

## Examples

Current master build:

```
$ echo 'open System;; let () = printf "Hello World"' | fantomas-tool.exe --stdin
ERROR: argument '--stdin' must be followed by <string>.
... omitted usage output

# when input is not redirected
$ fantomas-tool.exe --stdin
ERROR: argument '--stdin' must be followed by <string>.
... omitted usage output
```

Fantomas.CoreGlobalTool 3.1.0:

```
$ echo 'open System;; let () = printf "Hello World"' | dotnet fantomas --stdin
open System

let () = printf "Hello World"

# when input is not redirected
$ dotnet fantomas --stdin
Input path is missing...
```

This PR:

```
$ echo 'open System;; let () = printf "Hello World"' | fantomas-tool.exe --stdin
open System

let () = printf "Hello World"

# when input is not redirected
$ fantomas-tool.exe --stdin
Input path is missing...
```

## Notes

The stdIn reader functionality is currently line capped (to 2000 lines, see `StdLineLimit`), to avoid memory overflows. The default `Console.ReadLine()` character limit is 256 characters, so that takes care of memory overflows.

We could potentially relax that limit, depending on how people use the `--stdin` functionality i.e. I am assuming this can be used as an editor hook, to format code selections
.

### DIFF
--- a/src/Fantomas.CoreGlobalTool/Program.fs
+++ b/src/Fantomas.CoreGlobalTool/Program.fs
@@ -13,7 +13,7 @@ type Arguments =
     | [<Unique>] Force
     | [<Unique>] Profile
     | [<Unique>] Fsi of string
-    | [<Unique>] Stdin of string
+    | [<Unique>] Stdin
     | [<Unique>] Stdout
     | [<Unique>] Out of string
     | [<Unique>] Indent of int
@@ -41,7 +41,7 @@ with
             | Out _ -> "Give a valid path for files/folders. Files should have .fs, .fsx, .fsi, .ml or .mli extension only."
             | Profile -> "Print performance profiling information."
             | Fsi _ -> "Read F# source from stdin as F# signatures."
-            | Stdin _ -> "Read F# source from standard input."
+            | Stdin -> "Read F# source from standard input."
             | Stdout -> " Write the formatted source code to standard output."
             | Indent _ -> "Set number of spaces for indentation (default = 4). The value should be between 1 and 10."
             | PageWidth _ -> "Set the column where we break to new lines (default = 80). The value should be at least 60."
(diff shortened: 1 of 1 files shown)


src/Fantomas.CoreGlobalTool/Program.fs
```diff
@@ -148,16 +167,25 @@ let main argv =
...
-            let hasStdin = results.Contains<@ Arguments.Stdin @>
-            if hasStdin then
-                InputPath.StdIn input
-            elif Directory.Exists(input) then
-                InputPath.Folder input
-            elif File.Exists input && isFSharpFile input then
-                InputPath.File input
-            else
-                InputPath.Unspecified
+            let maybeInput = results.TryGetResult<@ Arguments.Input @>
```
nojaf: Interesting, by using `TryGetResult` the current workaround for version can be simplified I think.
Something for another PR but thanks for this insight.

nojaf:
> We could potentially relax that limit, depending on how people use the --stdin functionality i.e. I am assuming this can be used as an editor hook, to format code selections

For now, there aren't that many people using stdin so I would not take this into account just yet.
Editors will not use the cli to format code but the Fantomas dll are integrate it that way.

Thanks for this PR @lpedrosa !

PR #671, opened 2020-02-06
Initial changes for FCS 34.0.1

baronfel:
This release has a few new API changes:

* and! support
* slicing from end

I've tried to incorporate those changes, but I'm sure there are gaps.  The existing tests pass but I'm sure new cases will need to be added for these two new case.


src/Fantomas.Tests/FormattingPropertyTests.fs
```diff
@@ -171,8 +171,8 @@ and generateSynExpr size =
                 1, Gen.map2 (fun expr typ -> SynExpr.Upcast(expr, typ, zero)) genSubIdentExpr genSubSynType
                 1, Gen.map2 (fun expr typ -> SynExpr.Downcast(expr, typ, zero)) genSubIdentExpr genSubSynType
                 1, Gen.map2 (fun expr typ -> SynExpr.TypeTest(expr, typ, zero)) genSubIdentExpr genSubSynType
-                1, Gen.map2 (fun expr1 expr2 -> SynExpr.DotIndexedGet(expr1, [SynIndexerArg.One expr2], zero, zero)) genSubBasicExpr genSubBasicExpr
-                1, Gen.map3 (fun expr1 expr2 expr3 -> SynExpr.DotIndexedSet(expr1, [SynIndexerArg.One expr3], expr2, zero, zero, zero)) genSubBasicExpr genSubBasicExpr genSubBasicExpr
+                1, Gen.map2 (fun expr1 expr2 -> SynExpr.DotIndexedGet(expr1, [SynIndexerArg.One(expr2, false, zero)], zero, zero)) genSubBasicExpr genSubBasicExpr
```
baronfel: indexers are now tuples of `expr, isFromEnd : bool, range`, hence these changes

src/Fantomas/AstTransformer.fs
```diff
@@ -562,7 +562,7 @@ module private Ast =
              Properties = p []
              FsAstNode = synExpr
              Childs = [yield visitSynExpr expr]}
-        | SynExpr.LetOrUseBang(_,isUse,isFromSource,pat,rhsExpr,body,range) ->
+        | SynExpr.LetOrUseBang(_,isUse,isFromSource,pat,rhsExpr,andBangs,body,range) ->
```
baronfel: andBangs are a list of tuples, where the 4th and 5th items are the pattern and body expr, respectively.  so we iterate them in this order here

src/Fantomas/CodePrinter.fs
```diff
@@ -1401,9 +1401,12 @@ and genExpr astContext synExpr =
     | SynExpr.Set(e1,e2, _) ->
         addParenIfAutoNln e1 (genExpr astContext) -- sprintf " <- " +> genExpr astContext e2
 
-    | LetOrUseBang(isUse, p, e1, e2) ->
+    | LetOrUseBang(isUse, p, e1, ands, e2) ->
         atCurrentColumn (ifElse isUse (!- "use! ") (!- "let! ")
-            +> genPat astContext p -- " = " +> genExpr astContext e1 +> sepNln +> genExpr astContext e2)
+            +> genPat astContext p -- " = " +> genExpr astContext e1 +> sepNln
+            // TODO: use ands here
```
baronfel: this is the big gap: not sure how to write ands as of yet

src/Fantomas/CodePrinter.fs
```diff
@@ -1491,14 +1494,14 @@ and genInfixApps astContext (hasNewLine:bool) synExprs (ctx:Context) =
 /// Use in indexed set and get only
 and genIndexers astContext node =
     match node with
-    | Indexer(Pair(IndexedVar eo1, IndexedVar eo2)) :: es ->
+    | Indexer(Pair((IndexedVar eo1, e1FromEnd),(IndexedVar eo2, e2FromEnd))) :: es ->
```
baronfel: I'm not _using_ the from-end-ed-ness here, but I'm fairly certain it would need to result in a change to the indexer syntax

nojaf:
Hey Chet, many thanks for this PR.
Much appreciated!

I'm first gonna take care of some [other FCS impacted stuff](https://github.com/fsprojects/fantomas#upgrading-fsharpcompilerservice).
We heavily use our online debug tools to work on Fantomas, these would need the upgrade as well.
So, in short, it might take some time before we will merge this in.

Lastly, could you give me some concrete code samples of what we need to support?
A sample for both `and!` & `slicing from end` would be nice.

baronfel:
Sure thing @nojaf! Do you need help tackling the task list for doing the update? It'd be a good learning experience for me.

Regarding samples, I'll scrape through the RFCs and compiler test cases and see if we can port over some of those.

baronfel:
Also, there's a bit of a cyclic dependency for the ast-viewer project: it depends on AstTransformer.fs, but I've made changes to that file here. How do you normally resolve this conflict?

nojaf:
Yes, I realized a half hour ago that I'm part of the problem here.
You can't update ast-viewer without first updating Fantomas 😅.
Some time ago that was possible but not anymore it seems.
Hmm, we could create a fcs-34 branch first and then later switch, but I'm wondering if it is worth all the effort.

nojaf:
@baronfel I'm gonna merge this and we can update all the tools.
Please make issues with Fantomas online then with the new syntax.

PR #674, opened 2020-02-08
Performance optimizations

jindraivanek:
* WriterEvent list replaced with FIFO queue
* fix stack overlow in formatting AstTransformer.fs and CodePrinter.fs
* time for AstTransformer.fs decreased from > 2min to 22s


jindraivanek:
Other optimisations I had in mind turns out to be more complicated, so I guess we can merge this..

PR #688, opened 2020-02-17
Performance optimizations, part 2

jindraivanek:
* Current `WriterModel` is part of `Context` and updated immediately after adding `WriterEvent`
* Current column is part of `WriterModel`
* Instead of calling `dump` we inspect `WriterEvents`
* Some range eq calls optimizations (`range.Start`, `range.End` are expensive)

Time improvements:

AstTransformer.fs: 00:00:25.5178497 s -> 00:00:04.7649482 s

CodePrinter.fs: 00:02:06.5416571 s -> 00:00:27.8426108 s


src/Fantomas/Context.fs
```diff
@@ -589,26 +596,30 @@ let internal printContentBefore triviaNode =
 let internal printContentAfter triviaNode =
     col sepNone triviaNode.ContentAfter printTriviaContent
 
+let rangeStartEq (r1: range) (r2: range) = r1.StartLine = r2.StartLine && r1.StartColumn = r2.StartColumn
```
nojaf: Move to RangeHelpers? Keep the names though, there are better than the ones in RangeHelper ;)

jindraivanek: done

PR #715, opened 2020-03-06
Fix Setup .NET Core in GitHub Actions

nojaf:
A lot of builds are failing to setup .NET Core, I think this will fix this.


nojaf:
Ok this fixes it versus https://github.com/fsprojects/fantomas/runs/489645016

PR #721, opened 2020-03-13
Proof of concept for short expression check.

nojaf:
@jindraivanek this is a suggestion to approach formatting short expressions.
The main idea is that we try an expression similar to a dummy context but shortcut as soon as possible if we detect that the expression if too long or it is multiline.

```fsharp
let internal isShortExpression maxWidth (shortExpression: Context -> Context) (fallbackExpression) (ctx: Context) = ...
```

We can use the shortExpression writerEvents if it is indeed short, else we execute the fallback expression.
Please let me know what you think about this, and if you see any (major) performance concerns.

This PR is a draft, I would finish it once you validated the main concept.

Looping in @Smaug123 as well. 

---

Update:

- [X] Record instances
- [x] Record definitions
- [x] Anounymous record instances
- [x] Anounymous record types?
- [X] Array/Lists 
- ~~[ ] Revisit if/then/else~~
- [X] Infix operators
- [X] Tuples (**no setting for this for now**)
- [X] LetBindings
- ~~[ ] Remove `IsDummy`~~
- [x] Remove `multliline`

New settings:
- MaxInfixOperatorExpression: Num
- MaxRecordWidth: Num
- MaxArrayOrListWidth: Num
- MaxLetBindingWidth: Num


nojaf:
Related to issue https://github.com/fsprojects/fantomas/issues/697

nojaf:
@jindraivanek the build fails because I didn't take any other unit tests into account, please only look at the four new ones to see the impact of the code change.

src/Fantomas/Context.fs
```diff
@@ -149,6 +190,28 @@ let internal dumpAndContinue (ctx: Context) =
...
+        then fallbackExpression ctx
+        else
+            let shortExpressionEvents =
+                Seq.skip info.PreviousEventCount resultContext.WriterEvents
+                |> Queue.ofSeq
+            let events =
+                Queue.fold (fun acc event -> Queue.conj event acc) ctx.WriterEvents shortExpressionEvents
+
+            let model = WriterModel.updateAll shortExpressionEvents ctx.WriterModel
+            { ctx with WriterEvents = events; WriterModel = model }
```
jindraivanek: (potential perf problem)
unless I'm missing something this could be replaced by
```suggestion
            { resultContext with WriterModel = { resultContext.WriterModel with Mode = ctx.WriterModel.Mode } }
```

if that's not a case, then this should be quicker (`Queue.rev` is almost free here) :
```suggestion
            let shortExpressionEvents =
                resultContext.WriterEvents |> Queue.rev |> Seq.take (Queue.length resultContext.WriterEvents - info.PreviousEventCount) 
                |> Queue.ofSeq |> Queue.rev
            let events =
                Queue.fold (fun acc event -> Queue.conj event acc) ctx.WriterEvents shortExpressionEvents

            let model = WriterModel.updateAll shortExpressionEvents ctx.WriterModel
            { ctx with WriterEvents = events; WriterModel = model }
```

jindraivanek:
I like it, especially the idea of applying `shortExpression` only once. 👍

It looks like so similar to what `Dummy` is doing and shortcutting there also make sense, maybe idea of merging these modes into one (with more options) worth exploring?

src/Fantomas/Context.fs
nojaf: Good thinking, that is indeed a cheaper solution.

src/Fantomas/TriviaContext.fs
```diff
@@ -7,7 +7,7 @@ open Fantomas.Context
 open Fantomas.TriviaTypes
 
 let tok (range: range) (s: string) =
-    enterNodeToken range +> (!-s) +> leaveNodeToken range
+    enterNode range +> (!-s) +> leaveNode range
```
jindraivanek: I think better would be to make new function with this body, so we keep meaning of `tok` function as something dealing with Tokens.

nojaf: `tok` was only used in one place so that is why I just changed the implementation.
Made the new function you asked.

nojaf:
https://github.com/fsprojects/fantomas/pull/721/commits/3260b6345c027ae68443cacb151a333162f2efb0 and https://github.com/fsprojects/fantomas/pull/721/commits/f00e4967845be836a63487b2c89247ffaa113ac7 remove the `multiline` function.

This was heavily used to determine if module declarations, let bindings inside let bindings, member bindings and member definitions are multiline.
If those constructs are multiline they should have a blank before and after.

For example:
```fsharp
let someBinding =
    let a = 42 // short

    let b =
         someFunctionCall a 8
         |> anotherFunctionCallMakingAllOfThisMultiline
         |> foo

   let c = "short"
```

In `sepNlnBeforeMultilineConstruct` detection is done if the previous construct already was multiline. To avoid double blank lines between `a` and `b` in case they are both long.
The detection is being done via the writer events in order to avoid future newline checks.

If `a` were to be multiline then it should not have a blank line between `let someBinding =`. This is an general exception that applies when the first construct is multiline.

One thing to notice is that trivia can sometimes make a construct multiline.

nojaf:
`IsDummy` cannot fully be replaced by the new approach. There are still a couple of places where you need to check first, make some decisions and move forward.
If/Then/Else is one of those areas. I might revisit this another time.

This PR is ready for review.
@jindraivanek please check the last commits, just in case.

drhumlen:
🎉

PR #722, opened 2020-03-13
Use autoIndentNlnByFuture for let bang

### LINKED ISSUE #615: Newline after "bang" keywords in computation expressions.
Dear maintainers,

Recently updated my Rider which apparently included the new version of Fantomas.

I found this behavior to be very strange:

```f#
let f =
  async {
    // Without binding newline after assignment sign preserved, which is expected behavior
    let r =
      match 0 with
      | _ -> ()

    return r
  }

let f2 =
  async {
    // When binding, newline force-removed, which makes the whole expression
    // on the right side to be indented.
    let! r = match 0 with
             | _ -> () |> async.Return

    return r
  }
```

And it also true for all other "bang" methods in computation expressions (`return!`, `do!`, etc.)
Please note it's not exclusive to pattern matching on the right side. I just chose it as an example from my production code, but in fact, any code block gets indented.

You can see more examples [here](https://jindraivanek.gitlab.io/fantomas-ui/#?code=DYUwLgBAhhC8CwAoCEC2UwGMAWEAMEA7gJZjZIoA+EA+hALQB8EAFAJRJKiQBmcF0AM4BPAHaYIAbwEpuEAE79kKFeiy4CJMjJXU6TVh2Up54AK7zRCgQF9OiOTwBMSlFBHipO7gEIFrlRQ1HHwiUnJjXVoGZnYISmZ3MUwAOgAlc0t7QNMwCyt5W3tHAGYApM9pSNz8nx0gjBDNcPr46IM4hKFk9MzRIsQgA&config=N4IgkgdgJgphAuBlADgQwMYwHIFcC2IAXAEwC+QA).

While it doesn't look all that scary with examples above, production code has really weird looking indentation that's hard to work with.

I'm ready to contribute and adjust things above by myself, I just want some insights on why was this rule added in the first place? Should we make this behavior optional?

Thanks!

nojaf:
Fixes #615, @fahrenq and @drhumlen please provide some more example.
But I think this is definitely a step in the right direction.

### DIFF
--- a/src/Fantomas.Tests/ClassTests.fs
+++ b/src/Fantomas.Tests/ClassTests.fs
@@ -465,3 +465,36 @@ type A =
     [<Emit("b")>]
     abstract b: Unit -> string
 """
+
+[<Test>]
+let ``string parameter to inherited class, 720`` () =
+    formatSourceString false """type Child() =
+  inherit Parent ""
+"""  config
+    |> prepend newline
+    |> should equal """
+type Child() =
+    inherit Parent ""
+"""
+
+[<Test>]
+let ``float parameter to inherited class`` () =
+    formatSourceString false """type Child() =
+  inherit Parent 7.9
+"""  config
+    |> prepend newline
+    |> should equal """
+type Child() =
+    inherit Parent 7.9
+"""
+
+[<Test>]
+let ``unit parameter to inherited class`` () =
+    formatSourceString false """type Child() =
+  inherit Parent ()
+"""  config
+    |> prepend newline
+    |> should equal """
+type Child() =
+    inherit Parent()
+"""
--- a/src/Fantomas.Tests/ComputationExpressionTests.fs
+++ b/src/Fantomas.Tests/ComputationExpressionTests.fs
@@ -227,3 +227,33 @@ let f2 =
         return r
     }
 """
+
+[<Test>]
+let ``let bang, and bang should be formatted as regular let`` () =
+    formatSourceString false """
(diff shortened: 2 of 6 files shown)


jindraivanek:
looks good, but I think this should be extended also for `and!`. Failing test:

```
[<Test>]
let ``let bang, and bang should be formatted as regular let`` () =
    formatSourceString false """
let f2 =
  async {
    // When binding, newline force-removed, which makes the whole expression
    // on the right side to be indented.
    let! r = match 0 with
             | _ -> () |> async.Return
    and! s = match 0 with
             | _ -> () |> async.Return
    return r + s
  }
"""  config
    |> prepend newline
    |> should equal """
let f2 =
    async {
        // When binding, newline force-removed, which makes the whole expression
        // on the right side to be indented.
        let! r =
            match 0 with
            | _ -> () |> async.Return
        and! s =
            match 0 with
            | _ -> () |> async.Return
        return r + s
    }
"""
```

fahrenq:
Hey,

Sorry, don't have much time to dive into changes in PR, will just bring one more time that this behavior was observed with all "bang" methods, including `return!`, `do!`, `yield!` etc.

Thank you!

nojaf:
@jindraivanek, ready for review. `and!` is fixed.
@fahrenq and @drhumlen please create new issues for the other "bang" methods, no steps to reproduce (with [online tool](https://jindraivanek.gitlab.io/fantomas-ui/#?fantomas=preview)) = no service 😉

PR #738, opened 2020-03-30
Fix 'with' removed from record definition

isaevdan:


isaevdan:
Work is still in progress, I found some failing tests

PR #745, opened 2020-04-14
Fix Type for SynMatchClause.Clause in visitSynMatchClause

panesofglass:
As discovered by @Krzysztof-Cieslak [here](https://www.twitch.tv/k_cieslak).


nojaf:
Thanks!

PR #751, opened 2020-04-16
Revert "Restore ignored tests"

nojaf:
This reverts commit 4710748c49d8cbb38d0abd1b8661f0d79c8232fb.


nojaf:
Looks like this is not ok yet, will look into it later.

PR #771, opened 2020-04-19
Escape entire curl command as string

nojaf:
Initial attempt didn't work:
image

Trying to escape the entire curl command this time.


nojaf:
@jindraivanek thanks it works now.

PR #799, opened 2020-04-28
Respect byte-order mark.

nojaf:
This worked on my machine on Windows.
I'm not really an expert when it comes to these things, all suggestions to improve are welcome.

### DIFF
--- a/src/Fantomas.CoreGlobalTool.Tests/TestHelpers.fs
+++ b/src/Fantomas.CoreGlobalTool.Tests/TestHelpers.fs
@@ -24,7 +24,7 @@ type OutputFile internal () =
             if File.Exists(filename) then
                 File.Delete(filename)
 
-let private runFantomasTool arguments =
+let runFantomasTool arguments =
     let pwd = Path.GetDirectoryName(typeof<TemporaryFileCodeSample>.Assembly.Location)
     let configuration =
         #if DEBUG
@@ -49,10 +49,3 @@ let private runFantomasTool arguments =
 let checkCode file =
     let arguments = sprintf "--check \"%s\"" file
     runFantomasTool arguments
-
-let formatCode file =
-    runFantomasTool file
-
-let formatCodeToFile sourceFile destinationFile =
-    sprintf "--out %s %s" destinationFile sourceFile
-    |> runFantomasTool
--- a/src/Fantomas.CoreGlobalTool.Tests/TestHelpers.fs
+++ b/src/Fantomas.CoreGlobalTool.Tests/TestHelpers.fs
@@ -3,14 +3,26 @@ module Fantomas.CoreGlobalTool.Tests.TestHelpers
 open System
 open System.Diagnostics
 open System.IO
+open System.Text
 
-type TemporaryFileCodeSample internal (codeSnippet: string) =
+type TemporaryFileCodeSample internal (codeSnippet: string, ?hasByteOrderMark: bool) =
+    let hasByteOrderMark = defaultArg hasByteOrderMark false
     let filename = Path.Join(Path.GetTempPath(), Guid.NewGuid().ToString() + ".fs")
-    do File.WriteAllText(filename, codeSnippet)
+    do (if hasByteOrderMark
+        then File.WriteAllText(filename, codeSnippet, Encoding.UTF8)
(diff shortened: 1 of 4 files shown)


src/Fantomas.CoreGlobalTool.Tests/TestHelpers.fs
```diff
@@ -3,15 +3,27 @@ module Fantomas.CoreGlobalTool.Tests.TestHelpers
...
+    do (if hasByteOrderMark
+        then File.WriteAllText(filename, codeSnippet, Encoding.UTF8)
+        else File.WriteAllText(filename, codeSnippet))
 
     member _.Filename: string = filename
     interface IDisposable with
         member this.Dispose(): unit = File.Delete(filename)
 
+type OutputFile internal () =
+    let filename = Path.Join(Path.GetTempPath(), Guid.NewGuid().ToString() + ".fs")
```
Smaug123, line 20: This is required to write to the filesystem, isn't it, because of `formatCodeToFile`. I guess it would be a biiiig refactor to switch to `IFileSystem` throughout Fantomas (so that this could be done with an in-memory file system like `MockFileSystem`); probably not a job to be done now.

src/Fantomas.CoreGlobalTool.Tests/TestHelpers.fs
```diff
@@ -36,4 +48,11 @@ let private runFantomasTool arguments =
 
 let checkCode file =
     let arguments = sprintf "--check \"%s\"" file
-    runFantomasTool arguments
\ No newline at end of file
+    runFantomasTool arguments
+
+let formatCode file =
```
Smaug123, line 53: Is there a reason for not just using `runFantomasTool`, instead of defining a new function `formatCode`?

src/Fantomas.CoreGlobalTool/Program.fs
```diff
@@ -69,13 +70,27 @@ let rec allFiles isRec path =
     Directory.GetFiles(path, "*.*", searchOption)
     |> Seq.filter (fun f -> isFSharpFile f && not (isInExcludedDir f))
 
+let private hasByteOrderMark file =
+    if File.Exists(file) then
+        let preamble = Encoding.UTF8.GetPreamble()
```
Smaug123, line 75: This will do the wrong thing for e.g. UTF-32 encoded files. The best way I know of to discover the encoding of a file is https://stackoverflow.com/a/30393739/1462221 ; and if Fantomas wants to respect BOM then it'll have to carry around the *encoding*, rather than merely a flag noting whether a BOM is present.

src/Fantomas.CoreGlobalTool.Tests/TestHelpers.fs
nojaf: Yes, `runFantomasTool` does the real thing, spins up the cli tool in a new process and actually runs as it would from the command line. IFileSystem and MockFileSystem would be nicer but before February we didn't even have any unit tests of the GlobalCoreTool project so small steps I suppose.

src/Fantomas.CoreGlobalTool.Tests/TestHelpers.fs
nojaf: `runFantomasTool` is marked as `private`, so I exposed it with a new function.
Maybe a bit silly indeed.

nojaf:
I think it is a good thing to respect the BOM, the case about formatting outside the IDE seems like a valid one to have this.

FSharpSpec says:
> All input files are currently assumed to be encoded as UTF-8. 

See https://fsharp.org/specs/language-spec/4.1/FSharpSpec-4.1-latest.pdf#page=25
So I'm not sure we should support UTF16 and UT32 as well.
Any thoughts on this @jindraivanek?

@auduchinok how does Rider handle this? Can it work with UTF16 F# files for example?

Smaug123:
Huh - if the F# compiler is not guaranteed to handle non-UTF8 files then that's a very strong argument that we should at least *normalise* to UTF-8. Whether to support other encodings as input really boils down to "how hard will it be"; if it's very easy, then it would kind of be nice to receive an arbitrary .txt file with F# code in, and be able to format it. But that does seem low priority if it's easier to assume UTF-8.

PR #802, opened 2020-05-01
Determine upper or lowercase for SynExpr.DotIndexedGet

### LINKED ISSUE #798: Fantomas crash with evaluation of array member
I was unable to submit a bug using the online tool because "request too large". Testing 4.0.0-alpha, either 0.0.1 or 0.0.2, *.fs mode, default settings in the online tool.

```
namespace Foo

type T = { A : (unit -> unit) array }
module F =
  let f (a : T) =
    a.A.[0] ()
```

Output is a long stack trace starting with "System.Exception: cannot determine if synExpr DotIndexedGet... is uppercase or lowercase".

nojaf:
Fixes #798

### DIFF
--- a/src/Fantomas.Tests/ListTests.fs
+++ b/src/Fantomas.Tests/ListTests.fs
@@ -1548,3 +1548,22 @@ let d = list.[^1..] // 4,5
 let e = list.[^0..] // 5
 let f = list.[^2..^1] // 3,4
 """
+
+[<Test>]
+let ``calling indexed item in list, 798`` () =
+    formatSourceString false """namespace Foo
+
+type T = { A : (unit -> unit) array }
+module F =
+  let f (a : T) =
+    a.A.[0] ()
+"""  config
+    |> prepend newline
+    |> should equal """
+namespace Foo
+
+type T = { A: (unit -> unit) array }
+
+module F =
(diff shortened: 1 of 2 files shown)


nojaf:
The pattern match tries to determine if the [SynExpr](https://fsharp.github.io/FSharp.Compiler.Service/reference/fsharp-compiler-ast-synexpr.html) is upper- or lowercase.

In this case, the problem was that SynExpr.DotIndexedGet was passed to the function and was not yet supported.
Looking at the structure of the DotIndexedGet, its first value in the tuple is a SynExpr so we can the function recursively in this case.

```fsharp
DotIndexedGet
                       (LongIdent
                          (false,
                           LongIdentWithDots
                             ([a; A],[tmp.fsx (1,16--1,17) IsSynthetic=false]),
                           None,tmp.fsx (1,15--1,18) IsSynthetic=false),
                        [One
                           (Const
                              (Int32 0,tmp.fsx (1,20--1,21) IsSynthetic=false),
                            false,tmp.fsx (1,20--1,21) IsSynthetic=false)],
                        tmp.fsx (1,18--1,19) IsSynthetic=false,
                        tmp.fsx (1,15--1,22) IsSynthetic=false)
```
=> determine value of DotIndexedGet by checking SynExpr.LongIdent.

PR #808, opened 2020-05-04
Don't reformat negative single floating-point numbers

### LINKED ISSUE #785: Some floating-point numbers are changed
Issue created from [fantomas-ui](https://jindraivanek.gitlab.io/fantomas-ui/#?code=LYewJgrgNgpgBAOQsARjATgZwLACg+wAucAHnALxwC0AzAHQBMAjDQGYEzECeFcTdANmbtcROAC9eVfgFYRYgIa96zNkA)

The value of some floating-point numbers will change when formatting the code. It appears to only apply to negative floating-point numbers:

#### Code

```fsharp
module Numbers

let x = -3.213f
let y = 1.621f
let z = -1.5f
let a = 3.213f
```
            

#### Result

```fsharp
module Numbers

let x = -3.213000059f
let y = 1.621f
let z = -1.5f
let a = 3.213f

```
            
#### Options

Fantomas 3.2.0

| Name | Value |
| ---- | ----- |
| `IndentSpaceNum` | `4` |
| `PageWidth` | `120` |
| `StrictMode` | `false` |

ErikSchierboom:
Fixes #785

### DIFF
--- a/src/Fantomas/Trivia.fs
+++ b/src/Fantomas/Trivia.fs
@@ -118,6 +118,14 @@ let private findNodeAfterLineAndColumn (nodes: TriviaNode list) line column =
         (range.StartLine > line) || (range.StartLine = line && range.StartColumn > column)
     )
 
+let private findConstNodeOnLineAndColumn (nodes: TriviaNode list) line column =
+    nodes
+    |> List.tryFind (fun { Type = t; Range = r } ->
+        match t, line = r.StartLine, column = r.StartColumn with
+        | MainNode("SynExpr.Const"), true, true -> true
+        | _ -> false
+    )
+
 let private findConstNodeAfter (nodes: TriviaNode list) (range: range) =
     nodes
     |> List.tryFind (fun { Type = t; Range = r } ->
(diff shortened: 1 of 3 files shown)


src/Fantomas/Trivia.fs
```diff
@@ -325,7 +329,7 @@ let private addTriviaToTriviaNode (startOfSourceCode:int) (triviaNodes: TriviaNo
         |> updateTriviaNode (fun tn -> { tn with ContentItself = Some siNode }) triviaNodes
 
     | { Item = Number(_) as number; Range = range  } ->
-        findNodeOnLineAndColumn triviaNodes range.StartLine range.StartColumn
+        findFirstNodeOnLineAndColumn triviaNodes range.StartLine range.StartColumn
```
ErikSchierboom, line 332: I added the `findFirstNodeOnLineAndColumn ` function as searching backwards (which `findNodeOnLineAndColumn` does) will result in the match finding the `minus` token instead of the `const`, which both have the same start column but the minus token is the latter element in the list. I can fix this in other ways too, e.g. by filtering out the `minus` token before trying to search for the node. Let me know how you'd like to have this implemented.

nojaf: I would filtered out the minus token instead.
In the of numbers we actually know that we are always looking for a SynExpr.Const node type.
So I'd go for that.

ErikSchierboom: Okay! Will do that.

ErikSchierboom: I've gone with the approach of adding a `findConstNodeOnLineAndColumn` function, which is in line with having existing `findNodeOnLineAndColumn` and `findConstNodeAfter` functions. Is this okay, or would you rather do a `List.filter` within the `| { Item = Number(_) as number; Range = range  } ->` match?

ErikSchierboom:
@nojaf Is there something else that I should be doing? Just checking to see if I've missed anything :)

nojaf:
No I couldn't merge yet because the CI wasn't completed yet.
Many thanks for this PR, I will most likely release another alpha by the end of the week!

ErikSchierboom:
Brilliant! Thanks for the help :)

Issue #784, opened 2020-04-24
Adding newline before first comment in module


ErikSchierboom:
Issue created from [fantomas-ui](https://jindraivanek.gitlab.io/fantomas-ui/#?code=LYewJgrgNgpgBAIQIYGcCWBjFBYAUHgegLgBUB5AETIC44wYAzNAO3gBcALeAchgA8ADjAxsYYALIsIolAElmZAG4xm3OACMWYFgHM8hYuSq16TVnE48ATjGBIWuyc2kw5C5argMIzEWhDM+rhEpJQ0dIws7Fxw3AI2AkhWSGz+zCRowDDyTi4oat6+qQFBIUbhplEWMdxsIGxIUBlZOVIyBT5+JbhAA)

When formatting a module without any `let` bindings in it but only some comments, an additional newline is added before the comment. If I [add a let before the comment](https://jindraivanek.gitlab.io/fantomas-ui/#?code=LYewJgrgNgpgBAIQIYGcCWBjFBYAUH2AFzgA84BeOARjzjwHp64AVAeQBFWAuOMGAMzQA7eIQAW8AOQwSABxgZCMMAFlhEJSgCSQ1gDcYQyXABGwsMIDmeBkzacefQSLjipAJxjAkwq2qEaMNq6BkZw-BBCimggQja4jCwc3LwCwqIScJKynrJI7kiEMULMaMAwOv6BKMYRUUWx8Yn2KU7prpmShCCESFCl5ZXqmrWR0Y24QA) or [add a let between comments](https://jindraivanek.gitlab.io/fantomas-ui/#?code=LYewJgrgNgpgBAIQIYGcCWBjFBYAUHgegLgBUB5AETIC44wYAzNAO3gBcALeAchgA8ADjAxsYYALIsIolAElmZAG4xm3OACMWYFgHM8eWGzh84AXjgBGfbiKlKNOoxbsucbgCcYwJC12Tm0jByCsqqcAwQzCJoIMzWtuRUtPRMrHCcPAKeAkjuSGwxzCRowDDy-oEoahFRBbHxxIkOKc7prtxsIGxIUMWl5VIy1ZHR9bhAA), things work as expected.

#### Code

```fsharp
module Basics

// TODO: define the 'expectedMinutesInOven' binding

// TODO: define the 'remainingMinutesInOven' function

// TODO: define the 'preparationTimeInMinutes' function

// TODO: define the 'totalTimeInMinutes' function

```
            

#### Result

[...]

PR #809, opened 2020-05-04
Don't add newline to module where the first trivia is a newline

ErikSchierboom:
I've taken a stab at fixing #784, but I'd like to check if I'm heading in the right direction first. The problem is that the `sepModuleAndFirstDecl` checks for any declaration. If there is no such declaration, two newlines are added after the module declaration. As comments are no declarations, this means that the comment will also receive two newlines. The main problem is that running format one more time after that, will add another newlines (there will be three at this point), and so on.

Let me know if I'm heading in the right direction with my solution.

### DIFF
--- a/src/Fantomas.Tests/ModuleTests.fs
+++ b/src/Fantomas.Tests/ModuleTests.fs
@@ -439,10 +439,9 @@ type UrlModel =
       Defines: string }
 """
 
-[<TestCase(true)>]
-[<TestCase(false)>]
-let ``comment is first trivia in module should not add newline, 784`` (isFsiFile) =
-    formatSourceString isFsiFile """
+[<Test>]
+let ``comment is first trivia in module should not add newline, 784`` () =
+    formatSourceString false """
 module foo
 
 // bar
@@ -451,6 +450,54 @@ module foo
     |> prepend newline
     |> should equal """
 module foo
+
+// bar
+// baz
+"""
+
+[<Test>]
+let ``comment is first trivia in module in signature file should not add newline, 784`` () =
+    formatSourceString true """
+module foo
+
+// bar
+// baz
+"""  config
+    |> prepend newline
+    |> should equal """
+module foo
+
+// bar
+// baz
+"""
+
+[<Test>]
+let ``comment is first trivia in namespace should not add newline, 784`` () =
+    formatSourceString false """
+namespace foo.quz
+
+// bar
+// baz
+"""  config
+    |> prepend newline
+    |> should equal """
+namespace foo.quz
+
+// bar
+// baz
+"""
+
+[<Test>]
+let ``comment is first trivia in namespace in signature file should not add newline, 784`` () =
+    formatSourceString true """
+namespace foo.quz
+
+// bar
+// baz
+"""  config
+    |> prepend newline
+    |> should equal """
+namespace foo.quz
 
 // bar
 // baz
--- a/src/Fantomas.Tests/ModuleTests.fs
+++ b/src/Fantomas.Tests/ModuleTests.fs
@@ -438,3 +438,20 @@ type UrlModel =
       KeepNewlineAfter: bool
       Defines: string }
 """
+
+[<TestCase(true)>]
+[<TestCase(false)>]
+let ``comment is first trivia in module should not add newline, 784`` (isFsiFile) =
+    formatSourceString isFsiFile """
+module foo
+
+// bar
+// baz
+"""  config
+    |> prepend newline
+    |> should equal """
+module foo
+
+// bar
+// baz
+"""
--- a/src/Fantomas/Context.fs
+++ b/src/Fantomas/Context.fs
@@ -891,6 +891,14 @@ let internal sepNlnForEmptyModule (moduleRange:range) ctx =
     | _ ->
         sepNln ctx
 
+let internal sepNlnForEmptyNamespace (namespaceRange:range) ctx =
+    let emptyNamespaceRange = mkRange namespaceRange.FileName (mkPos 0 0) namespaceRange.End
+    match TriviaHelpers.findInRange ctx.Trivia emptyNamespaceRange with
+    | Some node when hasPrintableContent node.ContentBefore || hasPrintableContent node.ContentAfter ->
+        ctx
+    | _ ->
+        sepNln ctx
+
 let internal sepNlnTypeAndMembers (firstMemberRange: range option) ctx =
     match firstMemberRange with
     | Some range when (ctx.Config.NewlineBetweenTypeDefinitionAndMembers) ->
--- a/src/Fantomas/Context.fs
+++ b/src/Fantomas/Context.fs
@@ -883,6 +883,13 @@ let internal sepNlnConsideringTriviaContentBeforeWithAttributes (ownRange:range)
     |> Seq.exists (fun ({ ContentBefore = contentBefore }) -> hasPrintableContent contentBefore)
     |> fun hasContentBefore ->
         if hasContentBefore then ctx else sepNln ctx
+        
+let internal sepNlnForEmptyModule (moduleRange:range) ctx =    
+    match findTriviaMainNodeOrTokenOnStartFromRange ctx.Trivia moduleRange with
+    | Some node when hasPrintableContent node.ContentBefore || hasPrintableContent node.ContentAfter ->
+        ctx
+    | _ ->
+        sepNln ctx
 
 let internal sepNlnTypeAndMembers (firstMemberRange: range option) ctx =
     match firstMemberRange with
--- a/src/Fantomas/Trivia.fs
+++ b/src/Fantomas/Trivia.fs
@@ -146,9 +146,9 @@ let private mapNodeToTriviaNode (node: Node) =
     )
     
 let private commentIsAfterLastTriviaNode (triviaNodes: TriviaNode list) (range: range) =
-    triviaNodes
-    |> List.filter isMainNodeButNotAnonModule
-    |> List.forall (fun tn -> tn.Range.EndLine < range.StartLine)
+    match List.filter isMainNodeButNotAnonModule triviaNodes with
+    | [{ Type = MainNode("SynModuleOrNamespaceSig.NamedModule") }] -> true
+    | mainNodes -> mainNodes |> List.forall (fun tn -> tn.Range.EndLine < range.StartLine)
 
 let private updateTriviaNode lens (triviaNodes: TriviaNode list) triviaNode =
     match triviaNode with
--- a/src/Fantomas.Tests/ClassTests.fs
+++ b/src/Fantomas.Tests/ClassTests.fs
@@ -541,3 +541,16 @@ let ``unit parameter to inherited class`` () =
 type Child() =
     inherit Parent()
 """
+
+[<Test>]
+let ``class let binding with attribute should not add newline, 786`` () =
+    formatSourceString false """type A() =
+
+    [<Bar>]
+    let foo = ()
+"""  config
+    |> should equal """type A() =
+
+    [<Bar>]
+    let foo = ()
+"""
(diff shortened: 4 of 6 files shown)


src/Fantomas/Context.fs
```diff
@@ -883,6 +883,18 @@ let internal sepNlnConsideringTriviaContentBeforeWithAttributes (ownRange:range)
     |> Seq.exists (fun ({ ContentBefore = contentBefore }) -> hasPrintableContent contentBefore)
     |> fun hasContentBefore ->
         if hasContentBefore then ctx else sepNln ctx
+        
+let internal sepNlnConsideringTriviaContentBeforeAndAfter (moduleRange:range) ctx =    
+    match findTriviaMainNodeOrTokenOnStartFromRange ctx.Trivia moduleRange with
+    | Some node ->
+        if hasPrintableContent node.ContentBefore then
```
nojaf, line 890: Can be simplified to
```fsharp
        if hasPrintableContent node.ContentBefore || hasPrintableContent node.ContentAfter then
            ctx
        else
            sepNln ctx
```

Because I think this won't work if you have
```
module foo
// bar
// meh
```

nojaf:
@ErikSchierboom you are in the right direction here.
Maybe change `sepNlnConsideringTriviaContentBeforeAndAfter` to `sepNlnForEmptyModule` because I don't think this edge case can happen anywhere else at first glance.

Please also fix this for signature files, as [the bug](https://fsprojects.github.io/fantomas-tools/#/fantomas/preview?data=N4KABGBEDGD2AmBTSAuKBbBBXANosAYrLADoBOJAduVQPS1jqIAWkANOFAM6IAuvAS0oBzLqjABtThFAQ5UACS8AngAdkaSEN7tpcyAoBuAQxxYNkvfIAMHefMgBJSkkq8AyquPREAOSzouvZyACxWYAC6egC%2BdnKy9gYq6uJabkGJRqbm4lLBEACMccGQAArGwogA6gLwvKzF9gUATNbhUfKxegkOSmoWkABGxDgZvSZmFnn5zY0O7ojoAnA4sJQAgrwAoi4A8gBmADJCyHNy%2B6Y87TGNPfp9KZrDsKNnBhM5aNPBAMxnEJBPN5EAAhRD7WBkRDlMjGJi8RBkMb2XhkczXTq3cJJfqpZ6vbFZSa5cIQEL-KBAnxgiFQw6wADuiOgxh4zkMsBZgjWyPkqPRwQ6ci68juAIeA3xvMUHympLAAFYKYCvNTwZDEABVVTqMgstmUDlcgQ8ikXHBXQU3bqE5KSkbS97ZOX5MAANmVVNB6qhAGEcKyuL61lx%2BdBeJDpRBzZb7EKICL4rbcU8HW8iZ9LK6AOye1Xe2mIACyi0GiKjYBjiAxwqxJQleLThNlJNdAA488CaRrg6tKBWqzWE3XMnbGy9HRmXfkAJydnzrfYIsjB9DoYwV-nVq2Ym31sepifpltfeUFWzylVdn2IBZLFam%2BWDne1vejlNQKXH52t-IFIqXl6i7LneywvI%2BrpbkOYCJjIyaPJ%2BTb1ieWZ-rMl7OK4vC7JQAAqZDKDU9QDpc25xtaorwfaR7Nj%2Bp6ugUfyAfm6xkLAWAuAAIogOACEsy6bmiZHyPGMEjr0B5QNok4od8TTkpeRbGAAHo4%2By4cwiCUFsFq3swkK8DUdQNPKIRtC%2Bw5vhJH5pDo37EvRf5KopKnOPsAjKbsurGBGZBbMpqhQlwXAmv2FIKuZ5G7pR%2B42dJ9mZnJ8gFB6LnKQASogcBkPARnERSZnQbBYBiookm2TJdGocEBS5mlrGwsouxkMcoZ5SZrqFRZYlWfc5XxbRDnVU0HZpYcfAgkI8BCMI7UVl1UWvjF74IUMSGZLJZ5zopuCCLxlCgqs0AANYgrCJ18FwOHuHCiC9gEYVPqRRXiX1NlfoNiXyq0yq%2BIgDL7d6vBMlpuH9Nx7mUAI3IbC4JboGWZBiGaz3dcVpU4qtH3IVVSVyM0AGuoCqLLLwRYIKcT26UVnARHEWhcAQIXiFB0RAA) appears there as well.

You current fix will also not work for namespaces I think.
Consider [this example](https://fsprojects.github.io/fantomas-tools/#/fantomas/preview?data=N4KABGBEDGD2AmBTSAuKA7AhgW0QZwAdNpEwAxWWAHQCcr1b6B6JsXAC0gBpwo9EALgICW6AOZ5UYANq8IoCIqgASAQE8CyNJFEDucxZGUA3TABsArlpkGlABh5KlkAJLok6AQGUiJAHIW2PpOigAstmAAugYAvo6KCk5G6ppSOp7BSSbmVlKyIRAAjPEhkAAKmGKIAOrC8AKcJU6FAEx2EdFKcQaJzqoa1pAARpRmmX2mltb5BS1Nzl6I2MJwZrDoAIICAKLuAPIAZgAyosjzigfm-B2xTb2G-anaI7Bj50aTuWgzIQDM5xBID5iIgAEKIA6wGiICo0HCCRA0cZOAQ0Kw3Lp3CLJAZpF5vbHZKZ5CIQUIAqDAkjgyHQo6wADuiOgmH4bmMsBZInWyKUqPRIU6im6SnugMeg3xvJUn2mpLAAFYKUDfGCIVDEABVAiaGgstnoDlc4Q8imXMzXQW3HqElKS0bSj45OUFMAANmVVLVtMQAGEzKy8L71nh%2BdABFDpRBzZanEKICKErbcc8He8iV8bK6AOye1U0jUAWSWQ0RUbAMcQGOFWNKErxacJspJroAHHmQQXocG1uhy5Xqwna1k7Q3Xo6My6CgBODskDYHASI4PYbCYcv8qtWzE2uuj1Pj9PN77ywoOeUqzvq6GLZarU3ygfbmu7kcpqBSo-OlsFQrFC9eguS40LeKyvA%2BrqboOYCJvIyZPB%2BjZ1seWa-nMF5uB4Ah7OgAAqNBqLUDT9lcW5xtaorwfah5Nt%2BJ6uoU-wAaqGw0LAFjuAAIogZjCMswEbmiZFKPGMHDn0%2B5QLoE4oT8zTkhehaYAAHi4By4ewiDoNsFqIF47BQgItT1I08qhO0z5Dq%2BEnvukehfsS9G-kqikqW4BzCMpey6pgEY0NsykENCeB4CafYUgqFnkTulF7rZ0kOZmclKIUHqucpABKiBwDQ8DGcRFLmdBsFgGKKiSXZMl0ahISFLm6WsXCah7DQJyhvlpmukVllidZDwVQltGOTVzTtulRyCKCojwKIYgdeW3XRS%2BsVvghwxIVksmnrOikWGYIi8egYJrNAADWoJwmdgh4DhXjwj2gThY%2BpHFeJ-W2Z%2BQ1JfKbTKn4iAModaoCEyWm4QM3Eeegwjcps7jFtgpY0JIZovT1JVlTia2fch1XJYoLT-q6QKoisAiFggZzPbpxW8JE8Q6HgZChVIA4xEAA), notice that the namespace has a weird range in the AST.

```
ImplFile
  (ParsedImplFileInput
     ("tmp.fsx",true,QualifiedNameOfFile Tmp$fsx,[],[],
      [SynModuleOrNamespace
         ([Foo],false,DeclaredNamespace,[],
          PreXmlDoc ((1,9),FSharp.Compiler.XmlDoc+XmlDocCollector),[],None,
          tmp.fsx (3,6--3,6) IsSynthetic=false)],(true, true)))
```

The comment is linked after the ident instead there:
image

ErikSchierboom:
Alright thanks! I'll work on this some more.

src/Fantomas.Tests/ModuleTests.fs
```diff
@@ -437,4 +437,21 @@ type UrlModel =
     { IsFsi: bool
       KeepNewlineAfter: bool
       Defines: string }
-"""
\ No newline at end of file
+"""
+
+[<TestCase(true)>]
```
nojaf, line 442: Nitpick: I'd prefer a not using the `TestCase` attribute but duplicate the test to SignatureTests.fs.
To remain consistent with the rest of the codebase and to easier spot regressions.
This is clear now but it will be confusing that a test fails in ModuleTests while changing something in Signature AST for example.

ErikSchierboom: Will do!

src/Fantomas/Trivia.fs
```diff
@@ -145,9 +145,9 @@ let private mapNodeToTriviaNode (node: Node) =
     )
     
 let private commentIsAfterLastTriviaNode (triviaNodes: TriviaNode list) (range: range) =
-    triviaNodes
-    |> List.filter isMainNodeButNotAnonModule
-    |> List.forall (fun tn -> tn.Range.EndLine < range.StartLine)
+    match List.filter isMainNodeButNotAnonModule triviaNodes with
+    | [{ Type = MainNode("SynModuleOrNamespaceSig.NamedModule") }] -> true
```
ErikSchierboom, line 149: I needed this for comments in an empty module in a signature file, which has the named module being matched as a trivia node but which end line is equal to or greater than the comment

src/Fantomas.Tests/ModuleTests.fs
```diff
@@ -437,4 +437,68 @@ type UrlModel =
...
+    |> prepend newline
+    |> should equal """
+module foo
+
+// bar
+// baz
+"""
+
+[<Test>]
+let ``comment is first trivia in module in script file should not add newline, 784`` () =
```
nojaf, line 459: You mean signature files here right?

src/Fantomas/Context.fs
```diff
@@ -749,6 +749,12 @@ let private findTriviaMainNodeOrTokenOnEndFromRange nodes (range:range) =
     |> List.tryFind(fun n ->
         Trivia.isMainNode n && RangeHelpers.rangeEq n.Range range
         || Trivia.isToken n && RangeHelpers.rangeEndEq n.Range range)
+    
+let private findTriviaMainNodeOrTokenContainedInRange nodes (range:range) =
```
nojaf, line 753: I would move this function to TriviaHelpers.fs

src/Fantomas/Trivia.fs
nojaf: I checked this, and I don't see any other way around this.
Looks good.

nojaf:
Hey @ErikSchierboom! Very readable code, many thanks for this PR.

ErikSchierboom:
Thanks for the review! I'll apply your suggestions (tomorrow).

src/Fantomas.Tests/ModuleTests.fs
ErikSchierboom: I do. I've fixed this.

src/Fantomas/Context.fs
ErikSchierboom: Ah, I thought it should be added here due to `Context.fs` already having `findTriviaMainNodeOrTokenOnStartFromRange` and `findTriviaMainNodeFromRange` functions. But I can move it to `TriviaHelpers.fs`. Would you like it to be renamed?

nojaf: Yeah TriviaHelpers came after we already have some stuff in Context.fs
Not everything was perfectly moved to TriviaHelpers.
Please move and rename your function, we can move the other functions in a cleanup PR.

[TriviaNodeType ](https://github.com/fsprojects/fantomas/blob/master/src/Fantomas/TriviaTypes.fs#L49 )can only be two things so the `Trivia.isMainNode` and `Trivia.isToken` can also be left out I think.

ErikSchierboom: Alright, done! I've renamed the function to `findInRange` to better fit with the existing `findByRange` function in `TriviaHelpers.fs`. Let me know if that is okay or if you' prefer a different name.

Issue #819, opened 2020-05-08
multiline let bang should have a newline


nojaf:
[...]

PR #820, opened 2020-05-09
Add newline after multiline let bang

nojaf:
Fix for #819 
Fantomas will now check if `let!` and each `and!` is multiline and then decide if there should be an extra newline or not.

Suggestions to improve readability are welcome as it might be a bit hard to read at first glance.


src/Fantomas/CodePrinter.fs
```diff
@@ -1570,16 +1570,29 @@ and genExpr astContext synExpr =
...
-        ifElse isUse (!- "use! ") (!- "let! ") +> genPat astContext p -- " = "
-        +> autoIndentAndNlnIfExpressionExceedsPageWidth (genExpr astContext e1)
-        +> sepNln
-        +> genAndList astContext ands
-        +> genExpr astContext e2
+        let genLetBang =
+            ifElse isUse (!- "use! ") (!- "let! ") +> genPat astContext p -- " = "
+            +> autoIndentAndNlnIfExpressionExceedsPageWidth (genExpr astContext e1)
+
+        let genAnd astContext ((_,_,_,pat,expr,_): (DebugPointForBinding * bool * bool * SynPat * SynExpr * range)) =
```
jindraivanek, line 1577:
```suggestion
        let genAnd astContext (_, _, _, pat: SynPat, expr: SynExpr, _) =
```

PR #822, opened 2020-05-10
Update config settings in schema & docs

knocte:
I noticed this when receiving lots of warnings when trying to use
these settings with Fantomas 4 alpha.


docs/Documentation.md
```diff
@@ -252,22 +252,20 @@ A default configuration file would look like
...
   "SpaceBeforeClassConstructor":false,
+  "SpaceBeforeMember":false,
   "SpaceBeforeColon":false,
   "SpaceAfterComma":true ,
   "SpaceBeforeSemicolon": false,
   "SpaceAfterSemicolon":true ,
   "IndentOnTryWith":false,
   "SpaceAroundDelimiter":true ,
   "MaxIfThenElseShortWidth":40,
+  "MaxRecordWidth":40,
```
nojaf, line 267: This list is incomplete, [MaxArrayOrListWidth](https://github.com/fsprojects/fantomas/blob/master/src/Fantomas/FormatConfig.fs#L57) is missing for example.

knocte: oops, done!

knocte:
> Could you also update the schema.json while you are at it?

done!

Issue #825, opened 2020-05-11
Inconsistent if-then-else cut


knocte:
[...]

Issue #725, opened 2020-03-17
Document new features in v4


nojaf:
Since there are multiple moving pieces for the v4 release it seems easier for me to update the documentation once the final API is known.
If PR like #649 and #706 I've added, removed and renamed settings numerous times so I'd like to tackle everything documentation-wise in this issue.

PR #826, opened 2020-05-11
Setting to always keep if ... then in the same line

Bobface:
Implements a setting for #825 

The setting is called `KeepIfThenInSameLine` and defaults to `false`. If set to `true` the `then` in an `if` statement will be kept in the same line even if the line exceeds the maximum width.

Example (also implemented as test case):
```
type TransferAmount(valueToSend: decimal, balanceAtTheMomentOfSending: decimal) =
    do
        if balanceAtTheMomentOfSending < valueToSend then
            invalidArg "balanceAtTheMomentOfSending"
                "some very very long error message"
        if valueToSend <= 0m then
            invalidArg "valueToSend" "Amount has to be above zero"
```

This code will stay not change by formatting if `KeepIfThenInSameLine` is set to `true`. Otherwise, the `then` in the second `if` statement would be written to a new line.


nojaf:
Hey @Bobface, I really appreciate that you are willing to contribute to the project.
Please note that our [contribution guides](https://github.com/fsprojects/fantomas#before-touching-the-code) say: `please discuss before making a PR.`

So I would appreciate a heads up when PR like this pops up. As #825 was only created five hours ago, I did not have the time to validate whether this is a bug or not.

I don't have time to review this properly right now. It first glance one test for a new settings seems a bit on the low side. Would it not make sense to let this setting apply to IfThenElse as well?
Did you check the F# style guide?

In a perfect world we would have an RFC style of approach to discuss new settings.
We don't have this right now, so I'd like to confirm that we thought this through before merging.

Bobface:
Hi @nojaf, sorry I was not aware that we should discuss changes before making a PR. I will make sure to do that for future PRs. 

> Would it not make sense to let this setting apply to IfThenElse as well?
The changes apply to the `IfThenElse` 

> It first glance one test for a new settings seems a bit on the low side
I wasn't sure which other tests to add since it is a relatively small change. Do you have suggestions on what exactly you would like to have tested?

nojaf:
Hey @Bobface, no worries. I'm just a bit itchy when people make PR before approval.
They bring back some bad memories and I prefer a good conversation before introducing new settings.

> I wasn't sure which other tests to add since it is a relatively small change

It could use some additional test with comments before each expression and/or keyword.

> The changes apply to the IfThenElse

Your setting only has an impact on code that doesn't have an else branch.
Those rules a bit more complex, [see CodePrinter](https://github.com/fsprojects/fantomas/blob/master/src/Fantomas/CodePrinter.fs#L1266).

So this leads me to question: what is the scope of this setting? Maybe we want this for else cases as well?

[The F# style guide](https://docs.microsoft.com/en-us/dotnet/fsharp/style-guide/formatting#formatting-if-expressions) doesn't really touch upon scenarios that don't have an else branch. So one could wonder if we need to keep the same rules here? The most likely answer would yes, but it leaves room for interpretation.

In the case of #825, the rules do seem to be applied correctly.
Both the `if` and the `then` expression is longer than the threshold (MaxIfThenElseShortWidth  = 0) but are not multiline.

image

I understand that you might not like this style but I encourage you to first discuss this with the F# style guide. Maybe they agree and update the guide to always format longer if/then expressions on two lines. Then we don't need the setting at all.

So, to summary please contact the F# style guide regarding the if/then without else.
And consider the option to have the setting also applied for scenarios with an else statement.

knocte:
Hey Florian,

> Both the if and the then expression is longer than the threshold (MaxIfThenElseShortWidth = 0) but are not multiline.

Why is this? I was presuming that to achieve always multiline if-then-else blocks, MaxIfThenElseShortWidth can be simply set to 0, upon your past comment here: https://github.com/fsprojects/fantomas/pull/557#issuecomment-560272243 . So when I was testing that, I found the inconsistency which I filed as bug #825.

I agree this PR needs more tests to add if blocks with else elements. I think this setting (or simply fixing the inconsistency) would help achieving the G-Research style for non-short if-then-else blocks: https://github.com/G-Research/fsharp-formatting-conventions#formatting-if-expressions

nojaf:
Hey @knocte

> Why is this? I was presuming that to achieve always multiline if-then-else blocks, MaxIfThenElseShortWidth can be simply set to 0, upon your past comment here: #557 (comment) .

Sorry that comment isn't correct if I think about it.
The GR style seems to always have the `if` and `then` keyword on the same line.
And they never go for a 
```fsharp
if x
then y
else z
```
construction.

And I agree we could use a setting to achieve the GR behavior. I would just like the setting to also have an impact on the elif/else branches.

So if `KeepIfThenInSameLine = true` and the code is the following:
```fsharp
if a then b else c
```
Where a,b or c is larger than MaxIfThenElseShortWidth.
Then it should be
```fsharp
if a then
    b
else
    c
```

Other example:
if a,b,c,d, e are shorter than MaxIfThenElseShortWidth
it should be
```fsharp
if a then b
elif c then d
else e 
```

Hopefully, all the above can be achieved by revisiting the logic around [isAnyExpressionIsLongerButNotMultiline](https://github.com/fsprojects/fantomas/blob/master/src/Fantomas/CodePrinter.fs#L1392).

@Bobface are you willing to further investigate this? I can guide you along the way as that part of the code can be somewhat tricky.

Bobface:
@nojaf Yes I will further investigate and implement it to behave as you have described in the above examples. Any tips would be much appreciated!

knocte:
> So if KeepIfThenInSameLine = true and the code is the following:
> if a then b else c
> Where a,b or c is larger than MaxIfThenElseShortWidth.
> Then it should be
> if a then
>    b
> else
>    c

Sounds good, this above would be useful for both GR and me, I guess.

src/Fantomas.Tests/IfThenElseTests.fs
```diff
@@ -915,4 +915,157 @@ let ``comment after then in if/then, 730`` () =
...
+        if valueToSend <= 0m then // comment
+            invalidArg "valueToSend" "Amount has to be above zero" // comment
+"""  c
+    |> prepend newline
+    |> should equal """
+type TransferAmount(valueToSend: decimal, balanceAtTheMomentOfSending: decimal) =
+    do
+        // comment
+        if balanceAtTheMomentOfSending < valueToSend then // comment
+            invalidArg "balanceAtTheMomentOfSending"  // comment
```
Bobface, line 970: I noticed an additional whitespace is added between the code and the comment. I am not sure if this is desired behaviour and it seems to be unrelated to the changes I have introduced.

Bobface:
Implemented the `KeepIfThenInSameLine` for `else if` and `else` and added a few tests including comments.

src/Fantomas.Tests/IfThenElseTests.fs
nojaf: I would imagine this is unrelated. We can follow this up in a separate issue.

src/Fantomas.Tests/IfThenElseTests.fs
```diff
@@ -947,4 +944,128 @@ type TransferAmount(valueToSend: decimal, balanceAtTheMomentOfSending: decimal)
...
+                "some very very long error message" // comment
+        if valueToSend <= 0m then // comment
+            invalidArg "valueToSend" "Amount has to be above zero" // comment
+"""
+
+[<Test>]
+let ``keep then in same line as if, else if, else, 825`` () =
+
+    let c = { config with
```
nojaf, line 979: Could you move all these tests to a new file KeepIfThenInSameLineTests.
There we can override the config once and keep the test DRY.
Similar to what I did in [NewlineBetweenTypeDefinitionAndMembersTests](https://github.com/fsprojects/fantomas/blob/master/src/Fantomas.Tests/NewlineBetweenTypeDefinitionAndMembersTests.fs#L7)

src/Fantomas.Tests/KeepIfThenInSameLineTests.fs
```diff
@@ -0,0 +1,133 @@
+module Fantomas.Tests.KeepIfThenInSameLineTests
+
```
nojaf, line 2: @Bobface could you also add a test for when the MaxIfThenElseShortWidth is not exceeded.

Something like:
```fsharp
// Multiple conditionals with elif and else are indented at the same scope as the if:
if cond1 then e1
elif cond2 then e2
elif cond3 then e3
else e4
```
but every expression is shorter than MaxIfThenElseShortWidth.

src/Fantomas/FormatConfig.fs
```diff
@@ -57,6 +58,7 @@ type FormatConfig =
           MaxArrayOrListWidth = 40
           MaxLetBindingWidth = 40
           MultilineBlockBracketsOnSameColumn = false
+          KeepIfThenInSameLine = false
```
nojaf, line 61: Please also mention the new setting in the [documentation ](https://github.com/fsprojects/fantomas/blob/master/docs/Documentation.md#--config-path-to-file-or-folder) and [schema.json](https://github.com/fsprojects/fantomas/blob/master/src/Fantomas/schema.json).

No need to write proper documentation. I'll tackle this #725 .
Just make sure the setting is mentioned there so I won't forget.

src/Fantomas.Tests/KeepIfThenInSameLineTests.fs
```diff
@@ -131,3 +131,25 @@ else
...
+[<Test>]
+let ``MaxIfThenElseShortWidth not exceeded, 825`` () =
+    let c = { config with
+               MaxIfThenElseShortWidth = 100 }
+               
+    formatSourceString false """if foo then
+    bar
+else if fooo then
+    bar
+else if foooo then
```
nojaf, line 144: change this one to `elif` please. So both cases are covered.
This may seem like a nitpick now, but completeness is key to detect regressions later.

Issue #848, opened 2020-05-20
Setting for moving body of function to next line?


knocte:
I was a happy user of `--keepNewlineAfter`, and now that it's been removed, I feel there should be a way, at least, to tell fantomas to always start the body of a function (or a member, in case the function is a method of a class) always in a new line. In my view, it improves readability to only let constants (and not functions) to be declared and assigned in the same line. This way by a quick glance you can most of the time know if you're looking at a constant or at a function.

Could a setting be created for this? (I'm not sure about the best name for it though, `functionBodyToStartAfterNewLine`?) If yes, I will contribute it.

nojaf:
Hey @knocte, I understand your sentiment here and I feel a setting could be ok here.
However, home work first ;)
Could you identify a code example for all different cases where the setting would have impact.
Please come up with an excessive list of examples. Think about subtle differences like signature files versus implementation files. Top level let bindings versus let bindings within let bindings.

nojaf:
Once we have identified the cases, maybe the preferred behavior can also be achieved by introducing some settings that control the max length of expression.

[...]

PR #865, opened 2020-05-25
Split MaxLetBindingWidth into two separate settings

Bobface:
Implements #848 as discussed.

The `MaxLetBindingWidth` has been split into two separate settings:

- `MaxValueBindingWidth`: Handles values like `let someValue = 1`
- `MaxFunctionBindingWidth`: Handles function definitions like `let someFunction someParam = someParam + 1`

Schema and docs are updated for the new settings. Existing test cases are also updated to work with the new settings. However, I did not add new tests since, in my opinion, the existing tests should cover these new settings. If desired new tests can also be added.

### DIFF
--- a/src/Fantomas/CodePrinter.fs
+++ b/src/Fantomas/CodePrinter.fs
@@ -37,6 +37,8 @@ type ASTContext =
       /// Check whether the context is inside a SynMemberDefn.Member(memberDefn,range)
       /// This is required to correctly detect the setting SpaceBeforeMember
       IsMemberDefinition: bool
+      /// Check whether the context is inside a let binding which defines a function
+      IsInsideFunctionBinding: bool
     }
     static member Default =
         { TopLevelModuleName = ""
@@ -44,7 +46,7 @@ type ASTContext =
           IsCStylePattern = false; IsNakedRange = false
           HasVerticalBar = false; IsUnionField = false
           IsFirstTypeParam = false; IsInsideDotGet = false
-          IsMemberDefinition = false }
+          IsMemberDefinition = false; IsInsideFunctionBinding = false }
 
 let rec addSpaceBeforeParensInFunCall functionOrMethod arg (ctx:Context) =
     match functionOrMethod, arg with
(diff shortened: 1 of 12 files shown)


knocte:
This is missing testcases for members.

src/Fantomas/CodePrinter.fs
```diff
@@ -37,14 +37,16 @@ type ASTContext =
...
+      IsInsideFunctionBinding: bool
     }
     static member Default =
         { TopLevelModuleName = ""
           IsFirstChild = false; InterfaceRange = None
           IsCStylePattern = false; IsNakedRange = false
           HasVerticalBar = false; IsUnionField = false
           IsFirstTypeParam = false; IsInsideDotGet = false
-          IsMemberDefinition = false }
+          IsMemberDefinition = false; IsInsideFunctionBinding = false }
```
nojaf, line 49: Please only add things to the ASTContext as a last resort.
In this case, you get the information from SynPat, which is available in genExprSepEqPrependType.

src/Fantomas/CodePrinter.fs
```diff
@@ -470,6 +472,11 @@ and genExprSepEqPrependType astContext (pat:SynPat) (e: SynExpr) (valInfo:SynVal
                     (indent +> sepNln +> !- "=" +> unindent) ctx
         else
             (sepEq +> sepSpace)
+   
+    let maxBindingWidth = 
+        match astContext.IsInsideFunctionBinding with
```
nojaf, line 477: Please use if/then/else

src/Fantomas/CodePrinter.fs
```diff
@@ -540,6 +547,11 @@ and genLetBinding astContext pref b =
             | _ ->
                 genPat astContext p
 
+        let isFunctionBinding =
```
nojaf, line 550: Move helper function to SourceParser.

src/Fantomas/CodePrinter.fs
```diff
@@ -540,6 +547,11 @@ and genLetBinding astContext pref b =
...
+            match p with
```
nojaf, line 551: I'm also not sure if everything is cover with this pattern.
An active pattern seems to have a different [SynPat](https://fsharp.github.io/FSharp.Compiler.Service/reference/fsharp-compiler-syntaxtree-synpat.html).

F.ex. `let (|Add|_|) = Some bar`

nojaf: Destructured tuples is another case: `let (a,b) = foo`

Bobface:
Dropping in favor of #866

PR #866, opened 2020-05-26
MaxValueBindingWidth and MaxFunctionBindingWidth settings

Bobface:
This PR will implement the change from a single `MaxLetBindingWidth` setting to two separate settings `MaxValueBindingWidth` and `MaxFunctionBindingWidth` in two steps:

1. Rename `MaxLetBindingWidth` setting to `MaxBindingWidth`
2. Split `MaxBindingWidth` into two separate settings `MaxValueBindingWidth` and `MaxFunctionBindingWidth`

The new settings while handle these situations:
- `MaxValueBindingWidth`: Handles values like `let someValue = 1`
- `MaxFunctionBindingWidth`: Handles function definitions like `let someFunction someParam = someParam + 1`


Bobface:
@nojaf The first phase is done. Please have a look. This has been done so far:

- Rename `MaxLetBindingWidth` to `MaxBindingWidth`
- Make `member`s respect `MaxBindingWidth` instead of only `PageWidth`
- Update exisiting test cases which have changed because `member`s now respect `MaxBindingWidth`

When we get this stage to an acceptable level I will continue by splitting `MaxBindingWidth` into two separate settings `MaxValueBindingWidth` and `MaxFunctionBindingWidth`

src/Fantomas.Tests/AttributeTests.fs
```diff
@@ -16,7 +16,8 @@ type Funcs =
     |> should equal """[<Extension>]
 type Funcs =
     [<Extension>]
-    static member ToFunc(f: Action<_, _, _>) = Func<_, _, _, _>(fun a b c -> f.Invoke(a, b, c))
+    static member ToFunc(f: Action<_, _, _>) =
```
nojaf, line 19: Please don't change the original test, try and tinker with the settings until the existing behaviour is met.
This might not always be possible but the initial attempt should be to leave existing tests as they are.

src/Fantomas/Context.fs
```diff
@@ -567,21 +567,21 @@ let internal leadingExpressionIsMultiline leadingExpression continuationExpressi
     let hasWriteLineEventsAfterExpression = contextAfterLeading.WriterEvents |>Seq.skip eventCountBeforeExpression |> Seq.exists (function | WriteLine _ -> true | _ -> false)
     continuationExpression hasWriteLineEventsAfterExpression contextAfterLeading
 
-let private expressionExceedsPageWidth beforeShort afterShort beforeLong afterLong expr (ctx: Context) =
+let private expressionExceedsWidth width ignoreShort beforeShort afterShort beforeLong afterLong expr (ctx: Context) =
     // if the context is already inside a ShortExpression mode, we should try the shortExpression in this case.
     match ctx.WriterModel.Mode with
-    | ShortExpression _ ->
+    | ShortExpression _ when not ignoreShort ->
```
nojaf, line 573: What's the deal with this `ignoreShort`?

src/Fantomas.Tests/AttributeTests.fs
Bobface: Just to be sure: You do not want any of the existing tests to change? 
The reason I had to change a lot of tests is because the `member`s now respect `MaxBindingWidth` which added line breaks in some of the tests. 
The default value for `MaxBindingWidth` is 40, as was before with `MaxLetBindingWidth`. If the tests are not supposed to change I can either
- change the default value of `MaxBindingWidth` to something higher, like `80`, or
- use a custom `config` for all tests which would have changed with an increased `MaxBindingWidth`

Which of the two options would you prefer?

src/Fantomas/Context.fs
Bobface: Used in `Context.fs` (notice `true` as second parameter to `expressionExceedsWidth`):
```
let internal sepSpaceOrIndentAndNlnIfExpressionExceedsMaxBindingWidthIgnoreShort expr (ctx:Context) =
    expressionExceedsWidth
        ctx.Config.MaxBindingWidth true
        sepSpace sepNone // before and after for short expressions
        (indent +> sepNln) unindent // before and after for long expressions
        expr ctx
```

Which is used by `CodePrinter.fs` in `genMemberBinding`:
```
        match e with
        | TypedExpr(Typed, e, t) ->
            genAttributesAndXmlDoc
            +> leadingExpressionIsMultiline prefix (fun prefixIsLong ->
                ifElse
                    prefixIsLong
                    (indent +> sepNln +> sepColon +> genType astContext false t +> sepNln +> !- "=" +> sepNln +> genExpr astContext e +> unindent)
                    (sepColon +> genType astContext false t +> sepEq +> sepSpaceOrIndentAndNlnIfExpressionExceedsMaxBindingWidthIgnoreShort (genExpr astContext e)))
        | e ->
            genAttributesAndXmlDoc
            +> leadingExpressionIsMultiline prefix
                (fun prefixIsLong ->
                    ifElse
                        prefixIsLong
                        (indent +> sepNln +> !- "=" +> sepNln +> genExpr astContext e +> unindent)
                        (sepEq +> sepSpaceOrIndentAndNlnIfExpressionExceedsMaxBindingWidthIgnoreShort (genExpr astContext e)))
```

The `Context` seems to be in `ShortExpression` mode when generating member bindings which results in `MaxBindingWidth` being ignored because `expressionExceedsWidth` will match `ShortExpression`.

src/Fantomas.Tests/AttributeTests.fs
nojaf: So I prefer you change the config for each test individually to match the previous result.

Bobface: Done.

knocte:
@nojaf is this good so far? if yes, @Bobface will implement the rest

nojaf:
@knocte I still need to check on @Bobface last remark.
I would like to debug this to see if we really need the `ignoreShort` change.

nojaf:
So I took a look and tried to confirm my initial suspicions: this PR has more code changes than needed.
As an experiment, I copied the new unit tests and renamed the setting to see how the same result could have been achieved with [fewer changes](https://github.com/fsprojects/fantomas/compare/master...nojaf:max-value-binding-width).

Two things to note:
* There still might be some room for refactoring the right abstractions to Context.
* I haven't updated the Typed MemberBinding (deliberately) and all tests still pass so extra tests are needed there. Try to do this in a TDD style.

@Bobface could you revisit this PR based on my findings? 
As for the ShortExpression mode, this can be confusing from time to time when debugging but you shouldn't disable it. It will eventually pass through and try the fallback expressions. The thing is you can have nested expressions that both rely on ShortExpression mode, so that might have been something you encountered.

Bobface:
@nojaf I've updated the changes to use your implementation for the CodePrinter. The changes to the Context have been removed.

nojaf:
@Bobface could you add a test for when the member binding is typed, please.

Bobface:
@nojaf Done.

nojaf:
Hey @Bobface thanks for adding the test.
I think we can move forward to splitting the setting into MaxValue and MaxBinding if that is still the plan for this PR.

Bobface:
@nojaf I've split the setting into two.

There is currently one failing test:
https://github.com/fsprojects/fantomas/pull/866/files#diff-22a481f935212a16b8e8933b52bdd9c4L32-L61

The reason is that the active pattern is not identified as function by the helper:
https://github.com/fsprojects/fantomas/pull/866/files#diff-23c516460c0edb3213f811dc4261fc48R1408-R1411
```
let isFunctionBinding (p: SynPat) =
    match p with
    | PatLongIdent(_, _, ps, _) when (List.length ps > 0) -> true
    | _ -> false 
```

Maybe you could give me a hint how I can make that helper match active patterns as functions?

nojaf:
@Bobface on top my head I think you can get arround it with some regex magic.
Take this [example](https://fsprojects.github.io/fantomas-tools/#/ast?data=N4KABGBEAmCmBmBLAdrAzpAXFSAacUiaAYmolmPAIYA2as%2BEkAxgPZwWQ2wAuYAFAB8A8jwAWsAE6CA%2BoICUYALyUArsmY9ErZGEFgAyqwC2sMAA8wAWgB8hk2cv6AcjrO2wr1JBABfIA).
If the SynPat looks like
```fsharp
Named
                    (Wild tmp.fsx (2,5--2,14) IsSynthetic=false,|Other|_|,false,
                     None,tmp.fsx (2,5--2,14) IsSynthetic=false),
```

Maybe do a regex to matches `|something|_|`?

Or maybe try and deduce the information from SynBinding ->  SynValData -> SynValInfo -> SynArgInfo. See [docs](http://fsharp.github.io/FSharp.Compiler.Service/reference/fsharp-compiler-syntaxtree.html) for more info.

Bobface:
@nojaf Nevermind. `isFunctionBinding` actually detected the active pattern as function. The problem was that in the failing test `MaxFunctionBindingWidth` was set to `120`. I've updated it to `30` and the test passes now.

src/Fantomas/SourceParser.fs
```diff
@@ -1403,4 +1403,9 @@ let rec (|UppercaseSynExpr|LowercaseSynExpr|) (synExpr:SynExpr) =
 let rec isEmptySynSimplePats (ssp:SynSimplePats) =
```
nojaf, line 1403: I noticed this function is not being used anywhere. I know this has nothing to do with your PR but would you mind removing it?

Bobface:
@nojaf Thanks! I have removed the unused helper function, it did not have anything to do with this PR.

nojaf:
@Bobface I will most likely release a new alpha by the end of the week.

PR #872, opened 2020-05-29
Create CODE_OF_CONDUCT.md

c-rindi:
Adding a Code of Conduct to the Fantomas project.
Let us know if there is any feedback.


nojaf:
Thanks @c-rindi! This looks good, as is it written: it is a living document so I think we can always tweak the document afterwards if necessary.

PR #877, opened 2020-05-29
Correctly handle use binding follow by let bang

### LINKED ISSUE #876: Stackoverflow problem with let bang in match
The online tool can't get a response back because the following code leads to a StackOverflow:

```fsharp
    let private getAST log (req: HttpRequest) =
        async {
            use stream = new StreamReader(req.Body)
            let! json = stream.ReadToEndAsync() |> Async.AwaitTask
            let parseRequest = Decoders.decodeInputRequest json

            match parseRequest with
            | Result.Ok input when (input.SourceCode.Length < Const.sourceSizeLimit) ->
                let! astResult = parseAST log input
                match astResult with
                | Result.Ok ast ->
                    let node =
                        match ast with
                        | ParsedInput.ImplFile (ParsedImplFileInput.ParsedImplFileInput (_, _, _, _, hds, mns, _)) ->
                            Fantomas.AstTransformer.astToNode hds mns

                        | ParsedInput.SigFile (ParsedSigFileInput.ParsedSigFileInput (_, _, _, _, mns)) ->
                            Fantomas.AstTransformer.sigAstToNode mns
                        |> Encoders.nodeEncoder

                    let responseJson =
                        Encoders.encodeResponse node (sprintf "%A" ast)
                        |> Thoth.Json.Net.Encode.toString 2

                    return sendJson responseJson

                | Error error -> return sendBadRequest (sprintf "%A" error)

            | Result.Ok _ -> return sendTooLargeError ()

            | Error err -> return sendInternalError (sprintf "%A" err)
        }
```

All the default settings, most likely a regression after #840

nojaf:
Fixes #876


src/Fantomas/CodePrinter.fs
```diff
@@ -963,8 +963,10 @@ and genExpr astContext synExpr =
 
         let genCompExprStatement astContext ces =
             match ces with
-            | LetStatement(isRecursive, binding) ->
-                let prefix = if isRecursive then "let rec " else "let "
+            | LetOrUseStatement(isRecursive, isUse, binding) ->
+                let prefix =
+                    sprintf "%s%s" (if isUse then "use " else "let ") (if isRecursive then "rec" else "")
```
JackMatusiewiczGR, line 968: There's a slight change here, the trailing whitespace has been dropped. I think this would be flagged if the test had a recursive use / let case

Issue #867, opened 2020-05-26
Add unit tests/performance tests.


7sharp9:
Im finding that formatting takes ~4 seconds for an 800 line file it would be great to know the areas that are slowing things down so this could be improved.  Im finding myself not wanting to use fantomas as I know I have to wait every time I use it.  Similar tools in Elm,Go or Rust are instantaneous and it would be great to know where to aim improvements in this library.

nojaf:
Hey @7sharp9, yes Fantomas isn't that fast when the files become larger.
I'm don't really have a lot of expertise in finding this kind of issues so I'm open for suggestions here.

7sharp9:
I was surprised the 800 lines took 4 seconds, most formatter don't have a perceivable latency.  Benchmark.net is good for analysing but the functions that comprise fantomas would have to be tested in sections to see which parts are slow.

jindraivanek:
Biggest performance issue right now is Trivia collecting and querying. https://github.com/fsprojects/fantomas/blob/master/src/Fantomas/Trivia.fs

7sharp9:
```
    let triviaNodesFromAST =
        flattenNodeToList node
        |> filterNodes
        |> List.map mapNodeToTriviaNode
        |> List.choose id
```
The filter, map, choose  can be combined to avoid extra iterations, as choose is the map+filter optimisation.  Not sure if thats a slow bit though :-)

jindraivanek:
Slowest thing is probably that we use `list` to store `TriviaNode`s, and go through this list for every Triva query. Change it to something like `dict<LineNumber, TriviaNode list>` could be big perf win. Something more sophisticated like [Segment tree](https://en.wikipedia.org/wiki/Segment_tree) may be better, but maybe it is overkill.

7sharp9:
First thing I guess is to get a baseline with benchmark.net

nojaf:
Ok, I'll try to set up a baseline with benchmark.net.

gerardtoconnor:
If performance is a concern, you should restructure using of list piping in all the private query functions and potentially use an optimised tree structure as suggested already. Using streaming source/sink pattern is more performant but cannot use compiler hack on list creation head first

gerardtoconnor:
Profiling with Perfview will show where time is being spent, what is allocating heavily, just collect, run, stop and find process in profile export

7sharp9:
Isn't preview a windows tool?  I mean, who uses windows?

[...]

PR #878, opened 2020-05-29
Create initial benchmark.

nojaf:
Following up on #867, a first attempt to create a benchmark for the largest file in Fantomas.

So this works on my machine ™, how do we want to run this on CI?
It took around 30 min to run so I'm not sure we want to have this for each PR build.

image


jindraivanek:
> works on my machine ™

Interesting, I got
`System.IO.DirectoryNotFoundException: Could not find a part of the path 'C:\dev\github\fantomas\src\Fantomas.Benchmarks\bin\Release\netcoreapp3.1\c1cfcd54-1773-4652-b944-12f39eec3430\bin\Release\netcoreapp3.1\paket-files\fsprojects\fantomas\src\Fantomas\CodePrinter.fs'.`

`let path = __SOURCE_DIRECTORY__ + "/../../paket-files/fsprojects/fantomas/src/Fantomas/CodePrinter.fs"` fixes it.

nojaf:
> works on my machine ™

@jindraivanek, hmm yeah that might have been a lie there 🙈

I'm thinking about pushing the results of the benchmark to Azure, in the Application Insights.
That might be an easy way to dump the result and query the progress.

baronfel:
Regarding the overall speed of running the tests, I'd say the most impactful things to do would be to lower the amount of warming up/pre-work measurement that's done. Benchmark.net actually does a ton of additional work before running your tests to determine what sorts of jitter/constant factors to account for in your tests. Then, once the warmup is done it does your tests a large number of times.

Therefore you can speed up the iterations by 
* reducing the warmup count
* reducing the number of times your tests are run to get averages.

7sharp9:
I think you can tweak the attributes so that not as many tests are done so the CI would run faster.  the memory one does do a lot of iterations.

nojaf:
CI now runs between nine and elven minutes. Results are stored in Azure Table Storage.
I'm merging to master to see if the data gets pushed (Github actions doesn't allow usage of secrets on PR).

PR #891, opened 2020-06-05
Try online link updated to new URL.

jasondown:
Updated the _try online_ link to the new URL (on fsprojects.github.io).


jindraivanek:
Thanks!

PR #899, opened 2020-06-08
Update to FCS36 (and the SynType.Paren AST node :horror:)

baronfel:
This one has some second-order effects.  The addition of the simple branch to `genType` for wrapping the innerType of a `SynType.Paren` has got several tests failling due to double-parens. This _should_ just mean that the addition of parens in those special-cased situations needs to be removed, but that hasn't proven the case just yet.

### DIFF
--- a/src/Fantomas/CodePrinter.fs
+++ b/src/Fantomas/CodePrinter.fs
@@ -2396,9 +2396,7 @@ and genType astContext outerBracket t =
             fun (ctx:Context) ->
                 isShortExpression ctx.Config.MaxRecordWidth shortExpression longExpression ctx
         | TParen(innerT) ->
-            sepOpenT +>
-            loop innerT
-            +> sepCloseT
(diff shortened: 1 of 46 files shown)


src/Fantomas/CodePrinter.fs
```diff
@@ -2395,7 +2395,10 @@ and genType astContext outerBracket t =
 
             fun (ctx:Context) ->
                 isShortExpression ctx.Config.MaxRecordWidth shortExpression longExpression ctx
-
+        | TParen(innerT) ->
+            sepOpenT +>
```
baronfel, line 2399: in this case, to print a `TParen`, we render the parens around the rendering of the inner SynType.

nojaf:
@baronfel, I released a new alpha with this: 
https://www.nuget.org/packages/Fantomas/4.0.0-alpha-009

baronfel:
Thanks for finishing this up :)

PR #900, opened 2020-06-08
by-pass enterNode and leaveNode when no trivia are left

nojaf:
This won't massively improve anything but it is again a logic check to have.


nojaf:
Results of the Benchmark are a bit disappointing.

```
| Method |    Mean |   Error |  StdDev | Rank |       Gen 0 |       Gen 1 |      Gen 2 | Allocated |
|------- |--------:|--------:|--------:|-----:|------------:|------------:|-----------:|----------:|
| Format | 24.77 s | 18.38 s | 1.008 s |    1 | 762000.0000 | 225000.0000 | 28000.0000 |  13.09 GB |
```

The Error column increased significantly, so I wonder how accurate the result is.

PR #929, opened 2020-06-24
Split all parameters of a function application over multiple lines if they exceed page width

### LINKED ISSUE #907: When cutting off function invocations, should place each param in its own line (or align them to the 1st param)
Issue created from [fantomas-online](https://fsprojects.github.io/fantomas-tools/#/fantomas/preview?data=N4KABGBEDGD2AmBTSAuKBbBBXANosAwgIbQAWAlgHYDmYAvADqUQQAuAngA74CyRVxMogAUASnpMWUsOkToARogBOYAPqqAdACVErJeUQA3RABkiAZ1YFY6TrCyV4AISI4ilaPmFF48JYnNzNAAFLHkccmgAQV9-QNFJaSTklNS04WgsJX8PdjQCLJzodlE0ADlYVgAxONIAHiRocnRXAD4JZjSwHFhoAGswaBJSRCryPHMNQRH4Mt0Ad1glPoARIlYiMGEAMwc1MABaVsSulhbWMjB5V3dPMHnyVlIT07AAHzAK1ijDfjdw-BHT6VH5-IgAl6nD7TRDwLbXNweRAAGjArGaiHER0hrwg5G2gxsdgczhuSLAdTAAAYNFT0GiRpQcbipDo7EpWMElLAAfS4LZ7I4XIi7hVKPhMtlELkwD4-AFzGBzAryLBKDDZgslqt1kRma8xPqpAlOulRJBkeAoMrWOiaOZUGAANpWiCgJKQAAkHG4jsgVFYFtdLC9vxwWGQaBdKSplpSkAAko5pawAMqcEiIMpYdBBlIAFmDEAAusGAL5xlju6Ren2RqABvM1z1hiOO6PJACMlY9wSI1EQAHVyPAnk2kgAOKlFsClqQV4PVqS1rj1yDyWCwHDjkMt1xtqMziAAJh7NdTckiW7VUVYAFFHAB5bYmKjIM9SbauZUzucsBdSEuu51n6G5bjuEChvu9YdkkADMH4hummZOIg2xLIgfZKEQsisMoEEsHoEa-uWZ5AZB3qrqBm7bohUHhjBR5gPmiGQchnioeh-gmLA8zKEMypJoYvTrKqlAEWwSjEUkf4QABVYziuvpoOuNEEfRB7OkxACsrFQOxiCcRhACqnDcEoAmIEJInomqElgF%2BOA-jJpGLoplHKVAYG0e5raMSkABsemQAZRn%2BAQbiBNYlCWFJ0CsEs9mOc50iyWA8luu5IEqd56l7gx7ZMQA7MFoVoRhPByIoShJd%2BiAkfOZFZVROVqXR%2BWabB0gTqVGYceV4XXuJenJfVLmNW5HoeWuuXtX5hUpAAnL1mZRNseFKNY6AtPZRFjalrmAc1nmqeBc3QQtXaxkxIV9YZA2IBe6BXj0w1MaNDX-k1U3ZV5bW%2BRdh4pJ23Y3QZa0bU9L12Xpe2fXJ33Nr9p0%2BVN81A12p43UmSCUKwj6UAAKko7DDmOI11fD6WI8u03UWdAMFRjSSdghYN3VE3IkisiARM9G27VJ%2B1SGlGVgORUB0ypjbnUzWnAyxN18AAHgm2yE4yd5OY9pBLKww6jqQ9n5tO41fZNSMtQ2eN5ej8tdrpStEKrlDbOQyuPuZ6xLHeyucHE5hifZ2mmwdE1HT9Vv%2BjbsudUxnZBU7ys6HASjwAb5NMSbVNixLSlrjLjNx8DJVJ5z2HsI%2BSivpYGdG3p2dmwjFu08jhdo4D9ssz1ScAGrQU4VDwFQ1B18boci4dCmRyd7fNnbXVSJ2y1J1UDjxWJg%2BOCPY96RPLCizTwFR7NRf%2Bckx7XfGPC4OiETik4PT9E42H9Lo5gE6mOGINY4boG9KQPpN2pi3Y%2BJ1T4dzlovFgx5QbxjmPMe%2B91WB8WlITVcPM3aUEeGJKIjhKoKGUA6Cm2sc5HwosjCB89O7QJPFjeMABpRAiBOBqw1tKJMX9ZCvnFLVUhwDc7HRmv9SBxcL5s3jKmPQkRWA8AQO%2Bd6lMBFWmLJWf05gqiB0dB9MsQA)

**Please describe here fantomas problem you encountered**

#### Code

```fsharp
module Caching =
    type MainCache() =
        member __.RetrieveLastCompoundBalance (address: PublicAddress)
                                              (currency: Currency): NotFresh<decimal> =
            lock cacheFiles.CachedNetworkData (fun _ ->
                match balance with
                | NotAvailable -> NotAvailable
                | Cached (balance, time) ->
                    if compoundBalance < 0.0m then
                        ReportProblem compoundBalance None currency address sessionCachedNetworkData
                    ()
            )
            ()
```
            

#### Result

```fsharp
module Caching =
    type MainCache() =
        member __.RetrieveLastCompoundBalance (address: PublicAddress)
                                              (currency: Currency): NotFresh<decimal> =
            lock cacheFiles.CachedNetworkData (fun _ ->
                match balance with
                | NotAvailable -> NotAvailable
                | Cached (balance, time) ->
                    if compoundBalance < 0.0m then
                        ReportProblem compoundBalance None currency address
                            sessionCachedNetworkData
                    ())
            ()
```

#### Expected Result

The way that fantomas has cut off the parameters to the `ReportProblem` function looks a bit weird.

I think it should either do this:

```fsharp
module Caching =
    type MainCache() =
        member __.RetrieveLastCompoundBalance (address: PublicAddress)
                                              (currency: Currency): NotFresh<decimal> =
            lock cacheFiles.CachedNetworkData (fun _ ->
                match balance with
                | NotAvailable -> NotAvailable
                | Cached (balance, time) ->
                    if compoundBalance < 0.0m then
                        ReportProblem
                            compoundBalance None currency address sessionCachedNetworkData
                    ())
            ()
```

or this:

```fsharp
module Caching =
    type MainCache() =
        member __.RetrieveLastCompoundBalance (address: PublicAddress)
                                              (currency: Currency): NotFresh<decimal> =
            lock cacheFiles.CachedNetworkData (fun _ ->
                match balance with
                | NotAvailable -> NotAvailable
                | Cached (balance, time) ->
                    if compoundBalance < 0.0m then
                        ReportProblem
                            compoundBalance
                            None
                            currency
                            address
                            sessionCachedNetworkData
                    ())
            ()
```

or this:

```fsharp
module Caching =
    type MainCache() =
        member __.RetrieveLastCompoundBalance (address: PublicAddress)
                                              (currency: Currency): NotFresh<decimal> =
            lock cacheFiles.CachedNetworkData (fun _ ->
                match balance with
                | NotAvailable -> NotAvailable
                | Cached (balance, time) ->
                    if compoundBalance < 0.0m then
                        ReportProblem compoundBalance
                                      None
                                      currency
                                      address
                                      sessionCachedNetworkData
                    ())
            ()
```
            
#### Options

Fantomas Master at 06/10/2020 08:33:01 - 50ed1b7aab9bde89ea4d5d3fc6fd44b91f0741cb

| Name | Value |
| ---- | ----- |
| `IndentSpaceNum` | `4` |
| **`PageWidth`** | **`80`** |
| `SpaceBeforeParameter` | `true` |
| `SpaceBeforeLowercaseInvocation` | `true` |
| `SpaceBeforeUppercaseInvocation` | `false` |
| `SpaceBeforeClassConstructor` | `false` |
| `SpaceBeforeMember` | `false` |
| `SpaceBeforeSemicolon` | `false` |
| `MaxInfixOperatorExpression` | `50` |
| `MaxRecordWidth` | `40` |
| `MaxArrayOrListWidth` | `40` |
| `MaxValueBindingWidth` | `40` |
| **`MaxFunctionBindingWidth`** | **`0`** |
| `MultilineBlockBracketsOnSameColumn` | `false` |
| `NewlineBetweenTypeDefinitionAndMembers` | `false` |
| `KeepIfThenInSameLine` | `false` |
| `StrictMode` | `false` |

Bobface:
Fixes #907 by putting every parameter of a function application on a newline if the function application exceeds `PageWidth`.

cc @knocte

### DIFF
--- a/src/Fantomas.Tests/ElmishTests.fs
+++ b/src/Fantomas.Tests/ElmishTests.fs
@@ -433,7 +433,8 @@ table [ ClassName "table table-striped table-hover mb-0" ] [
         tokenDetailRow "ColorClass" (str colorClass)
         tokenDetailRow "CharClass" (str charClass)
         tokenDetailRow "Tag" (ofInt tag)
-        tokenDetailRow "FullMatchedLength"
+        tokenDetailRow
+            "FullMatchedLength"
             (span [ ClassName "has-text-weight-semibold" ] [
                 ofInt fullMatchedLength
              ])
--- a/src/Fantomas.Tests/RecordTests.fs
+++ b/src/Fantomas.Tests/RecordTests.fs
@@ -571,7 +571,8 @@ let expect =
                 Opts.oneOf
                     (Optional,
                      [ Opt.flag [ "third"; "f" ]
-                       Opt.valueWith "new value"
+                       Opt.valueWith
+                           "new value"
                            [ "fourth"
                              "ssssssssssssssssssssssssssssssssssssssssssssssssssss" ] ]) ]
           args = []
--- a/src/Fantomas/CodePrinter.fs
+++ b/src/Fantomas/CodePrinter.fs
@@ -1346,13 +1346,43 @@ and genExpr astContext synExpr =
             else
                 sepSpace ctx
 
-        atCurrentColumn
-            (genExpr astContext e
-             +> colPre sepSpace sepSpace es (fun e ->
-                    onlyIf (isCompExpr e) (sepSpace +> sepOpenSFixed +> sepSpace)
+        let shortExpression =
+            atCurrentColumn
+                (genExpr astContext e
+                 +> colPre sepSpace sepSpace es (fun e ->
+                        onlyIf (isCompExpr e) (sepSpace +> sepOpenSFixed +> sepSpace)
+                        +> indent
+                        +> appNlnFun e (indentIfNeeded +> genExpr astContext e)
+                        +> unindent))
+
+        let longExpression =
+            (atCurrentColumn
+                (genExpr astContext e
+                +> colPre sepSpace sepSpace es (fun e ->
+                ifElse (isCompExpr e)
+                    (sepSpace
+                    +> sepOpenSFixed
+                    +> sepSpace
                     +> indent
                     +> appNlnFun e (indentIfNeeded +> genExpr astContext e)
-                    +> unindent))
+                    +> unindent)
+                    (indent
+                    +> indentIfNeeded
+                    +> sepNln
+                    +> genExpr astContext e
+                    +> unindent))))
+
+        if List.exists (function
+            | Lambda _
+            | MatchLambda _
+            | Paren (Lambda (_))
+            | Paren (MatchLambda (_))
+            | MultilineString _
+            | CompExpr _ -> true
+            | _ -> false) es then
+            shortExpression
+        else
+            expressionFitsOnRestOfLine shortExpression longExpression
 
     | TypeApp(e, ts) -> genExpr astContext e -- "<" +> col sepComma ts (genType astContext false) -- ">"
     | LetOrUses(bs, e) ->
--- a/src/Fantomas.Tests/AppTests.fs
+++ b/src/Fantomas.Tests/AppTests.fs
@@ -76,3 +76,131 @@ let a s =
             (llloooooooooooooooooooooooooo s)
             (llloooooooooooooooooooooooooo s)
 "
+
+
+[<Test>]
+let ``should split parameters over multiple lines when they exceed page width``() =
+    formatSourceString false """module Caching =
+    type MainCache() =
+        member __.RetrieveLastCompoundBalance (address: PublicAddress)
+                                              (currency: Currency): NotFresh<decimal> =
+            lock cacheFiles.CachedNetworkData (fun _ ->
+                match balance with
+                | NotAvailable -> NotAvailable
+                | Cached (balance, time) ->
+                    if compoundBalance < 0.0m then
+                        ReportProblem compoundBalance None currency address sessionCachedNetworkData
+                    ()
+            )
+            ()""" { config with PageWidth = 60 }
+    |> prepend newline
+    |> should equal """
+module Caching =
+    type MainCache() =
+        member __.RetrieveLastCompoundBalance (address: PublicAddress)
+                                              (currency: Currency): NotFresh<decimal> =
+            lock cacheFiles.CachedNetworkData (fun _ ->
+                match balance with
+                | NotAvailable -> NotAvailable
+                | Cached (balance, time) ->
+                    if compoundBalance < 0.0m then
+                        ReportProblem
+                            compoundBalance
+                            None
+                            currency
+                            address
+                            sessionCachedNetworkData
+                    ())
+            ()
+"""
+
+[<Test>]
+let ``should split single parameter over multiple lines when it exceeds page width``() =
+    formatSourceString false """module Caching =
+    type MainCache() =
+        member __.RetrieveLastCompoundBalance (address: PublicAddress)
+                                              (currency: Currency): NotFresh<decimal> =
+            lock cacheFiles.CachedNetworkData (fun _ ->
+                match balance with
+                | NotAvailable -> NotAvailable
+                | Cached (balance, time) ->
+                    if compoundBalance < 0.0m then
+                        ReportProblem looooooooooooooooooooooooooooooooooooooooooooooooooooooooooooong
+                    ()
+            )
+            ()""" { config with PageWidth = 60 }
+    |> prepend newline
+    |> should equal """
+module Caching =
+    type MainCache() =
+        member __.RetrieveLastCompoundBalance (address: PublicAddress)
+                                              (currency: Currency): NotFresh<decimal> =
+            lock cacheFiles.CachedNetworkData (fun _ ->
+                match balance with
+                | NotAvailable -> NotAvailable
+                | Cached (balance, time) ->
+                    if compoundBalance < 0.0m then
+                        ReportProblem
+                            looooooooooooooooooooooooooooooooooooooooooooooooooooooooooooong
+                    ())
+            ()
+"""
+
+[<Test>]
+let ``should not split parameters over multiple lines when they do not exceed page width``() =
+    formatSourceString false """module Caching =
+    type MainCache() =
+        member __.RetrieveLastCompoundBalance (address: PublicAddress)
+                                              (currency: Currency): NotFresh<decimal> =
+            lock cacheFiles.CachedNetworkData (fun _ ->
+                match balance with
+                | NotAvailable -> NotAvailable
+                | Cached (balance, time) ->
+                    if compoundBalance < 0.0m then
+                        ReportProblem compoundBalance None currency address sessionCachedNetworkData
+                    ()
+            )
+            ()""" { config with PageWidth = 120 }
+    |> prepend newline
+    |> should equal """
+module Caching =
+    type MainCache() =
+        member __.RetrieveLastCompoundBalance (address: PublicAddress) (currency: Currency): NotFresh<decimal> =
+            lock cacheFiles.CachedNetworkData (fun _ ->
+                match balance with
+                | NotAvailable -> NotAvailable
+                | Cached (balance, time) ->
+                    if compoundBalance < 0.0m
+                    then ReportProblem compoundBalance None currency address sessionCachedNetworkData
+                    ())
+            ()
+"""
+
+[<Test>]
+let ``should not split single parameter over multiple lines when it does not exceed page width``() =
+    formatSourceString false """module Caching =
+    type MainCache() =
+        member __.RetrieveLastCompoundBalance (address: PublicAddress)
+                                              (currency: Currency): NotFresh<decimal> =
+            lock cacheFiles.CachedNetworkData (fun _ ->
+                match balance with
+                | NotAvailable -> NotAvailable
+                | Cached (balance, time) ->
+                    if compoundBalance < 0.0m then
+                        ReportProblem compoundBalance
+                    ()
+            )
+            ()""" { config with PageWidth = 120 }
+    |> prepend newline
+    |> should equal """
+module Caching =
+    type MainCache() =
+        member __.RetrieveLastCompoundBalance (address: PublicAddress) (currency: Currency): NotFresh<decimal> =
+            lock cacheFiles.CachedNetworkData (fun _ ->
+                match balance with
+                | NotAvailable -> NotAvailable
+                | Cached (balance, time) ->
+                    if compoundBalance < 0.0m then ReportProblem compoundBalance
+                    ())
+            ()
+"""
--- a/src/Fantomas.Tests/CompilerDirectivesTests.fs
+++ b/src/Fantomas.Tests/CompilerDirectivesTests.fs
@@ -476,7 +476,9 @@ type FunctionComponent =
                 (importValueDynamic f).``then``(fun x -> createObj [ "default" ==> x ]))
 
         fun props ->
-            ReactElementType.create ReactBindings.React.Suspense (createObj [ "fallback" ==> fallback ])
+            ReactElementType.create
+                ReactBindings.React.Suspense
+                (createObj [ "fallback" ==> fallback ])
                 [ ReactElementType.create elemType props [] ]
 #else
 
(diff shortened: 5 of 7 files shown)


Bobface:
@nojaf What is your first impression of this implementation? I am not quite sure about some changes I had to make to some tests.

Also, what should we do if the parameters do not exceed page width if they were in one line, but they are split over multiple lines in the original and have comments at EOL? I was thinking to keep them split over multiple lines in that case.

src/Fantomas.Tests/ElmishTests.fs
```diff
@@ -433,10 +433,9 @@ table [ ClassName "table table-striped table-hover mb-0" ] [
         tokenDetailRow "ColorClass" (str colorClass)
         tokenDetailRow "CharClass" (str charClass)
         tokenDetailRow "Tag" (ofInt tag)
-        tokenDetailRow "FullMatchedLength"
-            (span [ ClassName "has-text-weight-semibold" ] [
-                ofInt fullMatchedLength
-             ])
+        tokenDetailRow "FullMatchedLength" (span [ ClassName "has-text-weight-semibold" ] [
```
nojaf, line 436: This seems a bit weird as the span block is multiline.
So I would expect it to be more like
```fsharp
tokenDetailRow
    "FullMatchedLength"
    (span [ ...

    ])
```

src/Fantomas.Tests/RecordTests.fs
```diff
@@ -571,9 +571,8 @@ let expect =
                 Opts.oneOf
                     (Optional,
                      [ Opt.flag [ "third"; "f" ]
-                       Opt.valueWith "new value"
-                           [ "fourth"
-                             "ssssssssssssssssssssssssssssssssssssssssssssssssssss" ] ]) ]
+                       Opt.valueWith "new value" [ "fourth"
```
nojaf, line 574: Same remark, the last part is multiline

src/Fantomas/CodePrinter.fs
```diff
@@ -1322,13 +1322,34 @@ and genExpr astContext synExpr =
...
-        atCurrentColumn
-            (genExpr astContext e
-             +> colPre sepSpace sepSpace es (fun e ->
-                    onlyIf (isCompExpr e) (sepSpace +> sepOpenSFixed +> sepSpace)
-                    +> indent
-                    +> appNlnFun e (indentIfNeeded +> genExpr astContext e)
-                    +> unindent))
+        fun ctx ->
+            (ifElseCtx 
+                (exceedsWidth ctx.Config.PageWidth
```
nojaf, line 1327: Use `expressionFitsOnRestOfLine` instead here.
I would also create separate values for the short and long expression path.
Something like:
```fsharp
fun ctx ->
    let shortExpression = ...

    let longExpression = ...

    expressionFitsOnRestOfLine shortExpression longExpression ctx
```

nojaf: I think you might want to also check for `isCompExpr ` and use the existing expression in that case.

Bobface: I have tried to use `expressionFitsOnRestOfLine` however it directly goes into "short expression mode" and splits even very short apps over multiple lines in some cases. Is that intended?

src/Fantomas.Tests/ElmishTests.fs
Bobface: I believe the problem here is the following:
When the long expression is checked the printer tries to put all parameters in one line. Further down in the recursive calls where the array is printed it is detected that the page width will be exceeded if the array is printed in one line so it is split.
That satisfies the maximum width for the long expression again.

Bobface: Might need your help on what's the best way to fix that one.

src/Fantomas/CodePrinter.fs
nojaf: Did you try it in this PR? What you are describing sounds a bit weird.

Bobface: Yes, I have just pushed the version with `expressionFitsOnRestOfLine` so you can see the failing tests where even short expressions are newlined. I assume `shortExpressionWithFallback ` which is used by `expressionFitsOnRestOfLine` is the culprit (see the comment):
```
let private shortExpressionWithFallback (shortExpression: Context -> Context) (fallbackExpression) maxWidth startColumn (ctx: Context) =
    // if the context is already inside a ShortExpression mode and tries to figure out if the expression will go over the page width,
    // we should try the shortExpression in this case.
    match ctx.WriterModel.Mode with
    | ShortExpression infos when (List.exists (fun info -> info.ConfirmedMultiline || info.IsTooLong ctx.Config.PageWidth ctx.Column) infos) ->
        ctx
    | _ -> ....
```

src/Fantomas/CodePrinter.fs
```diff
@@ -1322,13 +1322,64 @@ and genExpr astContext synExpr =
...
                     +> appNlnFun e (indentIfNeeded +> genExpr astContext e)
-                    +> unindent))
+                    +> unindent)
+                    (indent
+                    +> indentIfNeeded
+                    +> genExpr astContext e
+                    +> unindent))))
+
+            //ifElseCtx (fun ctx -> exceedsWidth ctx.Config.PageWidth longExpression ctx || parametersMustBeNewlinedBecauseOfTriviaContent es ctx) shortExpression longExpression ctx
+            ifElseCtx (fun ctx -> parametersMustBeNewlinedBecauseOfTriviaContent es ctx) shortExpression (expressionFitsOnRestOfLine longExpression shortExpression) ctx
```
nojaf, line 1382: `expressionFitsOnRestOfLine longExpression shortExpression` that doesn't really make much sense.
The function `expressionFitsOnRestOfLine` will first try the longExpression and use the short one as fallback.
Those parameters should be swapped.

I also think you don't always want this behaviour, it will depend on the `es` of `| App(e, es)`.
If one expression in `es` is a lambda for example I don't think you necessarily want the suggested formatting of putting every argument on one line.

For example:
```fsharp
let foo =
    Array.map (fun x ->
                                  // some multiline code
                       x + x) [| 8;9;6;5 |]
```

I don't think you want to make 
```fsharp
let foo =
    Array.map
         (fun x ->
                       // some multiline code
           x + x)
     [| 8;9;6;5 |]
```
of it.

src/Fantomas/CodePrinter.fs
```diff
@@ -1322,13 +1322,64 @@ and genExpr astContext synExpr =
...
+        // This can stay as is because the last argument has the line comment
+        // However:
+        //
+        //     someFunc
+        //         a 
+        //         b // comment
+        //         c
+        //
+        // This has to be split over multiple lines to preserve the line comment
+        let parametersMustBeNewlinedBecauseOfTriviaContent es ctx =
```
nojaf, line 1341: For what unit test did you introduce this? I would assume that a comment in the expression will force `expressionFitsOnRestOfLine` to take the fallback route so I'm not quite sure this is necessary here.

src/Fantomas.Tests/ElmishTests.fs
nojaf: The code now makes
```fsharp
        tokenDetailRow
            "FullMatchedLength"
            (span [ ClassName "has-text-weight-semibold" ] [
                ofInt fullMatchedLength
             ])
```
of it. This is correct behaviour in my opinion. The function call doesn't fit on the remainders of the line because of the arguments doesn't fit. This makes sense.
I think the exception here will be lambda's, if you have those, you might want to keep the original behaviour. See next comment.

nojaf:
Hey @Bobface, what do you think of https://github.com/nojaf/fantomas/commit/6a28f6f9b714bf160fa648195b8aedf9667f4d22.
I limited the scope of where exactly the parameters are split into multiple lines.
Fantomas does not always consider multiline strings as multiline constructs ([example](https://fsprojects.github.io/fantomas-tools/#/fantomas/preview?data=N4KABGBEDGD2AmBTSAuKAbRAXMBDMAvGADqSnkBOxAdrABZjqywDWxV1%2BAtgK7pYBLdAOqJ2NAM5YKIgObkykSABpwUCdkHVZE1GADaaiKAimoAEiwBPAA7I0kEVhVHTkcwDdc6HvYOuzAAZVMzNIAElqJGosAGUbXGhEADkeLhdQ0wAWALAAXVcAXxDTE1D3azs9RxiM8s9vXz1DTIgARhLMyAAFXFlEAHUBeCw6Osy2gCZA3IKzYtcysMtbP0gAI2Z0cbcGnz8W1snO8tjELgE4JmoAQSwAUSiAeQAzABkRZBOzF%2B8NWaKJyWu0qa02sG23wsXn2zVyEAAzFCIJB4olEAAhRAvWAURC9Ci4LjYRAUHZmaS%2BAHzIG5CqrargyF0vZNNCHTJZZFQNFJLE4vFvWAAd1J0FwGkiHlg4sEsGo5NMlMQ1NMCzMwJRKyqDiZivcMLZ-laYAArNzUQk%2BdjcYgAKo2OwUcWS6jS2UCeWKiC-dD-TJzNW0rrasFbfWsg7wsAANgtvMxNrxAGF0BKJMn5VIKDxoFhcd6wL7-aFAxB1aUWaDGeGoQbGlGTQB2eNWxMCxAAWXO61JheLKoDgMWVYZutrLMNjdaAA5W%2Bj%2BbbM9d%2B39B6XhxrRzqoHq65G4SaAJzzpI3F5YUmZrhcXCF5Wq8vB%2BrV8cQiNTw%2BtNrBaOWhdJogZwXFcXrcgOj5gBWxjbmG777p%2B7LRm0HR-gm56XhQwGXBCYHRg%2BQ40iOIavruE4hohxrfscf6RNEWBPNQAAqFBWEMoyrn665mGWUHPsspEbOR9SURyoRtEiaFtjcFCwDwUQACKIMIFyYfeObcaYvHQWAmoWIJTgfg2X4TFyf6drgAAe4QvExdCINQ9xcbEdC4lgQwjGM3JZDMhFBsRL5jlAhkIcZSEmm05rmVZkQvAIllPE6uD5hQ9yWTYeISBInoKtypq%2BRuRFbiRQU1M4oWwuF35xtFlkAEqIHAFDwB5HHeQVPGbpWJU7mVRmVVREwtrVMmElYTwUB8UitV50Y%2BZBOl6fSvUhZOYWDeJc61QAag2GIiPAcgzYW81%2BU%2BAUCaVq0UetYlmG0J61QAYvJeY5ftURHcMbVzR1WldTBPVwcyN0DXdpjTBanZ8IIwiiBiTDQCwGKEkj2ASIxsREogy5pLl0YQWdfEXSCpV7mtYPRpMqEmpAySIMKcOJlgooOUxqxKXF1ACHKtxRN2XC9hQujgWuC38aTvXk6DRrgxAkw0bTADSiCIDYNl2Q5kRY8SHyiJxJadUV3WBSttQVbLVOSbTFmWU5FwSHQx3teLJNaoJ0sibdVNmbTsRyJgMmyGkDnuYg6ydggXwE2LROLbBNbwRTlsmpMUV%2B9IlxYJHSAG5pEDaWoeQlI4EhPdlegQYUQA), there in theory Fantomas should add the whole thing on a new line after the `=` but this is a known exception).

The same for lambda's, we want the `(fun x ->` part still on the same line, that makes more sense mostly.

Bobface:
@nojaf This PR is now using your implementation for the `CodePrinter`. It works fine, but there is one test in `AppTests` where I am not sure if it is correct:

```
[<Test>]
let ``require to ident at least +1 after function name #545 (long expression and short line settings)``() =
    formatSourceString false @"
let a s =
    if s <> """" then
        printfn """"""fooo
%s
%s
%s
%s""""""                (llloooooooooooooooooooooooooo s)
                            s
                               (llloooooooooooooooooooooooooo s)
                                     (llloooooooooooooooooooooooooo s)"
        { config  with PageWidth = 50 } 
    |> prepend newline
    |> should equal @"
let a s =
    if s <> """" then
        printfn """"""fooo
%s
%s
%s
%s""""""    (llloooooooooooooooooooooooooo s) s
            (llloooooooooooooooooooooooooo s)
            (llloooooooooooooooooooooooooo s)
"
```

Some of the parameters for `printfn` are newlined while others aren't. Is that intended for multiline strings?

nojaf:
Hey @Bobface , yes that test looks a bit weird but it still makes sense I think:

So when there is a multiline string in place, the old formatting will be used.
And there is the newline function is determined by
```fsharp
    let appNlnFun e =
        match e with
        | MultilineString _
        | Lambda _
        | Paren (Lambda _)
        | Paren (MatchLambda _) -> id
        | _ -> autoNlnIfExpressionExceedsPageWidth
```

so for the MultilineString we don't add a newline, yet for all the other parameter we add a newline when they reach the page width threshold.
Fantomas just doesn't always consider multiline strings as multiline expressions, otherwise you would have a newline for each multiline string and you might not want that.

@jindraivanek would you mind reviewing this as please?

PR #898, opened 2020-06-07
Format long member binding and constructors according to the F# style guide

nojaf:
Fixes #868.
Notice that not everything is possible to adhere to the max page width.
For example:
```fsharp
type Foo =
    static member Deserialize(loooooooooooooooooooooooooooooooooooooooooooooooooooooonnnnnnnnnnnnnnnnnnnnngggggggggggJsonVaaaaalueeeeeeeeeeeeeeee): SomeType =
        Decode.SomeType
            loooooooooooooooooooooooooooooooooooooooooooooooooooooonnnnnnnnnnnnnnnnnnnnngggggggggggJsonVaaaaalueeeeeeeeeeeeeeee
```

I think this is an acceptable trade-off for now. I'm open to discussing this though.

PR #935, opened 2020-06-30
Put return type on newline if expression does not fit on rest of line

Bobface:
Puts the return type of long member definitions on a new line if the expression does not fit on the rest of the line (as discussed in #898).

@nojaf 2 Questions:
- The current implementation feels a bit hacky: In `genMemberBinding` the column after the name of the member binding has been printed is calculated and later used by `atIndentLevel` to print the return type. The reason it is implemented like this is that the name of the binding and the parameters are handled by `genPat` while the return type is handled by `genMemberBinding`. The latter does not know about which column is used for the parameters - so we calculate it manually again in `genMemberBinding`. Are you happy with this?
- Should this behaviour also be implemented for all function definitions, or just for member bindings?

cc @knocte

### DIFF
--- a/src/Fantomas.Tests/AppTests.fs
+++ b/src/Fantomas.Tests/AppTests.fs
@@ -77,7 +77,6 @@ let a s =
             (llloooooooooooooooooooooooooo s)
 "
 
-
 [<Test>]
 let ``should split parameters over multiple lines when they exceed page width``() =
     formatSourceString false """module Caching =
@@ -98,7 +97,8 @@ let ``should split parameters over multiple lines when they exceed page width``(
 module Caching =
     type MainCache() =
         member __.RetrieveLastCompoundBalance (address: PublicAddress)
-                                              (currency: Currency): NotFresh<decimal> =
+                                              (currency: Currency)
+                                              : NotFresh<decimal> =
             lock cacheFiles.CachedNetworkData (fun _ ->
                 match balance with
                 | NotAvailable -> NotAvailable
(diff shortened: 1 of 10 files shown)


nojaf:
Hello @Bobface ,
* The implementation looks at first glance indeed a bit hacky, I think `atCurrentColumn` can be used instead. I'll try and check it myself later this week.
* For now I would keep it for just for member bindings. We have something similar for [let bindings](https://github.com/fsprojects/fantomas/blob/master/src/Fantomas.Tests/FunctionDefinitionTests.fs#L480). I'm trying to get some more feedback of the styleguides on that [here](https://github.com/dotnet/docs/issues/18806) and [here](https://github.com/G-Research/fsharp-formatting-conventions/issues/4).

Bobface:
I have tried it with `atCurrentColumn` but to no success. From my understanding the reason seems to be that `atCurrentColumn` will put the content at indentation level of the *end* of the last line. Example:

```
member someFunc (someParam1: string) (someParam2: string) : int = 
    ...
```

results in

```
member someFunc (someParam1: string) 
                (someParam2: string)
                                   : int = 
    ...
```

`genMemberBinding` will print the return type after `genPat` is done printing the name and parameters of the binding. The last line printed by `genPat` was the last parameter and the column is set to the end of that line.

nojaf:
You would need to do the atColumn thing after printing out the `someFunc` part.
I'll be able to help out more later this week.

nojaf:
Hey @Bobface, I tried this myself and I very much acknowledge that this is a hard thing to solve 🙈
I came up with https://github.com/fsprojects/fantomas/compare/master...nojaf:fix-return-type-member?expand=1
Which is a slightly different approach as I'm basically moving the code from getPat to member binding expression.
Then I have all the information to make some explicit choices.
I think this is somewhat more readable and will be easier to maintain.

In the near future, I'll try and implement the [GR approach](https://github.com/G-Research/fsharp-formatting-conventions#place-parameters-on-a-new-line-for-very-long-member-definitions) for this.
With that in mind, I think my approach will have some benefits there.

That being said, your PR was very helpful to reason about the problem.
So thank you for that.

Let me know what you think about this.

Bobface:
@nojaf That works fine. Thanks for the implementation!

PR #939, opened 2020-07-01
.editorconfig support

josefblaha:
This is my work-in-progress on the issue #650. In the first commit, I've implemented reading indentation size from a `.editorconfig` file. There is a new method ``CodeFormatter.ReadDocumentConfiguration`` to get the formatting configuration for a source file. It is intended to replace the current ``CodeFormatter.ReadConfiguration``.

Notable TODOs are:
- [ ] Define and read other Fantomas formatting options from the file.
- [ ] Implement assembly weaving (or something equivalent), so that the dependency library is hidden.
- [ ] Get rid of the restore warnings during build. 

Also feel free to comment my programming style. I come from the C# world and F# is still a new language for me.

Issue #650, opened 2020-01-25
Use .editorconfig to define Fantomas settings


xperiandri:
.editorconfig is a common solution to define all code related rules including formatting and code analysis.
It would be the right option for Fantomas to read settings from it.

nojaf:
Hey @xperiandri , this issue came just in time ;)
I'm about to release to 3.2 stable and that would mark `fantomas-config.json` as public api so glad we can discuss other options before that.

I'm a bit less familiar with .editorconfig so my main question would be: how does this work with custom settings? Can we just write anything we want in the file and parse it as we want?

How would our current default look like in .editorconfig?
```json
{
  "IndentSpaceNum":4,
  "PageWidth":120,
  "SemicolonAtEndOfLine":false,
  "SpaceBeforeArgument":true ,
  "SpaceBeforeColon":false,
  "SpaceAfterComma":true ,
  "SpaceAfterSemicolon":true ,
  "IndentOnTryWith":false,
  "ReorderOpenDeclaration":false,
  "SpaceAroundDelimiter":true ,
  "KeepNewlineAfter":false,
  "MaxIfThenElseShortWidth":40,
  "StrictMode":false
}
```

One advantage of using a JSON file is that we have listed a [JSON schema](http://json.schemastore.org/fantomas). Would there be something similar for editorconfig?

xperiandri:
@madskristensen could you explain what is the source of [Editor Config Language](https://github.com/madskristensen/EditorConfigLanguage/) autocomplete? How can Fantomas provide F# specific settings to a service?

xperiandri:
```
# All files
[*]
indent_style = space
# Code files
[*.{cs,csx,fs,fsx,vb,vbx}]
indent_size          = 4         #"IndentSpaceNum":4
insert_final_newline = true
charset              = utf-8-bom
max_line_length      = 120       #"PageWidth":120
###############################
# .NET Coding Conventions     #
###############################
[*.{cs,fs,vb}]
# Organize usings
dotnet_sort_system_directives_first = false            #"ReorderOpenDeclaration":false
###############################
# F# Formatting Rules         #
###############################
fsharp_strict_mode                  = false            #"StrictMode":false
fsharp_semicolon_at_end_of_line     = true             #"SemicolonAtEndOfLine":false
fsharp_space_before_argument        = true             #"SpaceBeforeArgument":true
fsharp_space_around_coma            = after            #"SpaceAfterComma":true
[...]

PR #940, opened 2020-07-01
Add support for editorconfig

nojaf:
Current attempt for solving #650.
EditorConfig.Core was forked to https://github.com/nojaf/editorconfig-core-net.
There I created a slightly modified version, mainly having a netstandard2.0 only configuration.

I used ILRepack to merge in the editorconfig core to Fantomas.dll.
I published `Fantomas.EditorConfig.Core` as unlisted to Nuget.

I fear that this PR is already competing with #939, for that @josefblaha I'm sorry.
Last Friday I already had those changes to `readConfiguration `.

Main things left to do here are:
- [x] update unit tests
- [x] integrate into CLI tool
- [x] remove all traces of JSON config (including schema)


josefblaha:
What an exciting world is this OSS development, with developers racing to complete features ;-)

@nojaf, you are on a good track to finish this, so I will no purse it anymore. Just two things came to my mind when reviewing your code:

1. For ``IndentSpaceNum`` option, there is no one-to-one mapping with editorconfig. IMO the value should be derived from ``indent_style``, ``indent_size``, and ``tab_width`` properties.
2. No warnings should be generated from unrecognized properties; any other custom properties or possible future global properties should be ignored.

nojaf:
Hey @josefblaha, again apologies for not creating my PR sooner. I should have made a draft PR last Friday, that may have avoided this.

As for your feedback:
1. F# doesn't allow tabs so we only need to take spaces into account.
2. Good point, warnings should not be listed anymore as they count have meaning to other things.

josefblaha:
No problem. Any code written has a value for me, as I'm still learning the language.

Ad tabs. You're right, but there is a corner case: ``indent_size = tab`` means that you need to look on ``tab_width`` to get the value. See (https://github.com/editorconfig/editorconfig/wiki/EditorConfig-Properties#widely-supported-by-editors).

nojaf:
Actually, the underlying EditorConfig.Core Nuget solves this internally already,
If I had for example:
```ini
[*.fs]
indent_style=tab
indent_size=tab
tab_width=5
fsharp_indent_on_try_with=true
```

I got
image

I'll add a unit test to cover this as well.

nojaf:
This is ready for now.
A couple of significant changes:
* You don't need to pass down the configuration file in the cmd tool. Fantomas will pick it up based on the original file you passed.
  One trade-off here is that you can't pass a configuration anymore when you use the stdIn. I would add this back in if requested.
  I'm not sure if many people actually use this.
* Same remark for using the FAKE Helpers, config will be picked up based on the F# file, to me this is the way to go.
  I'll update the sample once there is a new alpha released.
* I've renamed IndentSpaceNum and PageWidth to match the .editorconfig names.
* Fantomas-tools would need an update after this PR.

src/Fantomas/CodeFormatter.fsi
```diff
@@ -36,4 +36,4 @@ type CodeFormatter =
 
     /// Accepts a file or a folder and parses the found json to a FormatConfig
```
jindraivanek, line 37:
```suggestion
    /// Accepts a file or a folder and parses the found .editorconfig file to a FormatConfig
```

PR #947, opened 2020-07-03
Revisit long function definitions

### LINKED ISSUE #946: Review long function definitions
I propose that Fantomas formats long function definitions by default as described by the [Microsoft style guide](https://github.com/dotnet/docs/issues/18806) and place the [GR style](https://github.com/G-Research/fsharp-formatting-conventions/issues/4) behind a setting.

The existing way of Fantomas deals with this problem is using the G-Research style guide.

## Pros and Cons 

The advantages of making this adjustment to Fantomas are that we respect both style guides.

The disadvantages of making this adjustment to Fantomas are an extra setting to maintain.

## Examples

Please provide multiple examples (before and after scenarios) and explain what rules did apply to achieve the outcome.

Microsoft:
```fsharp
let fold (funcs: ResultFunc<'Input, 'Output, 'TError> seq)
         (input: 'Input)
         (input2: 'Input)
         (input3: 'Input)
         : Result<'Output list, 'TError list> =
    ()
```

G-Research
```fsharp
let fold
    (funcs: ResultFunc<'Input, 'Output, 'TError> seq)
    (input: 'Input)
    (input2: 'Input)
    (input3: 'Input)
    : Result<'Output list, 'TError list>
    =
    ()
```

## Extra information

Estimated cost: S

Related suggestions: #935, #492, #769

## Affidavit (please submit!)

* [X] I have [searched both open and closed suggestions on this site](http://github.com/fsharp/fslang-suggestions/issues) and believe this is not a duplicate

Please tick all that apply:
* [ ] This is not a breaking change to Fantomas
* [X] I or my company would be willing to help implement and/or test this
* [X] This suggestion is part of the [Microsoft style guide](https://docs.microsoft.com/en-us/dotnet/fsharp/style-guide/formatting) (please add a link to section if so)
* [X] This suggestion is part of the [G-Research style guide](https://github.com/G-Research/fsharp-formatting-conventions) (please add a link to section if so)

nojaf:
Adding support for both the Microsoft and G-Research style of formatting long function signatures.
Fixes #946


src/Fantomas.Tests/FunctionDefinitionTests.fs
```diff
@@ -526,14 +518,12 @@ let private addTaskToScheduler (scheduler : IScheduler) taskName taskCron prio (
...
-    prio
-    (task: unit -> unit)
-    groupName
-    =
+let private addTaskToScheduler (scheduler: IScheduler)
+                               taskName
+                               taskCron
+                               prio
+                               (task: unit -> unit)
+                               groupName =
```
knocte, line 526: @nojaf shouldn't the equals character here be in the next line? this way if someone adds a new parameter, only 1 line needs to change (instead of 2)

nojaf: Yeah I thought about it, but that would be inconsistent with how Microsoft formats member definitions. There the equals sign is not on a new line.

knocte: I don't think you're interpreting correctly (from the snippet I posted). The rule could be that the equals line is in the same line where the return type is (if there's no return type, then the equals is left alone).

knocte: and I just realised that the snippet I was referring to (https://github.com/dotnet/docs/issues/18806#issuecomment-651890110) actually has the equal char in its own line, even when there is a return type, but I didn't mean to do that; I think the equals char can stay in the same line as the return type in that case

nojaf: Well, I understood it as: if there is a return type, it and the equal sign can go to the next line.
Following the same logic as what was written [here](https://github.com/dotnet/docs/pull/18793/files#diff-eed63f6624dbb527e2e9533697f1ff6bR112).

nojaf: Updated according to the conversation in https://github.com/dotnet/docs/issues/18806

knocte:
LGTM

nojaf:
Thanks, I'm just going to wait for @jindraivanek to take a look at the implementation, as quite some code changes were made.

PR #957, opened 2020-07-10
Support GR very long member definitions

### LINKED ISSUE #913: Long members and constructors in GR style
The GR style guide has an [alternative way](https://github.com/G-Research/fsharp-formatting-conventions#place-parameters-on-a-new-line-for-very-long-member-definitions) of formatting long members and constructors.

I would add this functionality behind a setting. Default mode would be the MS style guide.

nojaf:
Fixes #913 , see [GR style guide](https://github.com/G-Research/fsharp-formatting-conventions#place-parameters-on-a-new-line-for-very-long-member-definitions)

I also added some more test to get the default MS style on point.

### DIFF
--- a/src/Fantomas/CodePrinter.fs
+++ b/src/Fantomas/CodePrinter.fs
@@ -2249,7 +2249,15 @@ and genTypeDefn astContext (TypeDef(ats, px, ao, tds, tcs, tdr, ms, s, preferPos
 
         let astContext = { astContext with InterfaceRange = interfaceRange }
 
-        typeName +> sepSpaceBeforeClassConstructor +> opt sepNone impCtor (genMemberDefn astContext) +> sepEq
+        typeName
+        +> sepSpaceBeforeClassConstructor
+        +> leadingExpressionIsMultiline
+               (opt sepNone impCtor (genMemberDefn astContext))
+               (fun isMulti ctx ->
+                    if isMulti && ctx.Config.AlternativeLongMemberDefinitions then
+                        sepEqFixed ctx
+                    else
+                        sepEq ctx)
         +> indent +> sepNln
         +> genTrivia tdr.Range
             (genTypeDefKind tdk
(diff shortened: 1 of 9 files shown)


nojaf:
@Smaug123 @JackMatusiewiczGR if you prefer another setting name than `AlternativeLongMemberDefinitions`, I'm game for suggestions.

src/Fantomas/CodePrinter.fs
```diff
@@ -681,48 +681,129 @@ and genMemberBinding astContext b =
             genMemberFlagsForMemberBinding astContext mf b.RangeOfBindingAndRhs
             +> ifElse isInline (!- "inline ") sepNone +> opt sepSpace ao genAccess
 
+        let rec genNameAndParameters p =
+            match p with
+            | PatLongIdent(ao, s, ps, tpso) ->
+                let aoc = opt sepSpace ao genAccess
+                let tpsoc = opt sepNone tpso (fun (ValTyparDecls(tds, _, tcs)) -> genTypeParamPostfix astContext tds tcs)
+                let s = if s = "``new``" then "new" else s
```
Smaug123, line 689: I think I don't understand why this would be necessary - could you fill me in please?

src/Fantomas/Context.fs
```diff
@@ -461,6 +461,7 @@ let internal sepNlnUnlessLastEventIsNewline (ctx: Context) =
 
 let internal sepStar = !- " * "
 let internal sepEq = !- " ="
+let internal sepEqFixed = !- "="
```
Smaug123, line 464: Why the name "fixed" here?

src/Fantomas/CodePrinter.fs
nojaf: Normally the pattern expression of the member would be composed in [genPat](https://github.com/fsprojects/fantomas/blob/515ce2ee6beaab631df9a47e0c9fd945c04e7283/src/Fantomas/CodePrinter.fs#L2971). There the name of the member and parameters would be printed.

To adhere to both style guides I needed more control over this. 
I could not control certain newline decisions in `genPat` as there are specific to the member binding.
`genPat` is used all over the place so that is where that extra `new` check will somewhere be useful.

As I copied over some of the code from `genPat` to `genNameAndParameters` I didn't really think about this.
I can indeed remove the extra `let s = if ...` line.

src/Fantomas/Context.fs
nojaf: The suffix `fixed` is used to indicate that the character does not include any spaces regardless of settings.
[Other example](https://github.com/fsprojects/fantomas/blob/4d038e57127e4e1a19c595ef57ea6ab7ef7b677a/src/Fantomas/Context.fs#L479)

src/Fantomas/CodePrinter.fs
```diff
@@ -681,48 +681,128 @@ and genMemberBinding astContext b =
...
+                    if ctx.Config.AlternativeLongMemberDefinitions then
+                        (genName
+                         +> genParameters
+                         +> indent
+                         +> sepNln
+                         +> sepColon
+                         +> genType astContext false t
+                         +> sepNln
+                         +> sepEqFixed
+                         +> unindent) ctx
```
deviousasti, line 746: Can avoid these parenthesis on each branch here, both return a function.

nojaf: I'm not really a fan of using the `<| ctx`, I'm guessing you meant:
```fsharp
                let longExpr ctx =
                    if ctx.Config.AlternativeLongMemberDefinitions then
                        genName
                         +> genParameters
                         +> indent
                         +> sepNln
                         +> sepColon
                         +> genType astContext false t
                         +> sepNln
                         +> sepEqFixed
                         +> unindent
                    else
                        genName +>
                         atCurrentColumn (genParameters
                                          +> sepNln
                                          +> onlyIf isSingleTuple (!- " ")
                                          +> sepColon
                                          +> genType astContext false t
                                          +> sepEq)
                    <| ctx
```
here?

deviousasti: Me neither. I usually use a let binding, then apply that.

nojaf:
```fsharp
                let longExpr ctx =
                    let expr =
                        if ctx.Config.AlternativeLongMemberDefinitions then
                            genName
                            +> genParameters
                            +> indent
                            +> sepNln
                            +> sepColon
                            +> genType astContext false t
                            +> sepNln
                            +> sepEqFixed
                            +> unindent
                        else
                            genName +>
                            atCurrentColumn (genParameters
                                             +> sepNln
                                             +> onlyIf isSingleTuple (!- " ")
                                             +> sepColon
                                             +> genType astContext false t
                                             +> sepEq)
                    expr ctx
```

deviousasti: That looks very clean. :)

Issue #944, opened 2020-07-02
FSI file has "abstract" stripped


Smaug123:
[...]

PR #967, opened 2020-07-13
Fix 944

nojaf:
So the problem we have with #944 is that the trivia detects two keywords and assigns them both to TriviaContentItSelf.
And so the last keyword (`member` in this case) wins.

I've added a rather targeted workaround to avoid `abstract` being replaced by `member` in `SynMemberSig.Member`.
I think this will make your code compile again, however, in the future it would be nice to preserve both keywords.
This fix might suffice short term...

### DIFF
--- a/src/Fantomas/Trivia.fs
+++ b/src/Fantomas/Trivia.fs
@@ -282,7 +282,17 @@ let private addTriviaToTriviaNode
 
     | { Item = Keyword({ Content = keyword} as kw); Range = range } when (keyword = "override" || keyword = "default" || keyword = "member" || keyword = "abstract") ->
         findMemberDefnMemberNodeOnLine triviaNodes range.StartLine
-        |> updateTriviaNode (fun tn -> tn.ContentItself <- Some (Keyword(kw))) triviaNodes
+        |> updateTriviaNode (fun tn ->
+            match tn.Type, tn.ContentItself with
+            | MainNode("SynMemberSig.Member"), Some( Keyword ({ Content = existingKeyword })) when (existingKeyword = "abstract" && keyword = "member") ->
+                // edge case when for a signature file with:
+                // type Foo =
+                //    abstract member Bar : Type
+                //
+                // here we don't want to override the abstract keyword
+                ()
+            | _ -> tn.ContentItself <- Some (Keyword(kw))
+        ) triviaNodes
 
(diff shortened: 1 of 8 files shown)


src/Fantomas/Trivia.fs
```diff
@@ -282,7 +282,17 @@ let private addTriviaToTriviaNode
 
     | { Item = Keyword({ Content = keyword} as kw); Range = range } when (keyword = "override" || keyword = "default" || keyword = "member" || keyword = "abstract") ->
         findMemberDefnMemberNodeOnLine triviaNodes range.StartLine
-        |> updateTriviaNode (fun tn -> tn.ContentItself <- Some (Keyword(kw))) triviaNodes
+        |> updateTriviaNode (fun tn ->
+            match tn.Type, tn.ContentItself with
+            | MainNode("SynMemberSig.Member"), Some( Keyword ({ Content = existingKeyword })) when (existingKeyword = "abstract" && keyword = "member") ->
```
knocte, line 287: nit: you can remove some unneeded parenthesis here (so that the line is not so long):
```
            | MainNode "SynMemberSig.Member", Some( Keyword { Content = existingKeyword }) when (existingKeyword = "abstract" && keyword = "member") ->
```
Maybe can also remove the parens after `when`? not sure

nojaf:
I'm able to restore both keywords now.

src/Fantomas/Trivia.fs
```diff
@@ -282,7 +282,25 @@ let private addTriviaToTriviaNode
 
     | { Item = Keyword({ Content = keyword} as kw); Range = range } when (keyword = "override" || keyword = "default" || keyword = "member" || keyword = "abstract") ->
         findMemberDefnMemberNodeOnLine triviaNodes range.StartLine
-        |> updateTriviaNode (fun tn -> tn.ContentItself <- Some (Keyword(kw))) triviaNodes
+        |> updateTriviaNode (fun tn ->
+            match tn.Type, tn.ContentItself with
+            | MainNode ("SynMemberSig.Member"), Some (Keyword ({ Content = existingKeywordContent } as token)) when existingKeywordContent =
```
Smaug123, line 287: Oh my :P the actual content of the code looks fine to me, and I know it's entirely not my place to comment on code layout in a codebase I barely know, but a 120-space indent seems a bit much to me!

nojaf: Yeah, I know, but uh the thing is that I formatted it with Fantomas, so yeah 🙈.
I guess we can later use it as sample to improve.

PR #987, opened 2020-07-24
Extract EditorConfig and FakeHelpers

### LINKED ISSUE #986: Extract editorconfig code to separate NuGet package.
## Intro

The merging of the [Fantomas.EditorConfig.Core](https://www.nuget.org/packages/Fantomas.EditorConfig.Core) (which is a fork of [editorconfig](https://www.nuget.org/packages/editorconfig/)) into Fantomas itself is not working out for JetBrains.
They need to strong name the dll and that has proven to be difficult.

## Proposal

I suggest we create a new project `Fantomas.EditorConfig` (that depends on the regular editorconfig package) and move the `CodeFormatter.ReadConfiguration` and `CodeFormatter.TryReadConfiguration` over there.
Fantomas.CoreGlobalTool would then use this package.

## Impact

This will impact mostly editors and maybe library authors.
Breaking changes:
* `CodeFormatter.ReadConfiguration` and `CodeFormatter.TryReadConfiguration` will move.
* FakeHelpers will require a configuration again.

Pinging @auduchinok @deviousasti @baronfel @Krzysztof-Cieslak @jindraivanek

nojaf:
Extracted `EditorConfig` and `FakeHelpers` to `Fantomas.Extras`.
It was easier to move the FakeHelpers as well as `formatFilesAsync files` and `formatCode files` would otherwise be up the consumer to figure out the config for each individual file.

To read configuration would now be:
`Fantomas.Extras.EditorConfig.readConfiguration` and `Fantomas.Extras.EditorConfig.tryReadConfiguration`.

Please let me know what you think about this.
We could go for also go for a signature file thing similar to [CodeFormatter](https://github.com/fsprojects/fantomas/blob/6df62f3e9dd32d5ea310d0145d22f2b57b44210f/src/Fantomas/CodeFormatter.fsi). 
I personally think we can be a bit less strict with the Fantomas.Extras package.

Fixes #986


nojaf:
We can still discuss details for `Fantomas.Extras`, I merged to move forward with Rider integration.

Issue #501, opened 2019-10-08
Improve formatting


cmeeren:
(Sorry for the non-descriptive title)

### Description

The following is hand-crafted, and I find it aesthetically pleasing and tidy:

```f#
module Program

open Microsoft.AspNetCore.Hosting
open Microsoft.Extensions.Hosting
open Serilog
open Startup

[<EntryPoint>]
let main args =
  Host
    .CreateDefaultBuilder(args)
    .ConfigureWebHostDefaults(fun builder ->
      builder
        .CaptureStartupErrors(true)
        .UseSerilog(dispose = true)
        .UseStartup<Startup>()
      |> ignore
    )
    .Build()
    .Run()
  0
```

Here is how Fantomas formats it, which I find rather messy:

```f#
module Program

open Microsoft.AspNetCore.Hosting
open Microsoft.Extensions.Hosting
open Serilog
open Startup

[<EntryPoint>]
let main args =
  Host.CreateDefaultBuilder(args)
      .ConfigureWebHostDefaults(fun builder ->
      builder.CaptureStartupErrors(true).UseSerilog(dispose = true).UseStartup<Startup>
        () |> ignore).Build().Run()
  0
```

There are several issues here:

* I'm not fond of the (initial) fluent calls to `Host` being aligned the way they are. If `Host` had a much longer name, all the interesting code would be pushed far to the right, leaving little space and yielding many unneeded line breaks.
* The body of `ConfigureWebHostDefaults` is indented on the same level as the line above. It should be indented an additional level.
* There's a weird line break after `UseStartup<Startup>` with the `unit` arg being on the next line
* `.Build()` and `.Run()` should be on the same level as the other fluent calls to `Host`

### Repro code

[...]

PR #1027, opened 2020-08-21
Introduce setting to limit max chained expressions

nojaf:
WIP but this would improve #501.
Not sure all cases are covered so far.
Feedback is welcome.

This would be a post v4 thing though.


nojaf:
Closing in favour of #1142

Issue #1143, opened 2020-09-18
Add option to make expressions multiline based on number of subexpressions rather than character length


Fizzixnerd:
I propose we add options for controlling when to make certain constructs multiline based on the logical size of those constructs (as opposed to their character width).  See below for examples.

The existing way of Fantomas deals with this problem is using character width as the primary method of determining when to split expressions on multiple lines.

## Pros and Cons 

The advantages of making this adjustment to Fantomas are that code becomes more consistent stylistically, especially for similarly structured expressions:

```fsharp
// input
let xs = [ 1; 2; 3; 4; 5; 6 ]
let ys = [ "one"; "two"; "three"; "four"; "five"; "six" ]
let zs = [ 1 .. 10 ]
ys |> List.zip zs |> List.reverse |> someFunctionExpr
ys |> List.zip xs |> List.head |> someFunctionExpr

// current Fantomas output
let xs = [ 1; 2; 3; 4; 5; 6 ]

let ys =
    [ "one"
      "two"
      "three"
      "four"
      "five"
      "six" ]

let zs = [ 1 .. 10 ]

ys
|> List.zip zs
|> List.reverse
|> someFunctionExpr

ys |> List.zip xs |> List.head |> someFunctionExpr

// proposed Fantomas output
let xs =
    [ 1
      2
      3
      4
      5
      6 ]

let ys =
    [ "one"
      "two"
      "three"
      "four"
      "five"
      "six" ]

let zs = [ 1 .. 10 ]

ys
|> List.zip zs
|> List.reverse
|> someFunctionExpr

ys
|> List.zip xs
|> List.head
|> someFunctionExpr
```

I (and the people I work with) believe these "parallel expressions" being formatted differently due to line length makes the code harder to understand on a structural level.

The disadvantages of making this adjustment to Fantomas are the added complexity of supporting more different output formats, as well as the configuration options to enable the alternative formatting behavior.

### Records

Desired Formatting
```fsharp
// desired to be single line, since only one field
type A = { x: int }

let a = { x = 0 }

// desired to be multiline, since multiple fields
type B =
    { y: int
      z: string }

// desired to be multiline, since multiple fields
let b =
   { y = 0
     z = "" }

// desired to be single line, even though long because single field
let a' = { a with x = AReallyLongExpressionThatIsMuchLongerThan50Characters }

// desired to be multiline, even though short because multiple fields
let b' =
    { b with
          y = b.y + 1
          z = "hello" }
```

[...]

PR #1147, opened 2020-09-20
Add logically sized list and array multiline formatter

Fizzixnerd:
See #1143.

I ended up adding the cutoff size feature after all, since it was so easy.


src/Fantomas.Extras/EditorConfig.fs
```diff
@@ -57,6 +61,9 @@ let configToEditorConfig (config: FormatConfig): string =
             sprintf "%s=%s" (toEditorConfigName k) (if b then "true " else "false")
             |> Some
         | :? System.Int32 as i -> sprintf "%s=%d" (toEditorConfigName k) i |> Some
+        | :? MultilineFormatterType as mft ->
+            sprintf "%s=\"%s\"" (toEditorConfigName k) (MultilineFormatterType.toConfigString mft)
```
Fizzixnerd: Whoops, this shouldn't have quotes

Fizzixnerd (commented):
Fixed.

nojaf (changes requested):
Hey @Fizzixnerd, 

This is a good first attempt!

In general, I'm not a fan of the whole `LogicalSize` terminology. I hardly see what is "logical" about any of it.
You wish to control based on the number of items inside the list expression, that is a stylistic decision nothing more nothing less. I'm going to have to ask you to rename this to something that `AmountOfExpressions` or `NumberOfItems`, something more along those lines.

Please also include some more tests for other combinations:
- fsharp_multiline_block_brackets_on_same_column = true
- fsharp_single_argument_web_mode = true
- add unit tests for the .editorconfig parsing in [FormatConfigEditorConfigurationFileTests](https://github.com/fsprojects/fantomas/blob/master/src/Fantomas.Tests/FormatConfigEditorConfigurationFileTests.fs)

This is for regression sake, just to sure that everything keeps working as expected over time.

I'll add some other feedback as code comments.

Keep it up!

src/Fantomas/Context.fs
```diff
@@ -27,6 +27,10 @@ type ShortExpressionInfo =
         - x.StartColumn > x.MaxWidth // expression is not too long according to MaxWidth
         || (currentColumn > maxPageWidth) // expression at current position is not going over the page width
 
+type Size =
+    | CharacterWidth of width: Num * maxWidth: Num
```
nojaf, line 31: Why does this have two values?

src/Fantomas/Context.fs
```diff
@@ -696,6 +700,13 @@ let private shortExpressionWithFallback (shortExpression: Context -> Context)
 let internal isShortExpression maxWidth (shortExpression: Context -> Context) (fallbackExpression) (ctx: Context) =
     shortExpressionWithFallback shortExpression fallbackExpression maxWidth None ctx
 
+let internal isSmallExpression size (smallExpression: Context -> Context) fallbackExpression (ctx: Context) =
+    match size with
+    | CharacterWidth (_width, maxWidth) -> isShortExpression maxWidth smallExpression fallbackExpression ctx
+    | LogicalSize (size, maxSize) ->
+        let effectiveMaxWidth = if size > maxSize then 0 else Num.max
+        isShortExpression effectiveMaxWidth smallExpression fallbackExpression ctx
```
nojaf, line 708: No need to call `isShortExpression` here right? if `size > maxSize` you can just `fallbackExpression ctx` otherwise `smallExpression`.

src/Fantomas/FormatConfig.fs
```diff
@@ -9,6 +9,23 @@ type FormatException(msg: string) =
 
 type Num = int
 
+type MultilineFormatterType =
+    | CharacterWidth
+    | LogicalSize
+    static member toConfigString =
```
nojaf, line 15: Nitpick: should have a capital letter I think. Did you ran fsharplint on this?
This is not yet part of our CI system but you can restore the dotnet tools (`dotnet tool restore`)
And run `dotnet fsharplint lint fantomas.sln`.

src/Fantomas/Context.fs
Fizzixnerd: I think it is necessary, because `isShortExpression` calls `shortExpressionWithFallback` which contains the logic to make it multiline if it's not under `MaxLineLength`.

Fizzixnerd:
> Hey @Fizzixnerd,
> 
> This is a good first attempt!
> 
> In general, I'm not a fan of the whole `LogicalSize` terminology. I hardly see what is "logical" about any of it.
> You wish to control based on the number of items inside the list expression, that is a stylistic decision nothing more nothing less. I'm going to have to ask you to rename this to something that `AmountOfExpressions` or `NumberOfItems`, something more along those lines.

While I'm happy to rename it, I wanted to clarify that my intention was not to claim that this size is the "more logical" one to use, or that the character width is "illogical"!  I meant that, in terms of the _syntax/term logic_, `[e1; e2; e3]` is "larger" than `[e1; e2]`.  `AmountOfExpressions` I think would be inappropriate because ideally we would keep the same name for each of the future changes, and the records change also affects type definition statements, which do not contain expressions.  I can change it to `NumberOfItems`.

> 
> Please also include some more tests for other combinations:
> 
> * fsharp_multiline_block_brackets_on_same_column = true
> * fsharp_single_argument_web_mode = true
> * add unit tests for the .editorconfig parsing in [FormatConfigEditorConfigurationFileTests](https://github.com/fsprojects/fantomas/blob/master/src/Fantomas.Tests/FormatConfigEditorConfigurationFileTests.fs)
> 
> This is for regression sake, just to sure that everything keeps working as expected over time.

I will add more tests.

knocte:
> I meant that, in terms of the syntax/term logic,

How about LogicSize instead of LogicalSize/NumberOfItems.

src/Fantomas/FormatConfig.fs
Fizzixnerd: I had not run the linter on it, but even when I did there was no warning associated with these two.  I think it thinks it is okay because it is a `static` member, so is treated the same as a module-local function, and so can be camelCase.

Would you like me to change it anyways?

src/Fantomas/Context.fs
Fizzixnerd: It was for uniformity with the other case, but I can remove it since it turns out never to be used.

Fizzixnerd: Changed.

Fizzixnerd:
> How about LogicSize instead of LogicalSize/NumberOfItems.

I think to avoid confusion `NumberOfItems` is fine.

src/Fantomas/FormatConfig.fs
nojaf: Yes please, the linter doesn't mention it as I was expecting but all the other `static member` occurrences are uppercased.

src/Fantomas.Tests/FormatConfigEditorConfigurationFileTests.fs
```diff
@@ -202,3 +202,24 @@ let ``print default editorconfig settings`` () =
     FormatConfig.Default
     |> EditorConfig.configToEditorConfig
     |> printfn "%s"
+
+[<Test>]
```
nojaf, line 206: Could you also add a test with only `fsharp_max_array_or_list_width` as single option, just to prove that the existing behavior still works as expected and does not need the `fsharp_array_or_list_multiline_formatter` setting.

src/Fantomas.Tests/NumberOfItemsTests.fs
```diff
@@ -0,0 +1,268 @@
+module Fantomas.Tests.LogicalSize
```
nojaf, line 1: Rename module please.

src/Fantomas.Tests/NumberOfItemsTests.fs
```diff
@@ -0,0 +1,268 @@
...
+    longValueThatIsALotOfCharactersSoooooLong
+]
+
+h [
+    longValueThatIsALotOfCharactersSoooooLong
+    longValueThatIsALotOfCharactersSoooooLong
+]
+"""
+
+let ``number of items sized Elmish lists with single argument web mode and multiline block brackets on same column are formatted properly`` () =
```
nojaf, line 242: Add the `[<Test>]` attribute please, I was already very puzzled why this could work 😅.
The test is not running, so that is why...

src/Fantomas.Tests/NumberOfItemsTests.fs
```diff
@@ -0,0 +1,268 @@
...
+
+h [
+    longValueThatIsALotOfCharactersSoooooLong
+    longValueThatIsALotOfCharactersSoooooLong
+  ] [
+    longValueThatIsALotOfCharactersSoooooLong
+]
+"""
+
+let ``number of items sized Elmish lists with single argument web mode are formatted properly`` () =
```
nojaf, line 215: Add `[<Test>]`

src/Fantomas/Context.fs
nojaf: If `NumberOfItems` is selected 
```fsharp
    | NumberOfItems (items, maxItems) ->
        if items > maxItems then
            fallbackExpression ctx
        else
            smallExpression ctx
```
suffices.

src/Fantomas.Tests/NumberOfItemsTests.fs
```diff
@@ -266,3 +260,18 @@ h [
...
+let ``long expressions with number of items set to 3 will get split on max line length`` () =
+    formatSourceString false """
+[ longValueThatIsALotOfCharactersSoooooLongAndlongValueThatIsALotOfCharactersSoooooLongAndlongValueThatIsALotOfCharactersSoooooLong; longValueThatIsALotOfCharactersSoooooLong ]
+    """
+        { config with
+              ArrayOrListMultilineFormatter = NumberOfItems
+              MaxArrayOrListNumberOfItems = 3 }
+    |> prepend newline
+    |> should equal """
+[ longValueThatIsALotOfCharactersSoooooLongAndlongValueThatIsALotOfCharactersSoooooLongAndlongValueThatIsALotOfCharactersSoooooLong
```
nojaf, line 274: I don't think this is valid behaviour to be honest if the user explicitly says to only going multiline when three items are present in the list, the length of each individual item should not matter.

This would be reported as an inconsistency when put together with
`let x = [ a; b ]`

src/Fantomas/FormatConfig.fs
Fizzixnerd: Will do.

src/Fantomas.Tests/NumberOfItemsTests.fs
Fizzixnerd: Renamed.

src/Fantomas.Tests/NumberOfItemsTests.fs
Fizzixnerd: In the original Issue (https://github.com/fsprojects/fantomas/issues/1143), the rule given was

> If the number of elements in the expression is > 1 then force a multiline version. If the number of elements in the expression = 1 then force a single line version, unless exceeding the maximum line length.

If we replace `1` with `n` to account for the configurable cutoff, then the expression should be made multiline if it exceeds the maximum line length.  This is what this test was meant to catch, and why one of the expressions is so excessively long: I wanted to exceed the maximum line length without exceeding the cutoff for number of items (3).  This test fails if we replace the `isShortExpression` call in your above review with `if size > maxSize then fallbackExpression ctx else smallExpression ctx`, which is why I added it.

nojaf: Please update the documentation then because I already forgot that detail from the original issue.
Now it makes more sense why you still wanted the check.
Perhaps `expressionFitsOnRestOfLine` can be used instead.

Fizzixnerd: Updated.

src/Fantomas.Tests/NumberOfItemsTests.fs
Fizzixnerd: Added

src/Fantomas.Tests/NumberOfItemsTests.fs
Fizzixnerd: Added.

src/Fantomas/Context.fs
Fizzixnerd: Updated to use `expressionFitsOnRestOfLine` as discussed below

src/Fantomas.Tests/FormatConfigEditorConfigurationFileTests.fs
Fizzixnerd: Added.

nojaf: I can't seem to find this, did you added it somewhere else maybe?

src/Fantomas.Tests/NumberOfItemsTests.fs
```diff
@@ -0,0 +1,303 @@
+module Fantomas.Tests.NumberOfItems
```
nojaf, line 1: Another nitpick I'm afraid but the module name should end on the word `Tests` to match the file name.

src/Fantomas/CodePrinter.fs
```diff
@@ -52,6 +52,11 @@ type ASTContext =
           IsMemberDefinition = false
           IsInsideDotIndexed = false }
 
+let getListOrArrayExprSize ctx maxWidth xs =
```
nojaf, line 55: Could you move this function to `Context.fs` please?

nojaf:
Hey @Fizzixnerd, this is looking really good! Thanks for having this much patience with me.

src/Fantomas/CodePrinter.fs
Fizzixnerd: Moved

src/Fantomas.Tests/NumberOfItemsTests.fs
Fizzixnerd: Sorry about that, changed.

src/Fantomas.Tests/FormatConfigEditorConfigurationFileTests.fs
Fizzixnerd: Sorry, I misunderstood your meaning and had added it as a test in `NumberOfItemsTests.fs` for something different.  I have added the test now.

Fizzixnerd:
> Hey @Fizzixnerd, this is looking really good! Thanks for having this much patience with me.

Hey @nojaf  thanks, this has been pretty fun.  Thanks for being so considerate and allowing me to make these changes. 
 The people where I work are very happy with the results so far.  Looking forward to making the additional changes as we go!

nojaf (approved):

Issue #1167, opened 2020-09-26
Formatting of anonymous record types does not respect fsharp_multiline_block_brackets_on_same_column


Fizzixnerd:
[...]

PR #1169, opened 2020-09-26
Add number of items based record multiline formatter

Fizzixnerd:
See https://github.com/fsprojects/fantomas/issues/1143.

I originally wrote the bulk of the code back when I was still calling it "LogicalSize".  I have gone through the diffs looking to make sure I removed all references to that, but just be aware if I missed one that I'm not trying to sneak it in!

Currently the tests assume the current behaviour discussed in https://github.com/fsprojects/fantomas/issues/1167.  If that changes, then one of the tests here will need to change as well.  I have affixed a `FIXME` to the test in question that can be removed when that issue is closed, but please let me know if you want it done a different way.


nojaf:
@Fizzixnerd I'll review this thoroughly another time but I already want to point out that records have a broader scope.
At first glance, I didn't see any tests for signature files and nothing related to object expressions.
Please consider test-driven development for this project, it can help in not introducing changes without test coverage.

Fizzixnerd:
> @Fizzixnerd I'll review this thoroughly another time but I already want to point out that records have a broader scope.
> At first glance, I didn't see any tests for signature files and nothing related to object expressions.
> Please consider test-driven development for this project, it can help in not introducing changes without test coverage.

Thanks @nojaf.

Yes, I am sort of winging it with what needs to be changed.  I don't think I've ever explicitly used signature files or object expressions before, so it didn't occur to me they would be affected.  I will write tests first for this project from now on, but if I'm not aware of everything that could be affected I'm afraid it won't be enough.  Is there an issue/PR in the past that touched all aspects of records that I could use as a reference?  I obviously don't want to miss anything else.

nojaf:
> I don't think I've ever explicitly used signature files or object expressions

Yeah, I know I'm asking more than you might need yourself but it is to remain consistent.
I would take a look at [MultilineBlockBracketsOnSameColumnRecordTests.fs](https://github.com/fsprojects/fantomas/blob/e2a6c0a8fc572cfabba602bfd7aedb20ddb553e9/src/Fantomas.Tests/MultilineBlockBracketsOnSameColumnRecordTests.fs).

You can write tests for signature files by using `formatSourceString true`.

src/Fantomas.Tests/NumberOfItemsListOrArrayTests.fs
```diff
@@ -1,10 +1,14 @@
-module Fantomas.Tests.NumberOfItemsTests
+module Fantomas.Tests.NumberOfItemsListOrArrayTests
 
 open NUnit.Framework
 open FsUnit
 open Fantomas.Tests.TestHelper
 open Fantomas.FormatConfig
 
+// ****************
```
nojaf, line 8: This comment can be removed as you split the file into two modules.

src/Fantomas.Tests/NumberOfItemsRecordTests.fs
```diff
@@ -0,0 +1,982 @@
...
+let g (x: {| x: AReallyLongTypeThatIsMuchLongerThan40Characters |}) = x
+
+type A =
+    {| x: int
+       y: obj |}
+
+type B = {| x: AReallyLongTypeThatIsMuchLongerThan40Characters |}
+"""
+
+// FIXME: Change with https://github.com/fsprojects/fantomas/issues/1167
```
nojaf, line 957: Please also ignore this test for now. The result now is indeed not correct as stated in #1167.

nojaf (approved):

nojaf:
Great PR @Fizzixnerd! And thanks for reporting the inconsistencies you found along the way.

Fizzixnerd:
No problem, thanks for fixing those nitpicks.

PR #1177, opened 2020-09-29
Add infix operator expression number of items multiline formatter

Fizzixnerd:
See https://github.com/fsprojects/fantomas/issues/1143.

Final change from the above issue.  I'm thinking that getting the basic functionality in and then refining it as needed over time is probably the best course of action; the current code gives the strange results I mentioned in the above issues when involving mathematical operators, but so can the `CharacterWidth` formatter, so I don't think it's introducing anything worse.  Let me know what you think.


nojaf:
Hey @Fizzixnerd, I finally found some time to take a look at this PR. The code looks OK but I'm not convinced of this setting to be honest.

Tests like
```fsharp
[<Test>]
let ``mathematical expressions`` () =
    formatSourceString false """
let w = x + y + z
let w = x - y - z
    """ config
    |> prepend newline
    |> should equal """
let w =
    x
    + y
    + z
let w =
    x
    - y
    - z
"""
```

seem to contract the original text in the feature request (#1143).

So I like to take a step back and go return to the drawing board here.
What problem are we trying to fix here?
If you format your codebase today with latest stable Fantomas. What results are not satisfying?
Maybe we could come up with something that only impacts the [newLineInfixOps](https://github.com/fsprojects/fantomas/blob/master/src/Fantomas/Context.fs#L939).

Fizzixnerd:
> Hey @Fizzixnerd, I finally found some time to take a look at this PR. The code looks OK but I'm not convinced of this setting to be honest.

Looking back, I'm in agreement.

> Tests like
> 
> ```fsharp
> [<Test>]
> let ``mathematical expressions`` () =
>     formatSourceString false """
> let w = x + y + z
> let w = x - y - z
>     """ config
>     |> prepend newline
>     |> should equal """
> let w =
>     x
>     + y
>     + z
> let w =
>     x
>     - y
>     - z
> """
> ```
> 
> seem to contract the original text in the feature request (#1143).

Yes.

> So I like to take a step back and go return to the drawing board here.
> What problem are we trying to fix here?
> If you format your codebase today with latest stable Fantomas. What results are not satisfying?
> Maybe we could come up with something that only impacts the [newLineInfixOps](https://github.com/fsprojects/fantomas/blob/master/src/Fantomas/Context.fs#L939).

Problems:
- Expressions such as
  ```fsharp
  x |> f |> g |> h
  
  ys
  |> List.choose (fun y -> if y = 3 then Ok y else None)
  ```
  Are undesirable to us.  We would prefer
  ```fsharp
  x
  |> f
  |> g
  |> h

  ys |> List.choose (fun y -> if y = 3 then Ok y else None)
  ```
  In short: for pipeline-like operators we want multiline after `n` operators, and single line for `n` and less.  I'm currently leaning on `n = 2` in this case, though that's a detail that I think is not so important.

- Expressions such as
  ```fsharp
  w + x + y - z
  ```
  We want to remain single line as much as possible.

Best option, I think, is to do as your suggest with the newLineInfixOps, and leave the normal behavior for everything else.  I will make a change and try it out on our internal test code.

Fizzixnerd:
Looking at the tests and the internal code we are using to trial Fantomas, I think the latest changes properly solve the problem we were having.  I have changed the names of the config options to reflect their new purpose.

docs/Documentation.md
```diff
@@ -452,7 +452,8 @@ else c.ToString()
 ### fsharp_max_infix_operator_expression
```
nojaf, line 452: Would it not be easier to understand if this setting doesn't apply for the newline infix operators?
Then you don't need to have any `*_multiline_formatter` setting in place and the default behaviour sorta is respected.

docs/Documentation.md
```diff
@@ -470,6 +471,80 @@ let WebApp =
...
+    >>= (fun x -> [ x * x ])
+
+let eigthPower =
+    [ 1; 2; 3 ]
+    |> List.map (fun x -> x * x)
+    >>= (fun x -> [ x * x ])
+    |> List.choose (fun x -> Some (x * x))
+```
+
+`{ defaultConfig with MaxNewlineInfixOperatorExpressionNumberOfItems = 2; NewlineInfixOperatorExpressionMultilineFormatter = 2 }`
```
nojaf, line 496: `NewlineInfixOperatorExpressionMultilineFormatter = 2` should be `NewlineInfixOperatorExpressionMultilineFormatter  = NumberOfItems` here right? Or am I missing something?

src/Fantomas.Tests/NumberOfItemsInfixOperatorExpressionTests.fs
```diff
@@ -0,0 +1,277 @@
+module Fantomas.Tests.NumberOfItemsInfixOperatorExpressionTests
```
nojaf, line 1: MultilineInfixOperators*

src/Fantomas/CodePrinter.fs
```diff
@@ -1792,8 +1818,11 @@ and genExpr astContext synExpr =
 
     | InfixApp (operatorText, operatorExpr, e1, e2) ->
         fun ctx ->
-            isShortExpression
-                ctx.Config.MaxInfixOperatorExpression
+            let size =
```
nojaf, line 1821: Does this not miss any further check whether the infix operator is a multiline one?

src/Fantomas.Tests/NumberOfItemsInfixOperatorExpressionTests.fs
```diff
@@ -0,0 +1,277 @@
...
+    |> prepend newline
+    |> should equal """
+let private distanceBetweenTwoPoints (latA, lngA) (latB, lngB) =
+    if latA = latB && lngA = lngB then
+        0.
+    else
+        let theta = lngA - lngB
+
+        let dist =
+            Math.Sin(deg2rad (latA)) * Math.Sin(deg2rad (latB))
```
nojaf, line 184: This doesn't seem correct to me, `Math.Sin(deg2rad (latA)) * Math.Sin(deg2rad (latB))` does not have any newline infix involved and is over 50 characters long.
image

src/Fantomas.Tests/NumberOfItemsInfixOperatorExpressionTests.fs
```diff
@@ -0,0 +1,277 @@
...
+"""
+
+[<Test>]
+let ``modulo operator on same line, 780`` () =
+    formatSourceString false """let hasUnEvenAmount regex line = (Regex.Matches(line, regex).Count - Regex.Matches(line, "\\\\" + regex).Count) % 2 = 1
+"""  config
+    |> prepend newline
+    |> should equal """
+let hasUnEvenAmount regex line =
+    (Regex.Matches(line, regex).Count - Regex.Matches(line, "\\\\" + regex).Count) % 2 = 1
```
nojaf, line 102: `(Regex.Matches(line, regex).Count - Regex.Matches(line, "\\\\" + regex).Count)` should be split here I think.

src/Fantomas.Tests/NumberOfItemsInfixOperatorExpressionTests.fs
```diff
@@ -0,0 +1,277 @@
...
+            aaaaaaaaaaaaaaaaaaaaaaaa
+            * bbbbbbbbbbbbbbbbbbbbbbbbb
+            + (ccccccccccccccccccccccccc
+               * ddddddddddddddddddddddd
+               * eeeeeeeeeeeeeeeeeeeeeee)
+"""  config
+    |> prepend newline
+    |> should equal """
+let dist =
+    aaaaaaaaaaaaaaaaaaaaaaaa * bbbbbbbbbbbbbbbbbbbbbbbbb
```
nojaf, line 255: `aaaaaaaaaaaaaaaaaaaaaaaa * bbbbbbbbbbbbbbbbbbbbbbbbb` should be split here.

nojaf (changes requested):
Hey @Fizzixnerd,
Apologies for the slow response, I've been preoccupied with other life I'm afraid.
The PR looks good yet I would like to emphasis that this only impact those few newline infix operators. In some places, I still found a bit of mix, so I'd like you to clear those up.

nojaf:
Hey @Fizzixnerd, I took the liberty of completing this PR. We can still discuss things if you like to.
I made some minor changes and added a tad more focus that the settings only impacts those multiline infix operators. 
I'll release an alpha shortly so you can try it out.

Fizzixnerd:
Hey @nojaf, thanks for that, I have been really busy and was trying to find the time on the weekend but just couldn't.  Sorry about that.

nojaf:
No worries, I totally get that. You can try the changes in https://www.nuget.org/packages/fantomas-tool/4.3.0-alpha-001.

Fizzixnerd:
Hi @nojaf, I think I found a couple of issues. Will try to clean them up into a simple repro tonight, but just giving a heads up.

nojaf:
Hey @Fizzixnerd, any luck with reproducing those issues?
And did you see my comment in https://github.com/fsprojects/fantomas/issues/1143#issuecomment-719666763
I'm waiting to release 4.3 stable until we got everything sorted out of you PRs, so could you take a look at this?
Thanks!

Issue #908, opened 2020-06-11
Shouldn't remove space after colon


knocte:
[...]

PR #1192, opened 2020-10-12
bugfix: Shouldn't remove space after colon

aarani:
fixes fsprojects/fantomas#908


src/Fantomas/CodePrinter.fs
```diff
@@ -481,27 +481,30 @@ and genAttribute astContext (Attribute (s, e, target)) =
     match e with
     // Special treatment for function application on attributes
     | ConstExpr (Const "()", _) ->
-        !- "[<" +> opt sepColonFixed target (!-)
+        !- "[<"
+        +> opt sepColonWithEndingSpaceFixed target (!-)
```
nojaf, line 485: I would actually go with `sepColon` in this case instead of introducing `sepColonWithEndingSpaceFixed`.

nojaf (changes requested):

src/Fantomas/CodePrinter.fs
aarani: oh sure, sorry

nojaf (approved):

nojaf:
Thanks @aarani

Issue #1198, opened 2020-10-18
Disable Elmish syntax


nojaf:
I propose we add a setting to disable the [Elmish syntax ](https://github.com/fsprojects/fantomas/blob/master/docs/Formatting-Elmish-code.md) improvement.
This came up in #1189 and I hear other people mentioning this as well.

The existing way of Fantomas deals with this problem is an all or nothing situation when the Syntax tree matches the Elmish patterns.

Initially, I'd go for a disable Elmish setting, after extensive feedback and for the next major version we might even flip this around that you need to enable Elmish formatting.

## Pros and Cons 

The advantages of making this adjustment to Fantomas are more consistent code when not doing anything Fable/Elmish related.
The setting could be enabled/disabled for specific files or folder using a `.editorconfig`.

The disadvantages of making this adjustment to Fantomas are more code to maintain and users need to be aware that of this setting. This could be a bit confusing.
Another contra I see here is that the `fsharp_max_elmish_width` should have no effect when `fsharp_disable_elmish=true`.
This again might be just a tad confusing for end-users.

## Examples

1.

```fsharp
let v =
    someFunctionHere [parameters] [yetMoreParameters]
    :?> _
```

currently formatted

```fsharp
let v =
    someFunctionHere [ parameters ] [
        yetMoreParameters
    ] :?> _
```

with setting should be

```fsharp
let v =
    someFunctionHere [ parameters ] [ yetMoreParameters ] :?> _
```

2.

```fsharp
InstanceObject.make [a; b; c]
```

currently format with `fsharp_max_elmish_width=10`

```fsharp
InstanceObject.make [ a
                      b
                      c ]
```

with setting should be

```fsharp
InstanceObject.make [ a; b; c ]
```

3.

```fsharp
let  f x =
    Case.fromCaseTimeline [ { mockClaim with x = 5 }
                              |> Event.claim "42"
                              |> Commit.caseCommit System.DateTime.Now ]
```

currently formatted

```fsharp
let f x =
    Case.fromCaseTimeline [ { mockClaim with x = 5 }
                            |> Event.claim "42"
                            |> Commit.caseCommit System.DateTime.Now ]
```

with setting should be

```fsharp
let f x =
    Case.fromCaseTimeline
        [ { mockClaim with x = 5 }
          |> Event.claim "42"
          |> Commit.caseCommit System.DateTime.Now ]

```

## Extra information

Estimated cost (XS, S, M, L, XL, XXL): S or M

[...]

PR #1199, opened 2020-10-18
Add DisableElmishSyntax setting

nojaf:
Initial attempt to solve #1198 
Left to do:

- [x] Add tests that cover all Elmish expression cases
- [x] Write documentation


knocte:
> Another contra I see here is that the fsharp_max_elmish_width should have no effect when fsharp_disable_elmish=true.
This again might be just a tad confusing for end-users.

Then how about renaming `fsharp_max_elmish_width` to `fsharp_max_elmish_width_if_enabled`?

nojaf:
Renaming a setting and not introducing a breaking change is though combination, not something I want at this moment.
Perhaps for the next major release (of which I have no intentions at the time of writing) but then again I don't think it adds that much value.

knocte:
Understand, I was suggesting something for having less confused users.

PR #1208, opened 2020-10-29
Sanitize strings that would lead to uneaven quotes in line

WalternativE:
- Adds two tests based on failing formatting
  behavior having hash directives
- Documentation of triage session in Fantomas & Friends

### DIFF
--- a/.fantomasignore
+++ b/.fantomasignore
@@ -1,1 +1,0 @@
-TokenParser.fs
--- a/.fantomasignore
+++ b/.fantomasignore
@@ -1,0 +1,1 @@
+src/Fantomas/TokenParser.fs
--- a/.fantomasignore
+++ b/.fantomasignore
@@ -1,1 +1,1 @@
-src/Fantomas/TokenParser.fs
+TokenParser.fs
--- a/src/Fantomas.Tests/StringTests.fs
+++ b/src/Fantomas.Tests/StringTests.fs
@@ -232,3 +232,23 @@ let ``newline in string`` () =
 
     List.length triviaNodes == 1
 """
+
+[<Test>]
+let ``double backslash in triple quote string`` () =
+    formatSourceString false "
+let a = \"\"\"\\\\\"\"\"
+"    config
+    |> prepend newline
+    |> should equal "
+let a = \"\"\"\\\\\"\"\"
+"
+
+[<Test>]
+let ``single string with compiler define`` () =
+    formatSourceString false """
+"#if FOO"
+"""  config
+    |> prepend newline
+    |> should equal """
+"#if FOO"
+"""
--- a/src/Fantomas.Tests/StringTests.fs
+++ b/src/Fantomas.Tests/StringTests.fs
@@ -224,10 +224,11 @@ let ``newline in string`` () =
     let triviaNodes =
         tokenize [] source
         |> getTriviaFromTokens
-        |> List.filter (fun { Item = item } ->
-            match item with
-            | StringContent ("\"\n\"") -> true
-            | _ -> false)
+        |> List.filter
+            (fun { Item = item } ->
+                match item with
+                | StringContent ("\"\n\"") -> true
+                | _ -> false)
 
     List.length triviaNodes == 1
 """
--- a/src/Fantomas.Tests/TestHelpers.fs
+++ b/src/Fantomas.Tests/TestHelpers.fs
@@ -289,7 +289,7 @@ type TemporaryFileCodeSample internal (codeSnippet: string) =
 
     do File.WriteAllText(filename, codeSnippet)
 
-    member _.Filename: string = filename
+    member __.Filename: string = filename
 
     interface IDisposable with
         member this.Dispose(): unit = File.Delete(filename)
--- a/src/Fantomas.Tests/TestHelpers.fs
+++ b/src/Fantomas.Tests/TestHelpers.fs
@@ -23,12 +23,7 @@ let private safeToIgnoreWarnings =
       "Identifiers containing '@' are reserved for use in F# code generation" ]
 
 let private isValidAndHasNoWarnings fileName source parsingOptions =
-    let allDefineOptions =
-        TokenParser.getOptimizedDefinesSets source
-        @ (TokenParser.getDefines source
-           |> List.map List.singleton)
-          @ [ [] ]
-        |> List.distinct
+    let allDefineOptions, _ = TokenParser.getDefines source
 
     allDefineOptions
     |> List.map (fun conditionalCompilationDefines ->
@@ -95,12 +90,12 @@ let formatSourceStringWithDefines defines (s: string) config =
         async {
             let! asts = CodeFormatterImpl.parse sharedChecker.Value parsingOptions formatContext
 
-            let ast =
-                Array.filter (fun (_, d) -> d = defines) asts
+            let ast, hashTokens =
+                Array.filter (fun (_, d, _) -> d = defines) asts
                 |> Array.head
-                |> fst
+                |> fun (ast, _, hashTokens) -> ast, hashTokens
 
-            return CodeFormatterImpl.formatWith ast defines formatContext config
+            return CodeFormatterImpl.formatWith ast defines hashTokens formatContext config
         }
         |> Async.RunSynchronously
         |> CodeFormatterImpl.addNewlineIfNeeded
@@ -137,8 +132,8 @@ let parse isFsiFile s =
     let fileName =
         if isFsiFile then "/tmp.fsi" else "/tmp.fsx"
 
-    CodeFormatter.ParseAsync
-        (fileName, SourceOrigin.SourceString s, FakeHelpers.createParsingOptionsFromFile fileName, sharedChecker.Value)
+    CodeFormatterImpl.createFormatContext fileName (SourceOrigin.SourceString s)
+    |> CodeFormatterImpl.parse sharedChecker.Value (FakeHelpers.createParsingOptionsFromFile fileName)
     |> Async.RunSynchronously
 
 let formatAST a s c =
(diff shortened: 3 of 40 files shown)


nojaf:
@Smaug123 we sorta went with a dirty hack to solve the problem. I'm going to revisit the solution and try and come up with something better ;)

nojaf:
So I've refactored the code a bit and tried to come up with a better way of detecting all the defines inside a raw string.
The algorithm is rather primitive were all characters in the string are processed while being aware in what kind of code block you are currently in (string, multiline code comment)

Lastly, I refactored the code so that the define detection dance is only performed once.

.fantomasignore
```diff
@@ -0,0 +1 @@
+TokenParser.fs
```
Smaug123, line 1: Why is this required?

src/Fantomas.Tests/StringTests.fs
```diff
@@ -232,3 +232,23 @@ let ``newline in string`` () =
 
     List.length triviaNodes == 1
 """
+
+[<Test>]
+let ``double backslash in triple quote string`` () =
+    formatSourceString false "
+let a = \"\"\"\\\\\"\"\"
```
Smaug123, line 239: Oh my

src/Fantomas.Tests/TestHelpers.fs
```diff
@@ -23,12 +23,7 @@ let private safeToIgnoreWarnings =
       "Identifiers containing '@' are reserved for use in F# code generation" ]
 
 let private isValidAndHasNoWarnings fileName source parsingOptions =
-    let allDefineOptions =
-        TokenParser.getOptimizedDefinesSets source
-        @ (TokenParser.getDefines source
-           |> List.map List.singleton)
-          @ [ [] ]
-        |> List.distinct
+    let allDefineOptions, _ = TokenParser.getDefines source
```
Smaug123, line 26: It was sad that there was a copy of this both here and in CodeFormatterImpl!

src/Fantomas/CodeFormatterImpl.fs
```diff
@@ -359,29 +354,26 @@ let isValidFSharpCode (checker: FSharpChecker) (parsingOptions: FSharpParsingOpt
             let! ast = parse checker parsingOptions formatContext
 
             let isValid =
-                ast |> Array.map fst |> Array.forall isValidAST
+                ast
+                |> Array.map (fun (a, _, _) -> a)
```
Smaug123, line 358: Obvious allocation reduction: `ast |> Array.forall (fun (a, _, _) -> isValidAST a)`. But I imagine this is unlikely ever to matter.

src/Fantomas/CodeFormatterImpl.fs
```diff
@@ -359,29 +354,26 @@ let isValidFSharpCode (checker: FSharpChecker) (parsingOptions: FSharpParsingOpt
...
         if String.IsNullOrWhiteSpace formatContext.Source
-        then None
-        else Some formatContext.Source
-
-    // Use '\n' as the new line delimiter consistently
-    // It would be easier for F# parser
-    let sourceCode = defaultArg input String.Empty
-    let normalizedSourceCode = String.normalizeNewLine sourceCode
+        then String.Empty, false
+        else formatContext.Source, true
```
Smaug123, line 372: This feels like a slightly broken domain. `sourceCode, hasInput` is really a `string option` that you're representing as a `string * bool`. This makes the `IsNullOrWhitespace` check on line 390 a bit harder to understand.

src/Fantomas/TokenParser.fs
```diff
@@ -90,91 +90,189 @@ let private createHashToken lineNumber content fullMatchedLength offset =
...
-            FullMatchedLength = fullMatchedLength } }
+            FullMatchedLength = String.length content } }
 
-type private SourceCodeState =
+type SourceCodeState =
     | Normal
-    | InsideSingleQuoteString
-    | InsideTripleQuoteString
+    | InsideString
+    | InsideTripleQuoteString of int
```
Smaug123, line 98: Could you put a docstring here indicating what the int is?

src/Fantomas/TokenParser.fs
```diff
@@ -90,91 +90,189 @@ let private createHashToken lineNumber content fullMatchedLength offset =
...
+        let hashChar = '#'
+        let spaceChar = ' '
+        let isDoubleQuoteChar = (=) doubleQuoteChar
+        let isBackSlashChar = (=) backSlashChar
+        let isNoBackSlashChar v = v <> backSlashChar
+
+        let tripleQuotes =
+            (doubleQuoteChar, doubleQuoteChar, doubleQuoteChar)
+
+        let isOpenParenChar = (=) openParenChar
```
Smaug123, line 122: Idle thought: could this all be represented better with active patterns? Instead of all these checks, you could have them all play nicely in a `match` statement.

src/Fantomas/TokenParser.fs
```diff
@@ -90,91 +90,189 @@ let private createHashToken lineNumber content fullMatchedLength offset =
...
+
+                { t with
+                      LineNumber = lineNumber
+                      TokenInfo = info })
+            |> fun rest ->
+                (createHashToken lineNumber hashContent offset)
+                :: rest
+
+        let sourceLength = String.length sourceCode
+        let lastIndex = sourceLength - 3
```
Smaug123, line 155: A comment here indicating what the `3` is referring to would be good

src/Fantomas/TokenParser.fs
```diff
@@ -90,91 +90,189 @@ let private createHashToken lineNumber content fullMatchedLength offset =
...
+        let defines = ResizeArray()
+        let mutable currentState = Normal
+        let mutable newlineIndexes = ResizeArray()
```
Smaug123, line 158: Pretty sure this won't need to be `mutable` (since the `ResizeArray` is already mutable).

If this were code in G-Research, I'd comment that this should be expressed immutably as a fold or by recursion; iterating over a list updating a state machine is a very suitable task for `fold`. But if imperative style is normal in Fantomas, then fine.

src/Fantomas/TokenParser.fs
```diff
@@ -260,15 +354,24 @@ let getDefineExprs sourceCode =
...
           |> BoolExpr.mergeBoolExprs maxSteps
           |> List.map snd with
     | [] -> [ [] ]
     | xs -> xs
 
+let getDefines sourceCode =
+    let hashTokens = getTokenizedHashes sourceCode
+
+    let defineCombinations =
+        getOptimizedDefinesSets hashTokens @ (getDefinesWords  hashTokens |> List.map List.singleton) @ [ [] ]
```
Smaug123, line 370: Hey, here's a formatting error (double space after `getDefinesWords`)! Sure we wanted that `.fantomasignore`? ;)

Smaug123, line 370: Hey, here's a formatting error (double space after `getDefinesWords`)! Sure we wanted that `.fantomasignore`? ;)

Smaug123 (commented):
I don't really understand this fully, but the shape of it certainly makes me think it looks better. (E.g. the fact that we parse out hash-tokens right at the beginning.)

The state machine is big and scary and I don't really feel like I understand it properly!

.fantomasignore
nojaf: I discovered https://github.com/fsprojects/fantomas/issues/1211 so that is why I needed to exclude this from the `fantomas --check` we do inside the build script. A necessary evil until the next version drops.

nojaf:
Thanks for the review @Smaug123, some very good remarks! I'll revisit this PR later this week.

src/Fantomas.Tests/StringTests.fs
nojaf: yup 🙈

src/Fantomas.Tests/TestHelpers.fs
nojaf: Indeed, although for the Benchmark it does not seem to have a noticeable impact that we now only call this once.
Unit tests do feel snappier though.

.fantomasignore
nojaf: After refactoring I do have the `elif` without `else` constructs anymore so I was able to remove the `.fantomasignore`.

src/Fantomas/CodeFormatterImpl.fs
nojaf: Good spot, never hesitate to mention things like this!

src/Fantomas/TokenParser.fs
nojaf: > Idle thought

You mean genius advice! The code to me is indeed more readable with the active patterns.

src/Fantomas/TokenParser.fs
nojaf: I went with an imperative style for no good reason. Refactoring to a `fold` was the right choice here.

nojaf:
@Smaug123 I'm ready for round two 😄.

src/Fantomas/TokenParser.fs
```diff
@@ -90,91 +89,213 @@ let private createHashToken lineNumber content fullMatchedLength offset =
...
+                |> Seq.take take
+                |> Seq.forall isSpace
+
+            if leadingCharactersBeforeLastNewlineAreSpaces then
+                let skip =
+                    if lastNewlineIdx = -1 then 0 else lastNewlineIdx + 1 // zero when the source starts with an #
+
+                let currentLine =
+                    sourceCode
+                    |> Seq.skip skip
```
Smaug123, line 183: The use of `Seq.skip` always terrifies me, because it makes it very easy to write things which are quadratic in complexity; but strings are easy to seek around in, so that's fine.

Smaug123 (approved):
I can't say I'm 100% convinced of its correctness, but I don't see anything obviously wrong, and if no tests are broken then that's good enough for me!

nojaf:
Sold!

PR #1228, opened 2020-11-07
Fantomas.Tests: fix build with VS4Mac

knocte:
Not sure why, but when trying to build this project with VS4Mac v8.6.7.2
(which uses the following underneath: mono 6.10.0.104 & fsharpc 10.2.3 with
F# 4.5 support) the following error was being reported:

```
Error FS0010: Unexpected symbol '.' in member definition. Expected 'with', '=' or other token. (FS0010) (Fantomas.Tests)
```


nojaf (approved):

nojaf:
A single underscore is an F# 4.7 thing I think. I hope VS4Mac catches up soon.

Issue #1231, opened 2020-11-08
[Feature request] Add support for editorconfig-defined newline separator characters


baronfel:
I propose we add support for the `end_of_line` setting in editorconfig, which is a [core setting](https://editorconfig-specification.readthedocs.io/en/latest/#supported-pairs) that allows a user to select the newline character to use in their code, centralizing that in one location.

The existing way Fantomas deals with this problem is to assume the value of `Environment.Newline` for using newlines, which does not work when the user selects to use `\n` everywhere in their editorconfig but writes code on Windows, for example.

## Pros and Cons 

The advantages of making this adjustment to Fantomas are more unification with other tools that embed line-ending information in editorconfig files.

The disadvantages of making this adjustment to Fantomas are slightly increased code complexity, because the chosen newline configuration would need to be threaded through the codebase wherever it was needed.

## Examples

Please provide multiple examples (before and after scenarios) and explain what rules did apply to achieve the outcome.

## Extra information

Estimated cost (XS, S, M, L, XL, XXL): M

Related suggestions: (put links to related suggestions here)

## Affidavit (please submit!)

Please tick this by placing a cross in the box:
* [x] I have read the [Contribution Guidelines](../../CONTRIBUTING.md).
* [x] I have [searched both open and closed suggestions on this site](https://github.com/fsprojects/fantomas/issues) and believe this is not a duplicate

Please tick all that apply:
* [x] This is not a breaking change to Fantomas
* [x] I or my company would be willing to help implement and/or test this
* [ ] This suggestion is part of the [Microsoft style guide](https://docs.microsoft.com/en-us/dotnet/fsharp/style-guide/formatting) (please add a link to section if so)
* [ ] This suggestion is part of the [G-Research style guide](https://github.com/G-Research/fsharp-formatting-conventions) (please add a link to section if so)

nojaf:
Approved.

kodfodrasz:
Hey!

I was having the need for the same thing a few days ago. I'm willing to help. (But would need some guidance in getting started)

nojaf:
Hey @kodfodrasz, @baronfel and I will pick this up in the next `Fantomas & Friends` session this Wednesday. See #1206

Issue #1196, opened 2020-10-18
--check should ignore the line endings


nojaf:
In https://github.com/G-Research/TypeEquality/pull/25 the CI fails because of the difference in line endings.
There was a fix proposed in the PR yet I feel Fantomas should ignore line endings anyways when comparing with the `--check` flag.

PR #1242, opened 2020-11-11
Support user-provided end-of-line characters

baronfel:
This is a WIP of #1231, I will fill in details shortly.


nojaf:
@baronfel I finished the implementation.
One thing though:
`--check` flag should not consider the line endings as a change (see #1196)
But formatting with a file with `\r\n` using setting `lf` should write the changed file to disk, so that is why I had some changed to the FAKEHelpers.

nojaf (approved):

PR #1245, opened 2020-11-12
WIP fix 964

btzo:
I created two tests to check the case separately with _new_ and _inherit_, plus a test with the original issue code.
I would like help to know what to do for each case.

For the _inherit_ test, analyzing the [tree](https://fsprojects.github.io/fantomas-tools/#/ast?data=N4KABGBEAmCmBmBLAdrAzpAXFSAacUiaAYmolmPAIYA2as%2BEkAxgPZwWQAuAngA6ww0VgHMAFAEowAXgA6yCBBQALWACdEXMFWSIAtrUmQQAXyA) I have an implicit inherit, called in ```genMemberDefn``` as ```MDImplicitInherit``` on ```CodePrinter.fs``` .  The solution would be to compare if _ctx.Config.SpaceBeforeClassConstructor_ is set and if it is the case insert a space after ```GenType``` on line 3939 and ignore the  ```addSpaceAfterType``` method.

For the _new_ test, I'm not sure, when I see the [tree](https://fsprojects.github.io/fantomas-tools/#/ast?data=N4KABGBEAmCmBmBLAdrAzpAXFSAacUiaAYmolmPAIYA2as%2BEkAxgPZwWQAuAngA6wwVZIgC2tMAAoasZAHMuAC2wouASjABeADrIIEVAHdpshYo2ahI8TSkz5StZBABfIA) I find a ```SynPat.LongIdent``` containing the new keyword and then a ```SynPat.Paren```, where I believe a space has to be inserted. 
If this is the case I would have to modify the ```PatParen``` of line 4198 to insert a space in and modify the same thing in the functions ```shortExpression``` and ```longExpression``` in the ```PatParen``` of line 4200.

Do these modifications make sense? Maybe there is another case I haven't taken into consideration.

### DIFF
--- a/src/Fantomas.Tests/SpaceBeforeClassConstructorTests.fs
+++ b/src/Fantomas.Tests/SpaceBeforeClassConstructorTests.fs
@@ -102,7 +102,7 @@ type animal(length: int) =
 """
 
 [<Test>]
-let ``SpaceBeforeParenthesisParameterInUppercaseClassConstructor should add space before lowercase constructor of class`` () =
+let ``SpaceBeforeParenthesisParameterInLowercaseClassConstructor should add space before lowercase constructor of class`` () =
     formatSourceString false """
 type animal(length:int) =
     class end
@@ -112,4 +112,56 @@ type animal(length:int) =
 type animal (length: int) =
     class
     end
+"""
+
+// Space before parentheses in secondary class constructor
+
+[<Test>]
+let ``should add space before secondary constructor of class declared with new, 964`` () =
+    formatSourceString false """
+type animal (length: int) =
+    new(length) = animal (length)
+"""  spaceBeforeConfig
+    |> prepend newline
+    |> should equal """
+type animal (length: int) =
+    new (length) = animal (length)
+"""
+
+// Space before parentheses in secondary class constructor
+
+[<Test>]
+let ``should add space after inherit base class declaration, 964`` () =
+    formatSourceString false """
+type dog() =
+    inherit animal()
+"""  spaceBeforeConfig
+    |> prepend newline
+    |> should equal """
+type dog () =
+    inherit animal ()
+"""
+
+// Original issue example
+
+[<Test>]
+let ``should add space before new and inherit on constructor of class, 964`` () =
+    formatSourceString false """
+type ProtocolGlitchException =
+    inherit CommunicationUnsuccessfulException
+
+    new(message) = { inherit CommunicationUnsuccessfulException(message) }
+
+    new(message: string, innerException: Exception) =
+        { inherit CommunicationUnsuccessfulException(message, innerException) }
+"""  spaceBeforeConfig
+    |> prepend newline
+    |> should equal """
+type ProtocolGlitchException =
+    inherit CommunicationUnsuccessfulException
+
+    new (message) = { inherit CommunicationUnsuccessfulException (message) }
+
+    new (message: string, innerException: Exception) =
+        { inherit CommunicationUnsuccessfulException (message, innerException) }
 """
--- a/src/Fantomas.Tests/SpaceBeforeClassConstructorTests.fs
+++ b/src/Fantomas.Tests/SpaceBeforeClassConstructorTests.fs
@@ -140,8 +140,6 @@ type dog () =
     inherit animal ()
 """
 
-// Original issue example
-
 [<Test>]
 let ``should add space before new and inherit on constructor of class, 964`` () =
     formatSourceString false """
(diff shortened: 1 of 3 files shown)


nojaf:
Hey @btzo, for the `inherit` case I would extend `addSpaceAfterType` and pass the context.
Something like:
```fsharp
let addSpaceAfterType ctx =
```
You can then use `ifElseCtx` instead of `ifElse addSpaceAfterType sepSpace sepNone`.

As for the `new` test this is trickier. `genPat` is called all over the place so I wouldn't fix it there just to not break anything.
Once you are inside `genPat` you also sorta lost the information that you came from `ExplicitCtor (ats, px, ao, p, e, so) ` clause.

So I would go with a specific check like:
```fsharp
    | ExplicitCtor (ats, px, ao, p, e, so) ->
        let prefix =
            let genPat ctx =
                match p with
                | SynPat.LongIdent (LongIdentWithDots.LongIdentWithDots([newIdent], _), _, _, SynArgPats.Pats([PatParen _ as pat]), _, _) when (newIdent.idText = "new") ->
                    (!- "new" +> (* TODO add space based on config *) sepSpace +> genPat astContext pat) ctx
                | _ ->
                    // this will probably never match
                    genPat astContext p ctx

            genPreXmlDoc px
            +> genAttributes astContext ats
            +> opt sepSpace ao genAccess
            +> genPat
            +> opt sepNone so (sprintf " as %s" >> (!-))
```

`SynPat.LongIdent (LongIdentWithDots.LongIdentWithDots([newIdent], _), _, _, SynArgPats.Pats([PatParen _ as pat]), _, _) when (newIdent.idText = "new")` will capture the `new (pattern)`. You could then later refactor this to an active pattern in `SourceParser.fs`.
Somelike like `let (|ExplicitCtorPat||_|) = ...`

btzo:
@nojaf , I made the modifications you indicated and I believe that everything is ok, but other modifications are necessary. 

The original test of the issue fails because at a certain moment it falls in the case of an explicit inheritance ```MDInherit``` (Codeprinter.fs  row 4177), only it doesn't do the same treatment that we do in implicit to insert a space.

I don't know if I should treat this case directly in ```MDInherit``` adding a ```sepSpace``` after the call of ```genType``` when the variable in the context is set or if I treat it inside ```genType``` when it reaches ```TLongIdent s```. I believe that the context should be checked inside ```genType```. What do you suggest?

nojaf:
Hey @btzo , there is a bit more ground to cover here when it comes to
> The original test of the issue fails

I didn't see it at first but to solve the original test, we need a change in the inheritExpr of a [SynExpr.Record](http://fsharp.github.io/FSharp.Compiler.Service/reference/fsharp-compiler-syntaxtree-synexpr.html).

I pushed an example but we will need to cover all scenarios:

- short record
- multiline record
- multiline record with setting `MultilineBlockBracketsOnSameColumn` true

btzo:
Ok, I'll look at each one of these cases

btzo:
I added the same function you suggested in the case of multiline records with setting ```MultilineBlockBracketsOnSameColumn``` true and created another test to verify this case.

The original test for the issue works fine now, but the new test has a problem. Seeing the [tree](https://fsprojects.github.io/fantomas-tools/#/ast?data=N4KABGBEAmCmBmBLAdrAzpAXFSAacUiaAYmolmPAIYA2as%2BEkAxgPZwWQAuAngA6wwAISr0AwjVFowAXgA6yCBAButMGi4AnFAHMAjNg3bkOhUrCoA7gAojASllhg6rbr2OjYAL5mlV6w4yTi7G%2Bo5ykBHeCgq8AmAAIrDayrDQElKyvhAoABbJiFzCorAZaGgxikqqNCG6AEyGribZFrA2Rnq4IfWBwXkFRSLikuW2Wnp2reYzszNGDR5a9dFVEP5GvY7OA9pDJWVoAQDcdSYrQZvekCBeQA), it is creating a ```Const Unit``` instead of a ```Paren``` in the last ```new``` constructor. Would this be a bug or another case to be treated? 

Notice that the first constructor with two parameters in the same test does not have this problem.

btzo:
@nojaf

nojaf:
Hey Mike, apologies for the delay, I'll get back to you by the end of the week.

nojaf:
@btzo both `Const(Unit)` and `Paren` should be tackled indeed.
I've made some changes to cope with this and added some extra tests.
If there are not remarks from your side to these changes I think we can merge this.

btzo:
@nojaf  I've no remarks, feel free to do the merge.

nojaf (approved):

nojaf:
Thanks again for this PR @btzo!

PR #1285, opened 2020-12-14
Print trivia after equal sign in SynBinding

nojaf:
Fixes #1248


knocte:
@nojaf wrong issue number?

Issue #1219, opened 2020-11-06
Swallows comment inside `with` block (of a try-with)


knocte:
[...]

PR #1357, opened 2021-01-10
Print trivia of SynMatchClause

HumbertoCortes:
The comments inside `with` block (of a try-with) are formatted. In the Clauses generation method ("genClause") a genTrivia call ("genTriviaMainNodesBeforeClausePipe") is inserted in the CodePrinter. The new method look for a trivia nodes with in TriviaMainNodes list.

Two unit test are added to test the correct behaviour.


knocte:
@nojaf CI on 1st commit was broken because Humberto forgot to run fantomas on fantomas itself, but 2nd commit should have passed, however, the failure seems unrelated to this PR (something about paket...?).

src/Fantomas.Tests/TriviaTests.fs
```diff
@@ -489,3 +490,65 @@ let ``number expression`` () =
     | [ { ContentItself = Some (Number (n))
           Type = TriviaNodeType.MainNode (SynExpr_Const) } ] -> n == "2.0m"
     | _ -> fail ()
+
+[<Test>]
+let ``line comment inside short `with` block (of a try-with), 1219`` () =
```
nojaf, line 495: Please move these tests to ControleStructureTests.fs

src/Fantomas/CodePrinter.fs
```diff
@@ -4103,6 +4103,7 @@ and genClause astContext hasBar (Clause (p, e, eo)) =
         +> leaveNodeTokenByName arrowRange RARROW
 
     genTriviaBeforeClausePipe p.Range
+    +> genTriviaMainNodesBeforeClausePipe p.Range
```
nojaf, line 4106: You introduce a very specific helper here. I'd prefer that an existing helper is used to print the trivia.
`|> genTriviaFor SynMatchClause_Clause p.Range`.

However, this does not solve your indentation problem.
I would add an extra check if you have TriviaContentBefore the `SynMatchClause`.

https://github.com/fsprojects/fantomas/blob/ca42f0d7cf05144c5838bba725378e2a856211bb/src/Fantomas/CodePrinter.fs#L2040

Something like

```fsharp
            let hasCommentBeforeClause = true // TODO: add actual check in TriviaHelpers
            let genClause (astContext: ASTContext) (b: bool) (c: SynMatchClause) =
                ifElse
                    hasCommentBeforeClause
                    (indent +> genClause astContext b c +> unindent)
                    (genClause astContext b c)
```

And if that is the case wrap the genClause between indents.

src/Fantomas/Context.fs
```diff
@@ -1448,6 +1448,15 @@ let internal genTriviaBeforeClausePipe (rangeOfClause: range) ctx =
         | None -> id
     <| ctx
 
+let internal genTriviaMainNodesBeforeClausePipe (rangeOfClause: range) ctx =
+    (Map.tryFindOrEmptyList SynMatchClause_Clause ctx.TriviaMainNodes)
+    |> List.tryFind (fun t -> RangeHelpers.rangeEq t.Range rangeOfClause)
+    |> fun trivia ->
+        match trivia with
+        | Some trivia -> indent +> (printContentBefore trivia)
```
nojaf, line 1456: `indent` without matching `unindent` is a very dangerous thing.
You need to add another test to verify that the indentation is correct again.
Something like:

```fsharp
[<Test>]
let ``foo `` () =
    formatSourceString false """
let outer =
    try
        //humberto
        TrySomething(someParam)
    with
        //comentario
        ex -> MakeSureToCleanup(someParam)

    let someBinding v = v + 2
"""  config
    |> prepend newline
    |> should equal """
let outer =
    try
        //humberto
        TrySomething(someParam)
    with
        //comentario
        ex -> MakeSureToCleanup(someParam)

    let someBinding v = v + 2
"""
```

But as mention above, I'd prefer to not have this helper function.

nojaf (changes requested):
Hello @HumbertoCortes, thank you for this contribution.
Unfortunately, I'd like to see a different implementation.
Please let me know if you need further assistance.

src/Fantomas/Context.fs
HumbertoCortes: @nojaf Thanks so much for your comments. I will work in that direction!

nojaf (approved):
Hey @HumbertoCortes, looks good.
Thanks for the requested changes.
I've tweaked some stuff here and there to work with the full range of SynMatchClause.
Related https://github.com/dotnet/fsharp/issues/10877

nojaf:
Fixes #1219

Issue #1238, opened 2020-11-10
MultilineBlockBracketsOnSameColumn should be honored inside match block


knocte:
[...]

PR #1373, opened 2021-01-19
Respect MultilineBlockBracketsOnSameColumn for SynPat.Record

HumbertoCortes:
@nojaf Working at bug #1238 we are proposing this solution. However the solution does not generate the expected result in the new test. The difference is where the symbol "->" is been generated. Furthermore, another previous test called ``SynPat.Record in pattern match with bracketOnSeparateLine`` does not work, however, in this case, I think the expected result should be modified because of the configuration param "MultilineBlockBracketsOnSameColumn" is set to true.

Thanks in advance for any comment!!


nojaf:
Hello Humberto, thank you for contributing to this project.
The outcome of the test `SynPat.Record in pattern match with bracketOnSeparateLine` can indeed be changed.
That test was mainly there to illustrate that Fantomas did not respect the setting in that scenario. It was not a priority for GR.
I'm happy you found a solution.

Please always give your PR a meaningful title. `Bug fix` + issue number does not really say anything.
I'd prefer a name that tells me what change you would introduce to Fantomas.

src/Fantomas.Tests/ControlStructureTests.fs
```diff
@@ -836,3 +836,35 @@ with ex ->
     Infrastructure.ReportWarning ex
     return None
 """
+
+[<Test>]
```
nojaf, line 840: Please move this test to `MultilineBlockBracketsOnSameColumnRecordTests`.

src/Fantomas.Tests/ControlStructureTests.fs
```diff
@@ -836,3 +836,35 @@ with ex ->
...
+        """
+module Foo =
+    let Bar () =
+        if x then
+            match foo with
+            | {
+                  Bar = true
+                  Baz = _
+              }
+                -> failwith "xxx"
```
nojaf, line 868: Multiline match patterns do not put the arrow on the next line as far as I know.
For example:
```fsharp
match t.Type with
| TriviaNodeType.Token (_, tok) when
    (range.StartLine = t.Range.StartLine
     && range.StartColumn < t.Range.StartColumn) -> Some(tok, t)
| _ -> None
```
so I would not make an exception for this unless there is clear guidance from a style guide.

nojaf (changes requested):
Good start Humberto!

src/Fantomas.Tests/ControlStructureTests.fs
HumbertoCortes: I have changed the expect result of the test according multiline match patterns.

HumbertoCortes:
@nojaf after merge branch "master" into "BugFix_1238" branch from the portal Web, the build process failed.

nojaf:
Hey @HumbertoCortes , could you also please add a test for usages inside a let binding.
Example:
```fsharp
let someFunction { Firstname = fn; Lastname = ln; Age = age } =
    ()
```

I expect this to work without any changes to CodePrinter.
Thanks!

nojaf:
Oh and sorry, I couldn't help myself for the title part 🙈.

HumbertoCortes:
@nojaf I have created the test called "MultilineBlockBracketsOnSameColumn should be honored inside let binding, 1238" in order to test MultilineBlockBracketsOnSameColumn=true inside a let binding. It is passed. But the ";" in the type declaration has been deleted and a '\nl' is added after each printfn statment. is it right?

nojaf (approved):

nojaf:
Hey Humberto, the test you added behaved correctly as it was respecting the MaxLineLenght of 30.
However, this is a bit of a short length and is not recommended in practice.
For me, this is just another test and is not really related to the `30` that was used in the original issue report.
For the same reason, I removed the issue number. But that is really a nitpick as well.

All that aside, many thanks again for this PR!

Issue #1207, opened 2020-10-28
Format in pre-commit hook


nojaf:
As mentioned in https://twitter.com/rastreus/status/1321296796124143617?s=20, how can Fantomas be used as a precommit hook?

On top of my head I think the [local dotnet tool](https://www.nuget.org/packages/fantomas-tool/4.3.0-alpha-002) can be used for this. However, as I find this a good question we should write something out in the documentation to explain this.

Also mention in the docs that `.fantomasignore` file exist in case Fantomas is not able to format the file for some reason.

nojaf:
@rastreus this seems to work for me (on Windows):
```sh
#!/bin/sh
FILES=$(git diff --cached --name-only --diff-filter=ACMR "*.fs" "*.fsx" | sed 's| |\\ |g')
[ -z "$FILES" ] && exit 0

# Format all selected files
echo "$FILES" | xargs dotnet fantomas

# Add back the modified/prettified files to staging
echo "$FILES" | xargs git add

exit 0
```

I stole some code from other places so I can't really explain each line 😅, but the `dotnet fantomas` part is doing the formatting.

agentlewis:
@nojaf i wrote something very similar this once take the hit diff and pass only those files to the shopify cli for deployment.

Basically:

Line 1 gets the changed files filtering the git diff
Line 2 passes the files as args to Fantomas executed with dotnet
Line 3 stage them with git

I thinK the way to to this in a more automated way will be to create a [git hook](https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks) that runs the lines 1 and 2 of your script as a ‘pre-commit hook’.

That should do it.

nojaf:
Hey @agentlewis, this is the content of the `pre-commit` file, so it is already a hook ;)

PR #1376, opened 2021-01-21
add missing raise for formatting functions 

btzo:
fix #1340 

Running `Fantomas.CoreGlobalTool` to format a file without the `--check` argument, it does the treatment in the else listed below, calling the appropriate function given type of `inputPath` and `outputPath`.

https://github.com/fsprojects/fantomas/blob/d22dc764910040fd70412c3f3c674c6f4a5a1250/src/Fantomas.CoreGlobalTool/Program.fs#L386-L406

The problem is that when an error occurred in one of these functions, it was treated locally but not propagated, 
so the `try/with` condition that changes the error code to 1 is never achieved.

The ideal solution would be to show only the error messages in the last `try\with` and add tests for each function in the `match` expression, besides the simple test that I inserted.

What do you think @nojaf ?

PR #1392, opened 2021-01-25
Documentation.md: add pre-commit hook sample

knocte:
Following https://stackoverflow.com/a/42345379/1623521

Fixes https://github.com/fsprojects/fantomas/issues/1207

### DIFF
--- a/docs/Documentation.md
+++ b/docs/Documentation.md
@@ -1068,6 +1068,16 @@ Exclusion applies both to formatting and the format checking.
 
 See [CodeFormatter.fsi](../src/Fantomas/CodeFormatter.fsi) to view the public API of Fantomas.
 
+## A git pre-commit hook sample
+
+A very elegant and transparent way to use fantomas is including it in a pre-commit git hook, by creating a `.git/hooks/pre-commit` file with:
+
+```
+#!/bin/sh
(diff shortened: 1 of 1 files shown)


knocte:
@nojaf FYI I tested this on macOS (which should mean it also works on Linux) and Afshin tested on Windows.

nojaf:
Hello Andres, thanks for writing this down.
I've added a few extra lines to manages people's expectations.
This can work wonderfully or horribly depending on what you do in your code.
Was I right to assume that you've installed the tool globally?

knocte:
> This can work wonderfully or horribly depending on what you do in your code.

What do you mean?

> Was I right to assume that you've installed the tool globally?

Yes indeed, but somehow I had to add the absolute path to fantomas, otherwise the hook would fail :(

nojaf:
> What do you mean?

Formatting before committing. Fantomas could mess up your code right before you commit.
So using this hook could be wonderful or horrible.
We are all working hard to improve this but I feel like we should still add a fair warning.

knocte:
> but I feel like we should still add a fair warning.

Ah yes sure. Anyway if fantomas fails to format, thanks to #1376 it would abort the hook AFAIU.

nojaf:
Well, that is the tricky part. Fantomas does not know what your code is doing. So formatting may have worked but it could change the semantics of the code. In any case, I think we can always add more warnings later if necessary.

nojaf (approved):

Issue #1223, opened 2020-11-06
Should not move the starting of a multi-line comment (stylistic)


knocte:
[...]

PR #1393, opened 2021-01-25
Bug 1223 Starting of multi-line comments should be not moved when are alone "(*"

HumbertoCortes:
@nojaf I'm trying to solve the bug 1223. As far I can see, when multiline comment starts with the token"(*" alone in the line, this token should be moved to the first column of the line. If the comment starts with the token "(*" plus any string, this should respect the Fantomas indentation rules. I have created three test that address this issue. Any comment, please let me know.

### DIFF
--- a/src/Fantomas.Tests/CommentTests.fs
+++ b/src/Fantomas.Tests/CommentTests.fs
@@ -1119,7 +1119,7 @@ let ``starting of a multi-line comment should not be removed (with Begin of Comm
         """
 module Foo =
     let v1 = method7
-            (* Begin of Comment 
+            (* Begin of Comment
        Body of Comment
              *)
     printf "Hello World!"
--- a/src/Fantomas.Tests/CommentTests.fs
+++ b/src/Fantomas.Tests/CommentTests.fs
@@ -1111,3 +1111,88 @@ namespace Foo
 module Bar =
     let Baz () = ()
 """
+
+[<Test>]
+let ``starting of a multi-line comment should not be removed (with Begin of Comment), 1223`` () =
+    formatSourceString
+        false
+        """
+module Foo =
+    let v1 = method7
+            (* Begin of Comment 
+       Body of Comment
+             *)
+    printf "Hello World!"
+"""
+        config
+    |> prepend newline
+    |> should
+        equal
+        """
+module Foo =
+    let v1 = method7
+    (* Begin of Comment
+       Body of Comment
+             *)
+    printf "Hello World!"
+"""
+
+[<Test>]
+let ``starting of a multi-line comment should not be removed (withOut Begin of Comment), 1223`` () =
+    formatSourceString
+        false
+        """
+module Foo =
+    let v1 = method7
+            (*
+       Body of Comment
+*)
+    printf "Hello World!"
+"""
+        config
+    |> prepend newline
+    |> should
+        equal
+        """
+module Foo =
+    let v1 = method7
+(*
+       Body of Comment
+*)
+    printf "Hello World!"
+"""
+
+[<Test>]
+let ``starting of a multi-line comment should not be removed`` () =
+    formatSourceString
+        false
+        """
+module Foo =
+    let private GetConfirmedEtherBalanceInternal (web3: Web3) (publicAddress: string): Async<HexBigInteger> =
+        async {
+             let! blockForConfirmationReference = GetBlockToCheckForConfirmedBalance web3
+(*
+             if (Config.DebugLog) then
+                 Infrastructure.LogError (SPrintF2 "Last block number and last confirmed block number: %s: %s"
+                                                  (latestBlock.Value.ToString()) (blockForConfirmationReference.BlockNumber.Value.ToString()))
+ *)
+            return blockForConfirmationReference
+        }
+"""
+        config
+    |> prepend newline
+    |> should
+        equal
+        """
+module Foo =
+    let private GetConfirmedEtherBalanceInternal (web3: Web3) (publicAddress: string): Async<HexBigInteger> =
+        async {
+            let! blockForConfirmationReference = GetBlockToCheckForConfirmedBalance web3
+(*
+             if (Config.DebugLog) then
+                 Infrastructure.LogError (SPrintF2 "Last block number and last confirmed block number: %s: %s"
+                                                  (latestBlock.Value.ToString()) (blockForConfirmationReference.BlockNumber.Value.ToString()))
+ *)
+            return blockForConfirmationReference
+        }
+"""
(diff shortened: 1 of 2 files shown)


knocte:
Hey Humberto I think you have a typo in the commit message. You wrote "starting of a multi-line comment should not be removed", but I guess you meant to write "move" instead of "remove".

By the way, mentioning the issue number you're trying to fix in the commit message is a good practice, but it's better to do it differently.

Instead of the commit message being a single-line like this:
```
#1223 starting of a multi-line comment should not be removed
```

It can be multi-line (with a title, and a body, like an e-mail), like this:
```
Do not move starting of a multi-line comment

Fixes #1223
```

src/Fantomas/Context.fs
```diff
@@ -84,6 +84,10 @@ module WriterModel =
                 { m with
                       Lines = "" :: m.Lines
                       Column = 0 }
+            | Write s when s.Equals("(*") ->
```
nojaf, line 87: I'm pretty sure this will match more than just code comments.
```fsharp
let v = //
    (*) 4 8
```

This does not look safe at all. @jindraivanek any thoughts?

src/Fantomas.Tests/CommentTests.fs
```diff
@@ -1111,3 +1111,88 @@ namespace Foo
...
+        """
+module Foo =
+    let private GetConfirmedEtherBalanceInternal (web3: Web3) (publicAddress: string): Async<HexBigInteger> =
+        async {
+            let! blockForConfirmationReference = GetBlockToCheckForConfirmedBalance web3
+(*
+             if (Config.DebugLog) then
+                 Infrastructure.LogError (SPrintF2 "Last block number and last confirmed block number: %s: %s"
+                                                  (latestBlock.Value.ToString()) (blockForConfirmationReference.BlockNumber.Value.ToString()))
+ *)
```
nojaf, line 1195: The trivia is assigned to [SynExpr_YieldOrReturn](https://fsprojects.github.io/fantomas-tools/#/trivia?data=N4KABGBEDGD2AmBTSAuKBbBBXANosAYrLGALwA6ATuQHYQR4AuYADpQJYBuAho-gOKJGAYVg0AZu0rpE8AKKMAFokoAhbjm41oiAJI0%2BlGhrAAKAO6IARgGY0AdWs2AlGZZYrOdtACC8eJSIAM5BaEGMHDQA5s5oPkEAntoAPAASiAAequxR%2BnxRKgB8ZFS09OVg3InaYMCldBWNYEwAhGCesNAA1kSUohJS6LzsYgBKiOIqiNr4pGCCjKo4nV0AKrDCyt29-ZLSsuqaM2CWtvWmAFT1TY3s4ma7OQB0ACLWWFEAMrAxYErT1xuN304koVQiWGgjCwgSe3yickolFglDMAGUAAqRRgEABMYHIkE%2B4Pay26YBoWHQVhUlRo8GaJLgA32DI65Mp1JUaAApKEwHzCYCgSLRWLGqZNHxwksVk8AGoaLCIJ7rNERdjRUzOVymdk9FGPaTDMYTKYzJ6y7oAOSpNMoCqVKrVGq1Ouc1wuHuoDRFgWhRlJKx2Yj2Q0YIxo40mgRmwoqAF9IAAacBQJCSGjBVBQFNpyDsIIEILsHPiDRBRAgBNAA) node.

It does not make any sense that it starts at the beginning of the file.
I miss a real reason why it the result did what it did.

If you had:
```fsharp
module Foo =
    let private GetConfirmedEtherBalanceInternal (web3: Web3) (publicAddress: string): Async<HexBigInteger> =
        async {
             let! blockForConfirmationReference = GetBlockToCheckForConfirmedBalance web3
    (*
             if (Config.DebugLog) then
                 Infrastructure.LogError (SPrintF2 "Last block number and last confirmed block number: %s: %s"
                                                  (latestBlock.Value.ToString()) (blockForConfirmationReference.BlockNumber.Value.ToString()))
     *)
            return blockForConfirmationReference
        }
```
as source I believe the result would be the same for no real reason.
This is wide open to interpretation that it should start at the beginning of the file.

nojaf:
Hey @HumbertoCortes, thanks for taking the time to work on this.
I'm not so confident that this is good approach to be honest.
It feels like we miss information of the block comment to make these sort of decisions.
@jindraivanek could you provide your input as well please? I'd like to hear your opinion as well on this one.

src/Fantomas.Tests/CommentTests.fs
knocte: > This is wide open to interpretation that it should start at the beginning of the file.

Hey @nojaf, 1st I guess you meant "line", not "file". Second, maybe Humberto missed the point of #1223 but my expected behaviour for fantomas is not to make it start multi-line comments at the beginning of the line, but to leave comments where they were originally in the source, and not move them.

nojaf: Hey, yes I meant line indeed.

src/Fantomas/Context.fs
jindraivanek: Agree.
Proper place to change BlockComment behavior is at https://github.com/fsprojects/fantomas/blob/795854da9df94c5c4e20bf1271697b4eb205a089/src/Fantomas/Context.fs#L1108-L1113

I would try to use `atIndentLevel true 0` to output something from the start of the line.

knocte: @jindraivanek but dont you think fantomas should put the comment where it originally was in the source? not just at the start of the line always (see convo in the other hunk below)

HumbertoCortes: @jindraivanek @knocte I was trying to modify the suggested block of code, however any changes only affects the generation of "live code", but it does not affect the comments generation. I can see that the comments are saved in ContentBefore/ContentAfter properties. These properties save the comments as raw string (I mean without "Range" that specifies start line and start column). Therefore, there is no way to control where the comment should be located. Is it rigth?

jindraivanek: @HumbertoCortes Right, to achieve this, I think `BlockComment` will need to be extended with original Column position.

nojaf: Indeed, I was also thinking in that direction.
You will need to detect the indentation in 
https://github.com/fsprojects/fantomas/blob/c1a4fe61fc4089eada7b38ef8f872ab28f8b914c/src/Fantomas/TokenParser.fs#L655-L703

And extend `BlockComment` to carry that the extra information.
Perhaps you can get away with just adding the missing spaces to the text that was found in the token.
Let me know if you get what I mean.

PR #1397, opened 2021-01-27
Rather use reraise() instead of raising same captured exception

knocte:
This bad practice is equivalent to doing `throw ex;` instead of `throw`
in C#: you lose the original error stack info.


nojaf (approved):
Thanks @knocte

PR #1398, opened 2021-01-27
Rename MultiLineLambdaClosingNewline to MultilineLambdaClosingNewline

knocte:
For consistency with the other older setting named
'MultilineBlockBracketsOnSameColumn', it's better to rename
this new one.

### DIFF
--- a/docs/Documentation.md
+++ b/docs/Documentation.md
@@ -93,7 +93,7 @@ fsharp_single_argument_web_mode=false
 fsharp_align_function_signature_to_indentation=false
 fsharp_alternative_long_member_definitions=false
 fsharp_disable_elmish_syntax=false
-fsharp_multi_line_lambda_closing_newline=false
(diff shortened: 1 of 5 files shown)


nojaf:
This was not discussed at all, so please do not open a PR with a breaking change.
I cannot allow this at the current time frame.
The proper time to propose this is when the next major version is in scope.

`MultiLineLambdaClosingNewline` is part of 4.3 so I cannot rename it.

knocte:
> MultiLineLambdaClosingNewline is part of 4.3 so I cannot rename it.

Oh I thought it had been introduced only recently. I'll re-propose after 4.4.0 final is released then.

nojaf:
No, it is canon right now. Sorry if my comment sounded a bit harsh. I expect to write out the plan for the next major in the next weeks. We can discuss it then.

knocte:
:+1:

Issue #1407, opened 2021-01-27
Incorrectly combines tokens when formatting


josh-degraw:
[...]

PR #1456, opened 2021-02-17
Add regression test to account for fix of problem in #1407

josh-degraw:
Add regression test to account for fix of problem in #1407, to verify that it was fixed by #1415

### DIFF
--- a/src/Fantomas.Tests/DotIndexedGetTests.fs
+++ b/src/Fantomas.Tests/DotIndexedGetTests.fs
@@ -6,20 +6,20 @@ open Fantomas.Tests.TestHelper
 
 [<Test>]
 let ``multiline function application inside DotIndexedGet`` () =
-  formatSourceString
-    false
-    """
+    formatSourceString
+        false
+        """
 foo.Bar(
             ziggy,
             jiggy,
             "looooooooooooooooooooooooooooooooonStringValue"
         ).[5]
 """
-    { config with MaxLineLength = 70 }
-  |> prepend newline
-  |> should
-       equal
-       """
+        { config with MaxLineLength = 70 }
+    |> prepend newline
+    |> should
+        equal
+        """
 foo.Bar(
     ziggy,
     jiggy,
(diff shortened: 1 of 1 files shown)


src/Fantomas.Tests/DotIndexedGetTests.fs
```diff
@@ -70,3 +70,27 @@ myList.[7]
...
+let ``should not merge tokens inside parentheses, 1407`` () =
+  formatSourceString
+    false
+    """
+let inline (=??) x = (=!) x
+let mySampleMethod() =
+    let result = Ok {| Results = [] |}
+    (Result.okValue result).Results.[0] |> Result.isOk =?? true
+"""
+    config
```
nojaf, line 84: Please update the config to match the settings from the originally reported issue.

nojaf:
I believe a local `.editorconfig` file is interfering with the formatting on your machine.
That would be why the build is still failing.

src/Fantomas.Tests/DotIndexedGetTests.fs
```diff
@@ -70,3 +70,27 @@ myList.[7]
...
+let ``should not merge tokens inside parentheses, 1407`` () =
+    formatSourceString
+        false
+        """
+let inline (=??) x = (=!) x
+let mySampleMethod() =
+    let result = Ok {| Results = [] |}
+    (Result.okValue result).Results.[0] |> Result.isOk =?? true
+"""
+        config
```
nojaf, line 84: Please update the config, the problem was reported using:
```fsharp
    { config with
                SpaceBeforeLowercaseInvocation = false
                SpaceBeforeColon = true
                MaxIfThenElseShortWidth = 25
                MultilineBlockBracketsOnSameColumn = true
                KeepIfThenInSameLine = true
                AlternativeLongMemberDefinitions = true
                MultiLineLambdaClosingNewline = true }
```

the bug occurred using this combination of settings.

nojaf (approved):
Thanks!

Issue #1422, opened 2021-01-30
Violation of "avoid name-sensitive alignments" clause


cmeeren:
[...]

PR #1466, opened 2021-02-20
Add regression test for #1422

theothornhill:
While I was hanging around this repo I though "why not pick some more stuff here".

So here is a quick and simple regression test for #1422.

The returned output reflects fantomas-tools as of e1054c1501b4eadb5fefa7b08daa8fb0f2b6153a

### DIFF
--- a/src/Fantomas.Tests/DotGetTests.fs
+++ b/src/Fantomas.Tests/DotGetTests.fs
@@ -890,3 +890,43 @@ let blah =
             { dasdasdsadsadsadsa = ""
               Sdadsadasdasdas = "sdsadsadasdsa" })
 """
+
+[<Test>]
+let ``avoid name-sensitive alignments, 1422`` () =
+    formatSourceString
+        false
+        """
+let retrySql<'a> =
+  Policy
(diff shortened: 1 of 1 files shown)


src/Fantomas.Tests/LetBindingTests.fs
```diff
@@ -1557,3 +1557,43 @@ let myFun (a: decimal) b c : decimal = a + b + c
 let expensiveToCompute : int = 0
 let myFun (a: decimal) b c : decimal = a + b + c
 """
+
+[<Test>]
```
nojaf, line 1561: Could you move this test to https://github.com/fsprojects/fantomas/blob/master/src/Fantomas.Tests/DotGetTests.fs please?

nojaf:
>  "why not pick some more stuff here".

You are much too kind good sir!

src/Fantomas.Tests/LetBindingTests.fs
theothornhill: Absolutely. Done :) rebased

nojaf (approved):
Many thanks!

PR #1497, opened 2021-02-27
Refactor ASTTransformer

nojaf:
Refactored ASTTransformer:
- optimized for tail recursion, heavily inspired by https://www.gresearch.co.uk/article/advanced-recursion-techniques-in-f/
- Threw out ASTNodes that we never use.
- Threw out properties and FsAstNode as they are not used.

Please review if I missed something.

### DIFF
--- a/src/Fantomas/AstTransformer.fs
+++ b/src/Fantomas/AstTransformer.fs
@@ -117,16 +117,15 @@ module private Ast =
                 visit
                     expr
                     (fun nodes ->
-                        [ mkNode SynExpr_Paren range
-                          yield! nodes ]
+                        mkNode SynExpr_Paren range :: nodes
                         |> finalContinuation)
             | SynExpr.Quote (operator, _, quotedSynExpr, _, range) ->
                 let continuations : ((TriviaNodeAssigner list -> TriviaNodeAssigner list) -> TriviaNodeAssigner list) list =
                     [ visit operator; visit quotedSynExpr ]
 
                 let finalContinuation (nodes: TriviaNodeAssigner list list) : TriviaNodeAssigner list =
-                    [ mkNode SynExpr_Quote range
-                      yield! (List.collect id nodes) ]
+                    mkNode SynExpr_Quote range
+                    :: (List.collect id nodes)
                     |> finalContinuation
 
                 Continuation.sequence continuations finalContinuation
@@ -141,8 +140,8 @@ module private Ast =
                     exprs |> List.map visit
 
                 let finalContinuation (nodes: TriviaNodeAssigner list list) : TriviaNodeAssigner list =
-                    [ mkNode SynExpr_Tuple range
-                      yield! (List.collect id nodes) ]
+                    mkNode SynExpr_Tuple range
+                    :: (List.collect id nodes)
                     |> finalContinuation
 
                 Continuation.sequence continuations finalContinuation
@@ -151,8 +150,8 @@ module private Ast =
                     exprs |> List.map visit
 
                 let finalContinuation (nodes: TriviaNodeAssigner list list) : TriviaNodeAssigner list =
-                    [ mkNode SynExpr_ArrayOrList range
-                      yield! (List.collect id nodes) ]
+                    mkNode SynExpr_ArrayOrList range
+                    :: (List.collect id nodes)
                     |> finalContinuation
 
                 Continuation.sequence continuations finalContinuation
@@ -173,12 +172,12 @@ module private Ast =
                           yield! visitSynType typeName ]
                         |> finalContinuation)
             | SynExpr.ObjExpr (objType, argOptions, bindings, extraImpls, _, range) ->
-                mkNode SynExpr_ObjExpr range
-                :: [ yield! visitSynType objType
-                     if argOptions.IsSome then
-                         yield! visitArgsOption argOptions.Value
-                     yield! extraImpls |> List.collect visitSynInterfaceImpl
-                     yield! bindings |> List.collect visitSynBinding ]
+                [ yield mkNode SynExpr_ObjExpr range
+                  yield! visitSynType objType
+                  if argOptions.IsSome then
+                      yield! visitArgsOption argOptions.Value
+                  yield! extraImpls |> List.collect visitSynInterfaceImpl
+                  yield! bindings |> List.collect visitSynBinding ]
                 |> finalContinuation
             | SynExpr.While (_, whileExpr, doExpr, range) ->
                 let continuations : ((TriviaNodeAssigner list -> TriviaNodeAssigner list) -> TriviaNodeAssigner list) list =
@@ -207,7 +206,7 @@ module private Ast =
                     [ visit enumExpr; visit bodyExpr ]
 
                 let finalContinuation (nodes: TriviaNodeAssigner list list) : TriviaNodeAssigner list =
-                    [ mkNode SynExpr_ForEach range
+                    [ yield mkNode SynExpr_ForEach range
                       yield! visitSynPat pat
                       yield! (List.collect id nodes) ]
                     |> finalContinuation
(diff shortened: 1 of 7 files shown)


nojaf:
# PR

## Ubuntu

| Method |    Mean |    Error |   StdDev | Rank |       Gen 0 |      Gen 1 |     Gen 2 | Allocated |
|------- |--------:|---------:|---------:|-----:|------------:|-----------:|----------:|----------:|
| Format | 2.513 s | 0.0383 s | 0.0358 s |    1 | 123000.0000 | 44000.0000 | 5000.0000 |   1.76 GB |


## Windows

| Method |    Mean |    Error |   StdDev | Rank |       Gen 0 |      Gen 1 |     Gen 2 | Allocated |
|------- |--------:|---------:|---------:|-----:|------------:|-----------:|----------:|----------:|
| Format | 3.071 s | 0.0488 s | 0.0433 s |    1 | 102000.0000 | 32000.0000 | 3000.0000 |   1.76 GB |

## Mac

| Method |    Mean |    Error |   StdDev | Rank |       Gen 0 |       Gen 1 |     Gen 2 | Allocated |
|------- |--------:|---------:|---------:|-----:|------------:|------------:|----------:|----------:|
| Format | 2.765 s | 0.0412 s | 0.0385 s |    1 | 300000.0000 | 121000.0000 | 5000.0000 |   1.76 GB |

# Master

## Ubuntu

| Method |    Mean |    Error |   StdDev | Rank |      Gen 0 |      Gen 1 |     Gen 2 | Allocated |
|------- |--------:|---------:|---------:|-----:|-----------:|-----------:|----------:|----------:|
| Format | 3.228 s | 0.0469 s | 0.0416 s |    1 | 77000.0000 | 21000.0000 | 5000.0000 |   1.79 GB |

## Windows

| Method |    Mean |    Error |   StdDev | Rank |      Gen 0 |      Gen 1 |     Gen 2 | Allocated |
|------- |--------:|---------:|---------:|-----:|-----------:|-----------:|----------:|----------:|
| Format | 2.676 s | 0.0526 s | 0.0541 s |    1 | 77000.0000 | 21000.0000 | 5000.0000 |   1.79 GB |

## Mac

| Method |    Mean |    Error |   StdDev | Rank |       Gen 0 |       Gen 1 |     Gen 2 | Allocated |
|------- |--------:|---------:|---------:|-----:|------------:|------------:|----------:|----------:|
| Format | 3.039 s | 0.0599 s | 0.0641 s |    1 | 304000.0000 | 121000.0000 | 4000.0000 |   1.79 GB |

Smaug123:
I'll try and review this tomorrow (Monday) - although I'm only just back from a period of leave, so I might be swamped.

Smaug123:
I've sat down to look at this a few times, and… it's just too big for me to meaningfully review. How confident are you in the change, and/or is there any particular bit you'd like me to look at?

Smaug123:
One thing that would help a lot, I think, is if you broke this out into two different reviews: one for "Threw out ASTNodes that we never use" and "Threw out properties and FsAstNode as they are not used.", and one for the CPS transform. I understand how painful this will be to extract, though :(

nojaf:
Hey Patrick, good feedback. This is indeed a too large PR that does multiple things.
I'd mostly like a review of the CPS stuff. I'll annotate one function and ask some specific question.
The answers will apply to all of them.
I'm actually very confident about the entire PR.

src/Fantomas/AstTransformer.fs
```diff
@@ -62,1634 +34,1211 @@ module private Ast =
...
+            match synExpr with
+            | SynExpr.Paren (expr, _, _, range) ->
+                visit
+                    expr
+                    (fun nodes ->
+                        [ mkNode SynExpr_Paren range
+                          yield! nodes ]
+                        |> finalContinuation)
+            | SynExpr.Quote (operator, _, quotedSynExpr, _, range) ->
+                let continuations : ((TriviaNodeAssigner list -> TriviaNodeAssigner list) -> TriviaNodeAssigner list) list =
```
nojaf, line 124: Does it makes sense to use the sequence helper function when the list is fixed and small (2 elements in this case)?

src/Fantomas/AstTransformer.fs
```diff
@@ -62,1634 +34,1211 @@ module private Ast =
...
+                    [ visit operator; visit quotedSynExpr ]
+
+                let finalContinuation (nodes: TriviaNodeAssigner list list) : TriviaNodeAssigner list =
+                    [ mkNode SynExpr_Quote range
```
nojaf, line 128: I used `List.collect` a lot of times, any problems or improvements there?

src/Fantomas/AstTransformer.fs
```diff
@@ -62,1634 +34,1211 @@ module private Ast =
...
+//                           Range = r range
+//                           Properties = Map.empty
+//                            }
+                        nodes @ visitSynType typeName |> finalContinuation)
+            | SynExpr.Tuple (_, exprs, _, range) ->
+                let continuations : ((TriviaNodeAssigner list -> TriviaNodeAssigner list) -> TriviaNodeAssigner list) list =
+                    exprs |> List.map visit
+
+                let finalContinuation (nodes: TriviaNodeAssigner list list) : TriviaNodeAssigner list =
+                    [ mkNode SynExpr_Tuple range
```
nojaf, line 151: Would `mkNode SynExpr_Tuple range :: (List.collect id nodes)` be better? I seem to switch styles here and there.

src/Fantomas/AstTransformer.fs
```diff
@@ -62,1635 +34,1170 @@ module private Ast =
...
+            | SynModuleDecl.NamespaceFragment (moduleOrNamespace) ->
+                visitSynModuleOrNamespace moduleOrNamespace
+                |> finalContinuation
+
+        visit ast id
+
+    and visitSynExpr (synExpr: SynExpr) : TriviaNodeAssigner list =
+        let rec visit
+            (synExpr: SynExpr)
+            (finalContinuation: TriviaNodeAssigner list -> TriviaNodeAssigner list)
```
Smaug123, line 113: By the way, I find these things easier to write correctly if I stay generic as long as possible. It makes it harder when you can't put generic type parameters on, like in an inner function, but if `finalContinuation` returned a `'ret` then the types constrain you further and it's harder to get it wrong. I'm not sure whether that's possible here.

src/Fantomas/AstTransformer.fs
Smaug123: I can't think of an alternative.

src/Fantomas/AstTransformer.fs
Smaug123: Pff, I really don't mind. Personally I prefer to be explicit if I use the `yield` keyword - I'm not a fan of implicit yields, because if you omit the yields then you have the chance in principle to get confused between `[if true then yield 3 else yield 4]` and `[yield (if true then 3 else 4)]`.

I suspect `mkNode SynExpr_Tuple range :: (List.collect id nodes)` might be more efficient - it would certainly rely less on the compiler's magic ability to optimise. It's not obvious to me that the computation expression avoids copying whatever is `yield!`'ed.

src/Fantomas/AstTransformer.fs
```diff
@@ -62,1635 +34,1170 @@ module private Ast =
...
+                :: (List.collect visitRecordField recordFields)
+                |> finalContinuation
+            | SynExpr.AnonRecd (_, _, recordFields, range) ->
+                mkNode SynExpr_AnonRecd range
+                :: (List.collect visitAnonRecordField recordFields)
+                |> finalContinuation
+            | SynExpr.New (_, typeName, expr, range) ->
+                visit
+                    expr
+                    (fun nodes ->
```
Smaug123, line 170: I've got a post coming out soon where I talk about continuation-passing style all over again, and in the course of that, I ended up much preferring the following layout just on aesthetic grounds:

```fsharp
(fun nodes ->
    blah
)
|> visit expr
```
Saves a layer of indentation, although it is *quite weird* from an imperative-programming point of view.

Smaug123 (approved):
I certainly haven't read the whole thing, but I think it looks basically good. I have no fundamental comments other than "is there anywhere you could be more generic?".

nojaf:
Thanks for the review @Smaug123!

Issue #1361, opened 2021-01-12
<Insert meaningful title>


Smaug123:
[...]

Issue #1501, opened 2021-03-05
Indentation of pattern match clause 


nojaf:
[...]

PR #1502, opened 2021-03-05
Initial support of setting KeepIndentInBranch.

nojaf:
A first attempt to get to the behaviour of https://github.com/fsprojects/fantomas/issues/1361.
I would need some more examples to detect if all use-cases are covered by this.
The solution is based on the AST of the source so that is still in spirit with the project.
Please take a look at the unit tests and let me know if this is a good start.

### DIFF
--- a/src/Fantomas.Tests/KeepIndentInBranch.fs
+++ b/src/Fantomas.Tests/KeepIndentInBranch.fs
@@ -408,7 +408,9 @@ let sum a b =
 
 [<Test>]
 let ``match followed by if should not indent in both cases`` () =
-    formatSourceString false """
+    formatSourceString
+        false
+        """
 let sum a b =
     if a < 0 then
         None
@@ -420,9 +422,12 @@ let sum a b =
             logMessage "a and b are both positive"
             // some grand explainer about the code
             Some (a + b)
-"""  config
+"""
+        config
     |> prepend newline
-    |> should equal """
+    |> should
+        equal
+        """
 let sum a b =
     if a < 0 then
         None
--- a/src/Fantomas.Tests/KeepIndentInBranch.fs
+++ b/src/Fantomas.Tests/KeepIndentInBranch.fs
@@ -1,0 +1,253 @@
+﻿module Fantomas.Tests.KeepIndentInBranch
+
+open NUnit.Framework
+open FsUnit
+open Fantomas.Tests.TestHelper
+open Fantomas.FormatConfig
+
+let config =
+    { config with
+          KeepIndentInBranch = true }
+
+[<Test>]
+let ``single expression in if branch, multiple expressions in else branch`` () =
+    formatSourceString
+        false
+        """
+if someCondition then
+    0
+else
+    let config = Configuration.Read "/myfolder/myfile.xml"
+    let result = Process.main config otherArg
+    if result.IsOk then
+        0
+    else
+        -1
+"""
+        config
+    |> prepend newline
+    |> should
+        equal
+        """
+if someCondition then
+    0
+else
+let config =
+    Configuration.Read "/myfolder/myfile.xml"
+
+let result = Process.main config otherArg
+if result.IsOk then 0 else -1
+"""
+
+[<Test>]
+let ``sequential in if branch, let or use in else branch`` () =
+    formatSourceString
+        false
+        """
+open Library
+
+type RunMode =
+    | Dry
+    | Wet
+
+let main argv =
+    let args = parse argv
+
+    let instructions = Library.foo args
+    if args.DryRun = RunMode.Dry then
+        printfn "Would execute actions, but --dry-run was supplied: %+A" instructions
+        0
+    else
+        // proceed with main method
+        let output = Library.execute instructions
+        // do more stuff
+        0
+"""
+        config
+    |> prepend newline
+    |> should
+        equal
+        """
+open Library
+
+type RunMode =
+    | Dry
+    | Wet
+
+let main argv =
+    let args = parse argv
+
+    let instructions = Library.foo args
+
+    if args.DryRun = RunMode.Dry then
+        printfn "Would execute actions, but --dry-run was supplied: %+A" instructions
+        0
+    else
+    // proceed with main method
+    let output = Library.execute instructions
+    // do more stuff
+    0
+"""
+
+[<Test>]
+let ``sequential in first clause, let or use in last clause`` () =
+    formatSourceString
+        false
+        """
+open Library
+
+type RunMode =
+    | Dry
+    | Wet
+
+let main argv =
+    let args = parse argv
+
+    let instructions = Library.foo args
+    match args.DryRun with
+    | RunMode.Dry ->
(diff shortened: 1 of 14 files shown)


src/Fantomas.Tests/IfThenElseTests.fs
```diff
@@ -1651,14 +1651,14 @@ let tryDecompile (ty: FSharpEntity) =
     match ty.TryFullName with
     | Some fullName -> return decompile ty.Assembly.SimpleName externalSym
     | None ->
-        // might be abbreviated type (like string)
```
Smaug123, line 1654: Just checking that in the final version, you'll also keep the tests for when the feature is turned off? (I'm happy with this test case though.)

src/Fantomas.Tests/ComputationExpressionTests.fs
```diff
@@ -1609,25 +1609,25 @@ let rec loop () =
 
     match msg with
     | Handle (eventSource, command, reply) ->
-        let! stream = eventSource |> eventStore.GetStream
```
Smaug123, line 1612: Looks sane to me

src/Fantomas.Tests/KeepIndentInBranch.fs
```diff
@@ -0,0 +1,253 @@
...
+        equal
+        """
+if someCondition then
+    0
+else
+let config =
+    Configuration.Read "/myfolder/myfile.xml"
+
+let result = Process.main config otherArg
+if result.IsOk then 0 else -1
```
Smaug123, line 39: I think some of the value of this test case has been lost with how short `0` and `-1` are. Maybe it's worth making these expressions long enough that Fantomas splits them onto new lines, so we can see how they behave?

src/Fantomas.Tests/KeepIndentInBranch.fs
```diff
@@ -0,0 +1,253 @@
...
+    let args = parse argv
+
+    let instructions = Library.foo args
+
+    if args.DryRun = RunMode.Dry then
+        printfn "Would execute actions, but --dry-run was supplied: %+A" instructions
+        0
+    else
+    // proceed with main method
+    let output = Library.execute instructions
```
Smaug123, line 87: Happy with this one

src/Fantomas.Tests/KeepIndentInBranch.fs
```diff
@@ -0,0 +1,253 @@
...
+
+    let instructions = Library.foo args
+
+    match args.DryRun with
+    | RunMode.Dry ->
+        printfn "Would execute actions, but --dry-run was supplied: %+A" instructions
+        0
+    | RunMode.Wet ->
+    // proceed with main method
+    let output = Library.execute instructions
```
Smaug123, line 139: Happy with this one

src/Fantomas.Tests/KeepIndentInBranch.fs
```diff
@@ -0,0 +1,253 @@
...
+type RunMode =
+    | Dry
+    | Wet
+
+let main argv =
+    let args = parse argv
+
+    let instructions = Library.foo args
+
+    match! args.DryRun with
```
Smaug123, line 185: Astonished this compiles, but sure

src/Fantomas.Tests/KeepIndentInBranch.fs
```diff
@@ -0,0 +1,253 @@
...
+
+    let instructions = Library.foo args
+
+    let exitCode =
+        function
+        | RunMode.Dry ->
+            printfn "Would execute actions, but --dry-run was supplied: %+A" instructions
+            0
+        | RunMode.Wet ->
+        // proceed with main method
```
Smaug123, line 247: Seems reasonable

src/Fantomas.Tests/PatternMatchingTests.fs
```diff
@@ -587,8 +587,8 @@ match x with
         """
 match x with
 | Some y ->
-    let z = 1
```
Smaug123, line 590: I'm not convinced by this. I think the feature should only affect the last block, and it shouldn't affect the *width* of an indentation at all - only the "number of tabs". This is the first block in the match, and not the last block, so I think it shouldn't change.

Smaug123 (commented):
The implementation is surprisingly simple!

I guess it would be useful to have a few more tests documenting the boundary conditions. For example, the difference (if any) between the following:

```fsharp
match true with
| false ->
    failwith ""
| true ->
    failwith ""

foo ()
```
and 
```fsharp
match true with
| false ->
    failwith ""
| true ->
    failwith ""
    0

foo ()
```

and also what happens when you remove the trailing `foo ()`. The test cases right now are more showing us what the feature does, which is very useful, but it's the edge cases where the subtlety lies. As I understand the feature, for example, my version with the extra line that returns `0` should trigger unindentation, but the version without should not; and the presence of `foo ()` must prevent unindentation. But I recognise that I have really not specified what the feature should do, so if my understanding is different from yours, that's probably not a problem!

src/Fantomas.Tests/ComputationExpressionTests.fs
nojaf: This is actually a bit of a side effect of a change in CodePrinter.
I saw the current behavior is somewhat weird and logged https://github.com/fsprojects/fantomas/issues/1501 as it is not in line with the style guides.
So, the modified tests are more a result of that change.

src/Fantomas.Tests/PatternMatchingTests.fs
nojaf: This test is actually unrelated to the feature but also is a side effect of solving https://github.com/fsprojects/fantomas/issues/1501.

src/Fantomas/CodePrinter.fs
```diff
@@ -4373,12 +4378,32 @@ and genClause astContext hasBar (Clause (p, e, eo) as ce) =
         +> fun ctx -> leaveNodeTokenByName (arrowRange ctx) RARROW ctx
 
     genTriviaBeforeClausePipe p.Range
-    +> ifElse
-        hasBar
-        (sepBar +> atCurrentColumnWithPrepend pat body
```
nojaf, line 4378: I remove `atCurrentColumnWithPrepend` to solve https://github.com/fsprojects/fantomas/issues/1501 and because it simplifies the flow of the code. 
The only reason we had this was because one nitpick of a user in a time where style guides weren't a thing yet.
When I looked it up in the style I noticed our current behavior was inconsistent.

`AtCurrentColumn` will lock the column indent and override the normal flow of indentation, it is a last resort and that is the main reason I wanted to get rid of it.

Smaug123:
One of our examples:

```fsharp
module Foo =
    let make () =
        { new IBar with
            member __.DoIt (i : int) =
                doSomeStuff ()
                match i with
                | 0 ->
                    handleIt ()
                    None
                | _ ->

                let baz = foo ()
                match int baz with
                | 0 ->
                    handleAgain ()
                    None
                | _ ->

                doEvenMore ()
                Some 3
        }
```

PR #1656, opened 2021-04-16
Don't use module or namespace as trivia candidate. 

### LINKED ISSUE #1649: Idempotency problem when ending with a comment
Issue created from [fantomas-online](https://fsprojects.github.io/fantomas-tools/#/fantomas/preview?data=N4KABGBEDGD2AmBTSAuKBiANrAhvMAOpABICWAdAGYDOAHkQQHawAOijYAyrALaIAuAC1KMA5kyYB6aZLAAxAPIKmAbSYQIARnUaATDogBmJgF0JjGbIAqgxGHbxIAGnBRqA-iNHVUYFa4hQDQ1IABJ%2BAE82X0gRfmcAkNCANxxMAFdkNH9g4IAGF1yQgElGJEZ%2BTlIAL2RCorAAFkSIE0SAX3rAlqhwqKyoOISisNSMgZyGzS7cyABZHFoAGRFEJfZRIWGp3TyetuDOxKDZvui0SAAjWFhMbeDRtMzfSaLdGYfORB5SOGxGACC-AAomUFJQVow6j0IJQ0u59h0ZicHmcBlcbncPr0xs9sjCwIZsRBIJwWDhoIgAEKISiwABOiAACjh6Tg%2BPxEPT7rl%2BPTMojDsiemFIucoNdbjySSknhMCY1iVAyRTqbSGWtYAB3LnQHDuUrJWB6zywRjSjR8gVFA4aI7BFFJMXoyVYkWy8YvAkAViVpPJlJpdMZAFUWGx6XqDYwjSbSGaLRArYhBXbhSM0TFXRbHp78Q0wAA2P0qwPqxkAYUw%2BuoFbN1Ct0H4DMTYGTqYg9o0jplzqzmJzHrxfgJAHYSwG1cHEHNvpcua32zakcd3X2Ltnsbnh69cgAOCeqoMauv-Rf8lPLoWrjPriUDrdD%2BUFgCch8pAMonPpdZ4PBw57Wrktqdumpx3hiUqPriz5TAUBL%2Bke5aIF8Px-AmSpLsBK4Omu-T9lB7owV6BaaNMCGlogn7fqhvy3BhBJYcEIFgF23S3vhG4PkRcokTsfqlOU-AKIwVj0hEADqpBbEqcKYAiV5pje4GcfehEZsR%2BZTESFGTgC9KwOkZQACKIJgpA-N%2BgGXth164Rx4qxBUg6aSOpGKghCy0MUlA2OwwLyShggMvwUnwDJCp7IpoHKaiEFDNBvFaUUmi%2Bp5iylJQpC0AoEY4M29LArQLCMtQ1DxuaSrelFtlKfZKmOQlPF5m5UzFultAAEqIHA9LwGFEUFo0NXMTh3Z4Y1zmJS1u7BJo44dd1vXwAAcukPDzvS4LFJyPA%2BEqmgdqxYFxapkA8OkmCeOZUJyAy-78N%2BVj4dNO4EpoB4IUtDLwHMl3Xasd30g9Vl%2BtAgishS34APpaqQ4WCJAR1sWAPa9PFU3NW9pFvh1%2BlshECj0isDYDYjSrDcjJ1OmdTUaUlrVvPBBbzIs%2BM4ITxOkA2a0bVy227ft71U7FNOORdV2kDdiBAyDXLPdEr2wW85Es%2BznMk-wf2S9Lsv5aDCHg5DTZcrD8NbCL9WnZN8RK3xbzvB1ABqcpUiI8BeGTraU9Fx2i72tOY-TM0EroOks15chGU2FVu2Unvm%2BTkWW%2BNDnonTpyubNegeRHizGbA-AAOICEVJWIGVFVe1VI0aCxKNo6KZ2bljyu5LoaUR-9UurFS2DQAA1lSbKDwI1AiZw7KIKe62VYxF4p%2BxDUutxwfYw0ujtSzK2IFq0s0vwOrsAriCmVljDSRVAJlLOfP0kLBZMXXY1L9bK-qZnDPZxAugLSzADSiBEAsB8n5RgpRJ58EhNCR%2BC9fYNwmunIOn8Q4Fl0J9POtAAo-GoIIauyd4HUwDo5Fua827BF0LjFmVQxCYCovSUQ612ChUQJcOYCAYENDkgpWqMUrZi3fm6Mh9tciGGZg0SAAJzKiEYFHRgMczRVBkfldIjIrCwEEsw-KFVrKL1RoggiQiUHryKIYVWEipHfkYNo5ImoxC302mfEQl96y6MIf7dGzdV7GPIRoQwjsu6S2gUsdklx4A4CrLAcqYgd571WG43hft%2BHEMES5L%2BBJDDhwkcZbmOBLh0OwdzQQnAIgVEWK2bhNlRp2VTsvGIDhwTQM4JEOhaTUENEMLnCRoJ4CNPiWDekmBKBI3cckzxJDvGoizhkzuEjAHAM0RUUow8cDyKTrAoC1S6q1LfoYtpJjRFbwkX3VZA9oHUH0oZMoO8GyIF%2Bt3aWZdSrlVcZhOBiSEFpz2XbZKoi-4SOafSX4WsOEVPhFU5%2BhxXAmHqLEagchyq%2BEqSAdoQA)

#### Formatted code

```fsharp
#load "Hi.fsx"
open Something

//// FOO
[ 1 ; 2 ; 3 ]
//// The end


```
            

#### Reformatted code

```fsharp
#load "Hi.fsx"
open Something

//// FOO
[ 1 ; 2 ; 3 ]
//// The end

```
            
#### Problem description

This was an FSX file. Fantomas couldn't decide how many newlines to put on the end.

#### Extra information

- [ ] The formatted result breaks by code.

#### Options

Fantomas Master at 04/14/2021 13:41:07 - e164e231f5456b9e128bc567462fcb39cbb0f491

```fsharp
    { config with
                SpaceBeforeUppercaseInvocation = true
                SpaceBeforeClassConstructor = true
                SpaceBeforeMember = true
                SpaceBeforeColon = true
                SpaceBeforeSemicolon = true
                MultilineBlockBracketsOnSameColumn = true
                NewlineBetweenTypeDefinitionAndMembers = true
                KeepIfThenInSameLine = true
                AlternativeLongMemberDefinitions = true
                MultiLineLambdaClosingNewline = true
                KeepIndentInBranch = true }
```

<sub>Did you know that you can ignore files when formatting from fantomas-tool or the FAKE targets by using a [.fantomasignore file](https://github.com/fsprojects/fantomas/blob/master/docs/Documentation.md#ignore-files-fantomasignore)?</sub>

nojaf:
Fixes #1649.
This PR has a bit of an impact, I would like to hear your opinion @jindraivanek.

So, in general, modules and namespace can have a bit of a weird range.
For example an empty named module in [a signature file](https://fsprojects.github.io/fantomas-tools/#/ast?data=N4KABGBEAmCmBmBLAdrAzpAXFSAacUiaAYmolmAC4BOArrPhJAMYD2cFkAtu7QDawwxVqwBGAQ2oAdachlTkAekVhWACzBpWXQWy47klSCAC%2BQA).
The range is influenced by the code comment, so that leads to problems here and there.
In short, it is better not to use them as trivia candidates.

To overcome this limitation, I constructed a range based on the `LongIndent` of a `SynModuleOrNamespace` or `SynModuleOrNamespaceSig`.
Trivia can safely be attached to this node, and then printed for the entire module name (keywords included).
See
```fsharp
    let moduleOrNamespace =
        genModuleOrNamespaceKind moduleKind
        +> opt sepSpace ao genAccess
        +> ifElse isRecursive (!- "rec ") sepNone
        +> col (!- ".") lids (fun (lid, r) -> genTriviaFor Ident_ r (!-lid))
        |> genTriviaFor LongIdent_ lidsFullRange
```

This all works quite well yet I encountered one problem with TriviaContent after in:
```fsharp
#if INTERACTIVE
#load "../FSharpx.TypeProviders/SetupTesting.fsx"
SetupTesting.generateSetupScript __SOURCE_DIRECTORY__
#load "__setup__.fsx"
#endif
```

`#endif` is now attached as content after `SynModuleDecl_HashDirective`, which in itself is quite sensible.
It used to be attached to the content after the anonymous module.
The three top-level `decls` are combined using the `colWithNlnWhenItemIsMultiline` helper function.
There we check if the constructs are multiline and we add new lines between them if so.

The problem I encountered is that `SynModuleDecl.HashDirective` now is multiline because of the trivia after it.
So this introduced an extra newline, where there was non.
I'm not sure how to deal with this in any other way.
A `WriteLine` event was detected so we concluded it is multiline.

The good news of these changes is that trivia got a bit simplified. Fewer hacks are necessary to deal with empty modules and namespaces. Newlines and comments on a single line follow the exact same logic in candidate election so that part is definitely a pro.

Does it outweigh the con listed above?

### DIFF
--- a/src/Fantomas.Tests/CommentTests.fs
+++ b/src/Fantomas.Tests/CommentTests.fs
@@ -1297,26 +1297,6 @@ open Something
 """
 
 [<Test>]
-let ``comment after let binding makes it multiline`` () =
-    formatSourceString
-        false
-        """
-let a = 7
-let b = 8
-// foo
-"""
-        config
-    |> prepend newline
-    |> should
-        equal
-        """
-let a = 7
-
-let b = 8
-// foo
-"""
-
-[<Test>]
 let ``block comment above let binding`` () =
     formatSourceString
         false
--- a/src/Fantomas.Tests/CommentTests.fs
+++ b/src/Fantomas.Tests/CommentTests.fs
@@ -1297,7 +1297,7 @@ open Something
 """
 
 [<Test>]
-let ``meh 123`` () =
+let ``comment after let binding makes it multiline`` () =
     formatSourceString
         false
         """
--- a/src/Fantomas.Tests/CommentTests.fs
+++ b/src/Fantomas.Tests/CommentTests.fs
@@ -1252,3 +1252,82 @@ let a = 8
 let a = 8
 // foobar
 """
+
+[<Test>]
+let ``file end with newline followed by comment, 1649`` () =
+    formatSourceString
+        false
+        """
+#load "Hi.fsx"
+open Something
+
+//// FOO
+[
+    1
+    2
+    3
+]
+
+//// The end
+"""
+        { config with
+              SpaceBeforeUppercaseInvocation = true
+              SpaceBeforeClassConstructor = true
+              SpaceBeforeMember = true
+              SpaceBeforeColon = true
+              SpaceBeforeSemicolon = true
+              MultilineBlockBracketsOnSameColumn = true
+              NewlineBetweenTypeDefinitionAndMembers = true
+              KeepIfThenInSameLine = true
+              AlignFunctionSignatureToIndentation = true
+              AlternativeLongMemberDefinitions = true
+              MultiLineLambdaClosingNewline = true
+              KeepIndentInBranch = true }
+    |> prepend newline
+    |> should
+        equal
+        """
+#load "Hi.fsx"
+open Something
+
+//// FOO
+[ 1 ; 2 ; 3 ]
+
+//// The end
+"""
+
+[<Test>]
+let ``meh 123`` () =
+    formatSourceString
+        false
+        """
+let a = 7
+let b = 8
+// foo
+"""
+        config
+    |> prepend newline
+    |> should
+        equal
+        """
+let a = 7
+
+let b = 8
+// foo
+"""
+
+[<Test>]
+let ``block comment above let binding`` () =
+    formatSourceString
+        false
+        """(* meh *)
+let a =  b
+"""
+        config
+    |> prepend newline
+    |> should
+        equal
+        """
+(* meh *)
+let a = b
+"""
--- a/src/Fantomas/AstTransformer.fs
+++ b/src/Fantomas/AstTransformer.fs
@@ -9,13 +9,6 @@ open Fantomas
 type Id = { Ident: string; Range: Range }
 
 module Helpers =
-    let i (id: Ident) : Id =
-        { Ident = id.idText
-          Range = id.idRange }
-
-    let li (id: LongIdent) = id |> List.map i
-
-    let lid (id: LongIdentWithDots) = li id.Lid
     let mkNode (t: FsAstType) (r: range) = TriviaNodeAssigner(MainNode(t), r)
 
 module private Ast =
@@ -23,27 +16,13 @@ module private Ast =
 
     let rec visitSynModuleOrNamespace (modOrNs: SynModuleOrNamespace) : TriviaNodeAssigner list =
         match modOrNs with
-        | SynModuleOrNamespace (longIdent, _, synModuleOrNamespaceKind, decls, _, attrs, _, range) ->
-            let collectIdents (idents: LongIdent) =
-                idents
-                |> List.map (fun ident -> mkNode Ident_ ident.idRange)
+        | SynModuleOrNamespace (longIdent, _, kind, decls, _, attrs, _, range) ->
+            let longIdentNodes =
+                match kind, decls with
+                | SynModuleOrNamespaceKind.AnonModule, _ :: _ -> []
+                | _ -> visitLongIdentIncludingFullRange longIdent
 
-            let typeName =
-                match synModuleOrNamespaceKind with
-                | SynModuleOrNamespaceKind.AnonModule -> SynModuleOrNamespace_AnonModule
-                | SynModuleOrNamespaceKind.NamedModule -> SynModuleOrNamespace_NamedModule
-                | SynModuleOrNamespaceKind.DeclaredNamespace -> SynModuleOrNamespace_DeclaredNamespace
-                | SynModuleOrNamespaceKind.GlobalNamespace -> SynModuleOrNamespace_GlobalNamespace
-
-            [
-              // LongIdent inside Namespace is being processed as children.
-              if typeName <> SynModuleOrNamespace_DeclaredNamespace then
-                  mkNode typeName range
-              yield!
-                  if synModuleOrNamespaceKind = SynModuleOrNamespaceKind.DeclaredNamespace then
-                      collectIdents longIdent
-                  else
-                      []
+            [ yield! longIdentNodes
               yield! (visitSynAttributeLists range attrs)
               yield! (decls |> List.collect visitSynModuleDecl) ]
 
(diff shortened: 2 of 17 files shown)


nojaf:
@jindraivanek I went back to the drawing board when it comes to the helper function `colWithNlnWhenItemIsMultiline`.
The main limitation we currently have is that trailing trivia (comments, ifdef) would make the whole expression multiline.
So, I've implemented a new way to check if the expression is multiline is not based on the last batches of added events in the Queue.

I exposed the number of arrays in the current `Queue` via `Queue.FragmentLength`.
This is to know later in how many of the last added arrays we need to look for a newline events in `Queue.RecentItemsContain`.
Trailing trivia is skipped initially (those are the last events that were added to the Queue).
And then leading trivia are also ignored in the check.

Another thing that I changed is that the first we assume the `ColMultilineItem` is multiline to begin with.
It is cheaper to write out as multiline (adding the extra newline) and then fallback on the single line expr (if the previous and current are single line).
In case it was multiline, the context is already in the correct.

I need to do a bit more tweaks before this is ready but I'm eager to hear you feedback.

src/Fantomas.Tests/CommentTests.fs
```diff
@@ -1252,3 +1252,62 @@ let a = 8
...
+    |> prepend newline
+    |> should
+        equal
+        """
+#load "Hi.fsx"
+open Something
+
+//// FOO
+[ 1 ; 2 ; 3 ]
+
```
nojaf, line 1295: This is the reason this PR exists in the first place.
Both the newline and comment trivia are now correctly assigned to the same node.

Previously, the newline was assigned to the anonymous module and the comment to the list.
No trivia can now be assigned to the anon module anymore.

src/Fantomas/TokenParser.fs
```diff
@@ -878,6 +884,15 @@ let rec private getTriviaFromTokensThemSelves
 
         getTriviaFromTokensThemSelves mkRange allTokens nextTokens info
 
+    | KeywordString ks :: rest ->
+        let range = getRangeBetween mkRange ks ks
+
+        let info =
+            Trivia.Create(KeywordString(ks.Content)) range
```
nojaf, line 891: Capture Keyword strings like `__SOURCE_DIRECTORY__` as a custom trivia type.
They are later assigned as ContentItSelf of a trivia node.

src/Fantomas/AstTransformer.fs
```diff
@@ -9,41 +9,20 @@ open Fantomas
...
-                  mkNode typeName range
-              yield!
-                  if synModuleOrNamespaceKind = SynModuleOrNamespaceKind.DeclaredNamespace then
-                      collectIdents longIdent
-                  else
-                      []
+        | SynModuleOrNamespace (longIdent, _, kind, decls, _, attrs, _, range) ->
+            let longIdentNodes =
+                match kind, decls with
+                | SynModuleOrNamespaceKind.AnonModule, _ :: _ -> []
```
nojaf, line 22: If an anonymous module has any declarations, those will be better trivia candidates later.

src/Fantomas/Trivia.fs
```diff
@@ -315,6 +298,12 @@ let private addTriviaToTriviaNode
             findNodeBeforeLineAndColumn triviaNodes range.StartLine range.StartColumn
 
         match nodeBefore, nodeAfter with
+        | Some nb, Some na when
```
nojaf, line 301: This made sense to me, does it to you?
I check if the comment is between two nodes on its own line like
```fsharp
let b = zp land (zp)
(* Comment 2 *)
let p = p1 land (b - 1)
```

jindraivanek: make sense

jindraivanek (approved):

PR #1702, opened 2021-05-06
Small fixes

pbiggar:
I got a little confused while running through the instructions. This fixes the confusion.

Note that I couldn't find any way to specify `dotnet test --filter 'record declaration'` with a space in the name. I tried `%20`, double quotes, and adding escape characters. I don't know if it works on powershell, but it does not seem to work at all when using bash. The error was:

```
An exception occurred while invoking executor 'executor://nunit3testexecutor/': Unexpected Word 'declaration' at position 7 in selection expression.
```

Instead, I specified an issue number.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -155,6 +155,9 @@ let ``preserve compile directive between piped functions, 512`` () = ...
 - Not mandatory, but when fixing a bug consider using `fix-<issue-number>` as the git branch name.<br />
 For example, `git checkout -b fix-1404`.
 
+- Code should be formatted to our standard style, using either `dotnet fake run build.fsx -t Format` which works on all files, or
+  `dotnet fake run build.fsx -t FormatChanged` to just change the files in git.
+
 ### Small steps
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -155,6 +155,9 @@ let ``preserve compile directive between piped functions, 512`` () = ...
 - Not mandatory, but when fixing a bug consider using `fix-<issue-number>` as the git branch name.<br />
 For example, `git checkout -b fix-1404`.
 
+- Code should be formatted using Fantomas. This command should format all your code:
+  `dotnet fake run build.fsd -t Format`
```
nojaf, line 159: build.fs**x**.
Perhaps it is also worth mentioning that `-t FormatChanged` exists.
This formats all the changed files in git.

pbiggar: Good idea, I added that

nojaf (approved):

nojaf (approved):
Thank you!

PR #1834, opened 2021-07-19
Honor .fantomasignore file when processing a folder

kentcb:
When processing an entire folder with a command such as this:

```
fantomas --recurse .
```

the `.fantomasignore` file is effectively useless because the `processSourceString` function is generating a fixed file name of either `tmp.fs` or `tmp.fsi`. This file name is in turn used by a call to `IgnoreFile.isIgnoredFile` by `formatContentInternalAsync`, so a `.fantomasignore` file like this presents an issue:

```
*.fs

!/path/to/important.fs
```

Even when formatting `important.fs`, it will be "seen" as `tmp.fs` by `formatContentInternalAsync` and therefore ignored.

This PR is an intuitive attempt at fixing the issue with minimal change. That said, I do feel like the ignore check is possibly occurring in the wrong place and would be better hoisted to an outer layer of the system. In fact, `Program.fs` seemingly already does the appropriate ignore checks, so the secondary check within `formatContentInternalAsync` seems redundant in this context. However, I haven't enough experience with this system to know that one way or another without a lot of digging (open to feedback on that).

### DIFF
--- a/src/Fantomas.CoreGlobalTool.Tests/CheckTests.fs
+++ b/src/Fantomas.CoreGlobalTool.Tests/CheckTests.fs
@@ -1,5 +1,6 @@
 module Fantomas.CoreGlobalTool.Tests.CheckTests
 
+open System.IO
 open NUnit.Framework
 open FsUnit
 open Fantomas.CoreGlobalTool.Tests.TestHelpers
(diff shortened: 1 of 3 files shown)


nojaf:
I haven't checked your PR yet, but claiming `.fantomasignore` is not respected when passing in a folder seems like a bold statement.
I would expect the flow to be:

https://github.com/fsprojects/fantomas/blob/6869e2e9219adb23a0f5ce4ae110453223834b22/src/Fantomas.CoreGlobalTool/Program.fs#L447

https://github.com/fsprojects/fantomas/blob/6869e2e9219adb23a0f5ce4ae110453223834b22/src/Fantomas.CoreGlobalTool/Program.fs#L388

https://github.com/fsprojects/fantomas/blob/6869e2e9219adb23a0f5ce4ae110453223834b22/src/Fantomas.CoreGlobalTool/Program.fs#L79-L91

So, the check being `IgnoreFile.isIgnoredFile f`.

The problem you most likely have is probably related to https://github.com/markashleybell/MAB.DotIgnore/issues/9#issuecomment-767518028.

kentcb:
Please see my PR, which includes tests to prove my bold statement :)

This is the (second) check that fails:

https://github.com/fsprojects/fantomas/blob/6869e2e9219adb23a0f5ce4ae110453223834b22/src/Fantomas.Extras/FakeHelpers.fs#L56

And it's using the file name determined by:

https://github.com/fsprojects/fantomas/blob/6869e2e9219adb23a0f5ce4ae110453223834b22/src/Fantomas.CoreGlobalTool/Program.fs#L110-L114

nojaf (approved):
I stand corrected, if the file renaming happens in `processSourceString`, we cannot do the ignore check in `formatContentInternalAsync` with the renamed file.
Thank you for this contribution!

PR #1838, opened 2021-07-21
Don't crash with StackoverflowException when long triple-quoted strings are parsed

### LINKED ISSUE #1837: StackoverflowException when formatted long triple-quoted strings
**NOTE 1**: I cannot use the online tool to report this bug, since it requires a lot of code and the attempt to submit to GitHub is refused because the querystring is too long. Instead, I have created a repro [here](https://gist.github.com/kentcb/aa69871af41da6dfcd8df80d521dfb74)

**NOTE 2**: I have a fix for this, which I'll PR later. I just wanted an issue for reference

### Code

See [gist](https://gist.github.com/kentcb/aa69871af41da6dfcd8df80d521dfb74)

### Result

A `500` error. When I run locally, it's due to a `StackoverflowException`. I assume the same is true when running the online tool, though the underlying error is understandably hidden from the client.

### Problem description

Long triple-quoted strings cause a `StackoverflowException` within the code that parses interpolated strings, since it has recursive logic that is not tail-call recursive.

kentcb:
Modifies the algorithm for the `EndOfInterpolatedString` active pattern so that it is non-recursive. I do not see a way of keeping the recursion whilst attaining tail-call (nor do I like the fact that [the compiler can't yet enforce tail-call recursion](https://github.com/fsharp/fslang-suggestions/issues/721), so accidental breakage is too real a risk)

Fixes #1837

### DIFF
--- a/src/Fantomas.Tests/UtilsTests.fs
+++ b/src/Fantomas.Tests/UtilsTests.fs
@@ -117,7 +117,7 @@ SetupTesting.generateSetupScript __SOURCE_DIRECTORY__
 
 module List =
     [<Test>]
-    let ``When input is empty`` () =
+    let ``when input is empty`` () =
         let property (p: bool) : bool =
             let before, after = List.partitionWhile (fun _ _ -> p) []
             before = [] && after = []
@@ -126,7 +126,7 @@ module List =
         Check.QuickThrowOnFailure(property false)
 
     [<Test>]
-    let ``When predicate always returns false`` () =
+    let ``when predicate always returns false`` () =
         let property (xs: int list) : bool =
             let before, after =
                 List.partitionWhile (fun _ _ -> false) xs
@@ -136,7 +136,7 @@ module List =
         Check.QuickThrowOnFailure property
 
     [<Test>]
-    let ``When predicate always returns true`` () =
+    let ``when predicate always returns true`` () =
         let property (xs: int list) : bool =
             let before, after =
                 List.partitionWhile (fun _ _ -> true) xs
@@ -146,7 +146,7 @@ module List =
         Check.QuickThrowOnFailure property
 
     [<Test>]
-    let ``When predicate returns true until certain index`` () =
+    let ``when predicate returns true until certain index`` () =
         let property (xs: int list, i: int) : bool =
             let before, after =
                 List.partitionWhile (fun index _ -> i <> index) xs
--- a/src/Fantomas.Tests/UtilsTests.fs
+++ b/src/Fantomas.Tests/UtilsTests.fs
@@ -4,45 +4,47 @@ open System
 open NUnit.Framework
 open Fantomas
 open Fantomas.Tests.TestHelper
+open FsCheck
 
-let private mergeAndCompare a b expected =
-    let result =
-        String.merge Environment.NewLine a b
-        |> String.normalizeNewLine
+module String =
+    let private mergeAndCompare a b expected =
+        let result =
+            String.merge Environment.NewLine a b
+            |> String.normalizeNewLine
 
-    let normalizedExpected = String.normalizeNewLine expected
-    normalizedExpected == result
+        let normalizedExpected = String.normalizeNewLine expected
+        normalizedExpected == result
 
-[<Test>]
-let ``Merging of source code that starts with a hash`` () =
-    let a =
-        """#if NOT_DEFINED
+    [<Test>]
+    let ``Merging of source code that starts with a hash`` () =
+        let a =
+            """#if NOT_DEFINED
     printfn \"meh\"
 #else
 
 #endif
 """
 
-    let b =
-        """#if NOT_DEFINED
+        let b =
+            """#if NOT_DEFINED
 
 #else
     printfn \"foo\"
 #endif
 """
 
-    """#if NOT_DEFINED
+        """#if NOT_DEFINED
     printfn \"meh\"
 #else
     printfn \"foo\"
 #endif
 """
-    |> mergeAndCompare a b
+        |> mergeAndCompare a b
 
-[<Test>]
-let ``Merging of defines content work when source code starts with a newline`` () =
-    let a =
-        """
+    [<Test>]
+    let ``Merging of defines content work when source code starts with a newline`` () =
+        let a =
+            """
 [<Literal>]
 let private assemblyConfig() =
     #if TRACE
@@ -53,8 +55,8 @@ let private assemblyConfig() =
     x
 """
 
-    let b =
-        """
+        let b =
+            """
 [<Literal>]
 let private assemblyConfig() =
     #if TRACE
@@ -65,7 +67,7 @@ let private assemblyConfig() =
     x
 """
 
-    """
+        """
 [<Literal>]
 let private assemblyConfig() =
 #if TRACE
@@ -75,12 +77,12 @@ let private assemblyConfig() =
 #endif
     x
 """
-    |> mergeAndCompare a b
+        |> mergeAndCompare a b
 
-[<Test>]
-let ``Only split on control structure keyword`` () =
-    let a =
-        """
+    [<Test>]
+    let ``Only split on control structure keyword`` () =
+        let a =
+            """
 #if INTERACTIVE
 #else
 #load "../FSharpx.TypeProviders/SetupTesting.fsx"
@@ -91,8 +93,8 @@ SetupTesting.generateSetupScript __SOURCE_DIRECTORY__
 #endif
 """
 
-    let b =
-        """
+        let b =
+            """
 #if INTERACTIVE
 #else
 
@@ -101,7 +103,7 @@ SetupTesting.generateSetupScript __SOURCE_DIRECTORY__
 #endif
     """
 
-    """
+        """
 #if INTERACTIVE
 #else
 #load "../FSharpx.TypeProviders/SetupTesting.fsx"
@@ -111,4 +113,73 @@ SetupTesting.generateSetupScript __SOURCE_DIRECTORY__
 #load "__setup__.fsx"
 #endif
 """
-    |> mergeAndCompare a b
+        |> mergeAndCompare a b
+
+module List =
+    [<Test>]
+    let ``When n greater than or equal to list length`` () =
+        let property (xs: int list, n: int) : bool =
+            let actual = List.splitAround n xs
+
+            match actual with
+            | None -> n >= List.length xs
+            | Some (before, None) -> before = xs
+            | _ -> false
+
+        let gen =
+            gen {
+                let! xs = Arb.generate<int> |> Gen.listOf
+                let len = List.length xs
+
+                let! n =
+                    Arb.generate<int>
+                    |> Gen.filter (fun n -> n >= len)
+
+                return (xs, n)
+            }
+
+        property
+        |> Prop.forAll (Arb.fromGen gen)
+        |> Check.QuickThrowOnFailure
+
+    [<Test>]
+    let ``When n less than list length`` () =
+        let property (xs: int list, n: int) : bool =
+            let actual = List.splitAround n xs
+
+            match actual with
+            | None -> n < 0
+            | Some (_, None) -> List.isEmpty xs
+            | Some (before, Some (at, after)) ->
+                let xsLength = List.length xs
+                let beforeLength = List.length before
+                let afterLength = List.length after
+
+                beforeLength = max 0 n
+                && afterLength = xsLength - beforeLength - 1
+                && before @ (at :: after) = xs
+
+        let gen =
+            gen {
+                let! xs = Arb.generate<int> |> Gen.listOf
+                let len = List.length xs
+                let! n = Arb.generate<int> |> Gen.filter (fun n -> n < len)
+                return (xs, n)
+            }
+
+        property
+        |> Prop.forAll (Arb.fromGen gen)
+        |> Check.QuickThrowOnFailure
+
+
+    [<Test>]
+    let ``When n equals list length`` () =
+        let property (xs: int list) : bool =
+            let actual = List.splitAround (List.length xs) xs
+
+            match actual with
+            | None -> false
+            | Some (before, None) -> before = xs
+            | Some (_, Some _) -> false
+
+        Check.QuickThrowOnFailure property
--- a/src/Fantomas.Tests/UtilsTests.fs
+++ b/src/Fantomas.Tests/UtilsTests.fs
@@ -18,7 +18,7 @@ let private mergeAndCompare a b expected =
 let ``Merging of source code that starts with a hash`` () =
     let a =
         """#if NOT_DEFINED
-printfn \"meh\"
+    printfn \"meh\"
 #else
 
 #endif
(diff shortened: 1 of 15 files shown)


nojaf:
Thank you for the report of #1837 and this PR.
I'm going to ask the big guns to review this one, just to be on the safe side.
@Smaug123 or @jindraivanek are you up to review this one?

Smaug123 (approved):
If I were being super nitpicky, I'd insist on a more efficient helper function that doesn't involve indexing three times into the same list. The following is completely untested and may not even type check:

```fsharp
let splitAt<'a> (n : int) (xs : 'a list) : ('a list * ('a * 'a list) option) option =
    let rec go n heads rest =
        if n = 0 then
            match rest with
            | [] -> Some (List.rev heads, None)
            | x :: rest -> Some (List.rev heads, x, Some rest)
        else
            match rest with
            | [] -> None
            | x :: rest ->
                go (n - 1) (x :: heads) rest
    go n [] xs
```

kentcb:
Thanks @Smaug123. I'm an F# noob so nitpicky is good for my learning. I had the same concerns about indexing into an F# list and played around with a couple of things:

* `List.partition`. Decided it's not the right tool for the job, since it iterates the entire list and creates two new ones.
* Converting tokens to array so indexing/slicing was super cheap. Didn't love the result because I had to convert back to lists again, so was concerned about perf there too.

I'll play around with your suggestion over the weekend.

src/Fantomas/TokenParser.fs
```diff
@@ -547,11 +547,43 @@ let private (|InterpStringEndOrPartToken|_|) token =
...
+        if n = 0 then
+            match rest with
+            | [] -> Some(List.rev heads, None)
+            | x :: rest -> Some(List.rev heads, Some(x, rest))
+        else
+            match rest with
+            | [] -> None
+            | x :: rest -> go (n - 1) (x :: heads) rest
+
+    go n [] xs
```
kentcb, line 561: Is there a suitable place for this function? It seems like it could be useful elsewhere. If not, I could just leave as is or make it an inner function of `EndOfInterpolatedString`.

src/Fantomas/TokenParser.fs
```diff
@@ -547,11 +547,43 @@ let private (|InterpStringEndOrPartToken|_|) token =
...
+
+        match maybeLastTokenIndex with
+        | None -> None
+        | Some lastTokenIndex ->
+            match tokens |> splitAround (lastTokenIndex + 1) with
+            | Some (tokens, Some (endToken, rest)) ->
+                match tokens with
+                | [] -> None
+                | _ -> Some(tokens, endToken, rest)
+            | _ -> None
```
kentcb, line 586: I wanted to simplify this as follows:

```fsharp
| Some (head :: tail, Some (endToken, rest)) ->
    Some (head :: tail, endToken, rest)
| _ -> None
```

However, the need to re-concat `head` and `tail` immediately after pattern matching them felt a bit awkward and I'm not sure of perf implications. Any thoughts?

kentcb:
OK @nojaf @Smaug123 I've played around and got it working with some minor adjustments. I've left a couple of questions for consideration.

src/Fantomas/TokenParser.fs
Smaug123: If it were made internal and put somewhere at the top level, it could be tested!

There's a nice simple property-based test this could be covered by (assuming there's any prior art for FsCheck in Fantomas, which I don't know):

```fsharp
[<Test>]
let ``my test`` () =
    let property (xs : int list) (n : int) : bool =
        let actual = List.splitAround n xs
        match actual with
        | None -> n > List.length xs
        | Some (init, r) ->
            n <= List.length xs &&
            match r with
            | None -> xs = init
            | Some (x, tail) ->
                xs = init @ (x :: tail)
    Check.QuickThrowOnFailure property
```

Smaug123: A number of ways this property-based test could be improved, by the way. The most obvious to me: split into three tests, one for `n > List.length xs`, one for `n < List.length xs`, and one for `n = List.length xs`. For example:

```fsharp
open FsCheck
[<Test>]
let ``When n equals list length`` () =
    let property (xs : int list) : bool =
        let actual = List.splitAround (List.length xs) xs
        match actual with
        | None -> false
        | Some (init, None) -> init = xs
        | Some (init, Some _) -> false
    Check.QuickThrowOnFailure property

[<Test>]
let ``When n less than list length`` () =
    let property (xs : int list, n : int) : bool =
        let actual = List.splitAround (List.length xs) xs
        match actual with
        | None -> false
        | Some (init, None) -> init = xs
        | Some (init, Some _) -> false

    let gen =
        gen {
            let! xs = Gen.list<int> // or however you do this
            let len = List.length xs
            let! n = Gen.int |> Gen.filter (fun n -> n < len)
            return (xs, n)
        }

    property
    |> Prop.forAll (Arb.fromGen gen)
    |> Check.QuickThrowOnFailure
```

src/Fantomas/TokenParser.fs
Smaug123: You can match as follows if you like:

```fsharp
| Some (head :: tail as list, Some (endToken, rest)) ->
    Some (list, endToken, rest)
```

src/Fantomas/TokenParser.fs
nojaf: > Is there a suitable place for this function?

How about https://github.com/fsprojects/fantomas/blob/4c3aa3d6c3d26a396f1781da614000b7ec90d406/src/Fantomas/Utils.fs.

Matching unit test file: https://github.com/fsprojects/fantomas/blob/4d73011bb2d0e96b4c15cac9262dfe8b0757e0e6/src/Fantomas.Tests/UtilsTests.fs

src/Fantomas/TokenParser.fs
```diff
@@ -547,11 +547,43 @@ let private (|InterpStringEndOrPartToken|_|) token =
...
+                (function
+                | StringTextToken _
+                | InterpStringEndOrPartToken _ -> true
+                | _ -> false)
+            |> Seq.tryFindIndex
+                (function
+                | InterpStringEndOrPartToken _ -> true
+                | _ -> false)
+
+        match maybeLastTokenIndex with
```
nojaf, line 578: Perhaps go for `Option.bind` in this case?

jindraivanek:
Late to the party, but I suggest to simplify this to something along the lines:

```fsharp
let private (|MultipleStringTextTokens|_|) tokens =
    let f =
        function
        | StringTextToken _ -> true
        | _ -> false

    tokens
    |> List.takeWhile f
    |> fun xs ->
        if List.isEmpty xs then
            None
        else
            Some(xs, List.skipWhile f tokens)


let private (|EndOfInterpolatedString|_|) tokens =
    match tokens with
    | MultipleStringTextTokens (stringTokens, rest) ->
        match rest with
        | InterpStringEndOrPartToken endToken :: rest2 -> Some(stringTokens, endToken, rest2)
        | _ -> None
    | _ -> None
```

Perf of above can be improved by some utility function that returns `(List.takeWhile f, List.skipWhile f)` in one pass.

src/Fantomas/TokenParser.fs
kentcb: Oh, that's perfect - thanks!

src/Fantomas/TokenParser.fs
kentcb: Great, thanks for the suggestions.

kentcb:
OK all, I've applied feedback. I tried both implementations. My latest commit takes the approach @jindraivanek and removes the `List.splitAround` function and tests, since they wouldn't be needed in that case. Let me know which approach is preferred and any further feedback.

src/Fantomas.Tests/UtilsTests.fs
```diff
@@ -4,45 +4,47 @@ open System
 open NUnit.Framework
 open Fantomas
 open Fantomas.Tests.TestHelper
+open FsCheck
 
-let private mergeAndCompare a b expected =
-    let result =
-        String.merge Environment.NewLine a b
-        |> String.normalizeNewLine
+module String =
```
kentcb, line 9: Used a separate module here to distinguish tests against `String` versus those against `List`. Not sure if that's a standard approach - can remove if not.

src/Fantomas.Tests/UtilsTests.fs
```diff
@@ -111,4 +113,67 @@ SetupTesting.generateSetupScript __SOURCE_DIRECTORY__
...
+
+module List =
+    [<Test>]
+    let ``When input is empty`` () =
+        let property (p: bool) : bool =
+            let before, after = List.partitionWhile (fun _ _ -> p) []
+            before = [] && after = []
+
+        Check.QuickThrowOnFailure(property true)
+        Check.QuickThrowOnFailure(property false)
```
kentcb, line 126: Guessing this isn't quite idiomatic use of FsCheck, but I couldn't find a simple way to say "use true and false". I could do this:

```fsharp
Check.QuickThrowOnFailure property
```

But then it runs multiple times with the same Boolean input. I could also do:

```fsharp
[true;false]
|> List.iter (fun p -> Check.QuickThrowOnFailure(property p)
```

But not sure if that's any better.

src/Fantomas.Tests/UtilsTests.fs
nojaf: Please remove this, there are no modules in any of the other test files, so this would be a first for this project.

src/Fantomas.Tests/UtilsTests.fs
```diff
@@ -111,4 +113,67 @@ SetupTesting.generateSetupScript __SOURCE_DIRECTORY__
...
+
+    [<Test>]
+    let ``When predicate always returns false`` () =
```
nojaf, line 129: Could you lower-case the unit test name? Most of the unit tests start with a lower-case, see https://github.com/fsprojects/fantomas/blob/master/CONTRIBUTING.md#guidelines.

src/Fantomas/Utils.fs
```diff
@@ -152,6 +152,18 @@ module List =
...
+            match after with
+            | head :: tail ->
+                match f i head with
+                | true -> go (i + 1) (head :: before) tail
+                | false -> List.rev before, after
+            | [] -> List.rev before, after
+
+        go 0 [] xs
+
+
```
nojaf, line 166: Super nitpick, but could you remove one of the blank lines here 😅?

src/Fantomas.Tests/UtilsTests.fs
nojaf: @Smaug123 or @jindraivanek will have to comment on this one, I haven't really used FsCheck in a while.

nojaf:
Hi Kent, many thanks again for your time and patience to solve this issue.
Almost there!

src/Fantomas.Tests/UtilsTests.fs
kentcb: Woops, this file happens to have tests with upper-case but I've switched mine to lower. I'll do a separate PR after this one to lower-case any other stragglers.

src/Fantomas/Utils.fs
kentcb: Haha, of course! I remember reading something about Fantomas not fixing the number of new lines between syntax elements...? Can't remember why though. Anyway, fixed.

kentcb:
@nojaf Done!

src/Fantomas/Utils.fs
nojaf: No, it will currently keep the newlines you had in the source code.

src/Fantomas.Tests/UtilsTests.fs
```diff
@@ -112,3 +113,65 @@ SetupTesting.generateSetupScript __SOURCE_DIRECTORY__
...
+        let beforeLength = List.length before
+        let afterLength = List.length after
+
+        beforeLength = i
+        && afterLength = List.length xs - i
+        && before @ after = xs
+
+    let gen =
+        gen {
+            let! xs =
```
Smaug123, line 161: Tiny point, you don't have to change anything but I do think it's useful for learning the right mental model: it's generally better to *construct* the right cases rather than to *filter down* to them. Here, you could generate `x : int` and `xs : int list`, and then return `x :: xs`, constructing a nonempty list.

src/Fantomas.Tests/UtilsTests.fs
```diff
@@ -112,3 +113,65 @@ SetupTesting.generateSetupScript __SOURCE_DIRECTORY__
...
+                Arb.generate<int>
+                |> Gen.listOf
+                |> Gen.filter (fun l -> l.Length > 0)
+
+            let len = List.length xs
+
+            let! n =
+                Arb.generate<int>
+                |> Gen.filter (fun n -> n >= 0 && n < len)
```
Smaug123, line 170: Similarly here, you can avoid a lot of filtering with `|> Gen.map abs` and getting rid of the `n >= 0` filter.

src/Fantomas.Tests/UtilsTests.fs
```diff
@@ -112,3 +113,65 @@ SetupTesting.generateSetupScript __SOURCE_DIRECTORY__
...
 """
     |> mergeAndCompare a b
+
+[<Test>]
+let ``when input is empty`` () =
+    let property (p: bool) : bool =
+        let before, after = List.partitionWhile (fun _ _ -> p) []
+        before = [] && after = []
+
+    Check.QuickThrowOnFailure(property true)
```
Smaug123, line 123: You're not actually testing a property here, because there are no variables for FsCheck to generalise over. These should just be ordinary test cases, with `[<Test false>]` and `[<Test true>]`. Alternatively you could do `Check.QuickThrowOnFailure property` without supplying `true` and `false`.

src/Fantomas.Tests/UtilsTests.fs
Smaug123: The reason it's better to construct rather than filter is because:

* it's easier to reason about the distribution of the resulting test cases
* you make FsCheck do less wasted computation in creating test cases only to discard them immediately.

In this case, neither of those is a big deal, but if your condition is hard to satisfy or you're specifically testing some edge case where you want precise control over the distribution, it's better to be in the habit of constructing cases which automatically have the right property.

src/Fantomas.Tests/UtilsTests.fs
Smaug123: In fact it turns out that `Gen.choice(0, len)` is a thing, so you can just directly construct what you need.

src/Fantomas.Tests/UtilsTests.fs
Smaug123: It turns out that `Gen.nonEmptyListOf` is a thing, by the way.

kentcb: This is great - thank you! I actually already hit an issue yesterday where I attempted to filter in a way that caused FsCheck to get "stuck" trying to find a compliant value.

src/Fantomas.Tests/UtilsTests.fs
kentcb: Awesome. I need to spend some time reading the FsCheck docs now that I've got a taste for it.

Smaug123 (approved):
Nice!

nojaf:
Many thanks @kentcb for this contribution!
Let me know if you'd like me to release a new version with this fix.

PR #1843, opened 2021-07-26
Start test names with lowercase where relevant

kentcb:
As part of my [previous PR](https://github.com/fsprojects/fantomas/pull/1838), it came to light that there are tests with capitalized names where the guidelines state they should be lowercase. This PR fixes all instances, apart from those test names that start with acronyms or symbol names.


nojaf (approved):
Yes! Many thanks for this one!

PR #1845, opened 2021-07-29
Format out of process

nojaf:
This is an initial attempt at fixing #1844.
Here, a new NuGet package `Fantomas.Client` is introduced that contains `LSPFantomasService` (exposed via interface `FantomasService`).
`LSPFantomasService` starts a process and offloads the formatting over there using [LSP-types](https://github.com/kaby76/lsp-types).

The `FantomasService` can be constructed by providing the ProcessStartInfo or using the `createForWorkingDirectory` helper function. This helper function should be the main choice as it will look for a local or global `fantomas-tool`.

I still have a couple of open questions:
- Should there be a heartbeat mechanism to verify the health of the daemon process?
- What to do with logging? I believe the lsp could feed notifications back in from the process to the `FantomasService`. These could then be exposed via the interface somehow.
- What should the response be when Fantomas failed to format a file? (Incomplete AST, a bug in Fantomas). Result type? A more loose error message like a simple string or a more detailed DU type?
- The current implementation exposes all methods as F# Async, while under the hood they are `Tasks`. Does it make sense to expose Task instead?
- In the current implementation, the `parsingOptions` are created by the daemon. The relevance of the F# parsingOptions is to know if a file should have a module or namespace. [Scripts and the last file in a project](https://github.com/dotnet/fsharp/blob/d6fba3ae20c4ae8c9382eb0e65d956635e1e22f2/src/fsharp/ParseAndCheckInputs.fs#L117) do not require this. How do we want to deal with this?
- Related to the previous remark, I'm puzzled how `format selection` currently even works. I believe the filename would always need to be a script as the selection in the source is passed to parser. Not sure how that is supposed to work for `*.fsi` files.


baronfel:
My opinions, point by point:
* That would be a 'nice to have' IMO, as any critical failures could also be detected by the fantomas daemon shutting down, which would terminate the stdin/stdout pipes
* logging could be done either by the [$/logTrace](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#logTrace) or [window/logMessage](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#window_logMessage) LSP messages, and FSAC could forward those along to the calling LSP client (e.g in VScode) for display
* I think the ideal for errors in a client/server system where either side can change without recompilation is error _codes_ that can be linked to for documentation purposes. Internally Fantomas might represent these as a DU, but the external representation could look something very much like `{ message: string; code: string}`?
* No opinion here, whatever's easiest for you.
* I'm not sure what you mean here specifically, might need more guidance
* I'm not sure myself :D

src/Fantomas.Client/Contracts.fs
```diff
@@ -0,0 +1,87 @@
...
+    let FormatSelection = "fantomas/formatSelection"
+
+    [<Literal>]
+    let Configuration = "fantomas/configuration"
+
+type FormatDocumentRequest =
+    { SourceCode: string
+      /// File path will be used to identify the .editorconfig options
+      /// Unless the configuration is passed
+      FilePath: string
```
baronfel, line 25: Is this path expected to be the LSP-standard URI for file paths? https://microsoft.github.io/language-server-protocol/specification#uri

nojaf: No, not really actually as the client doesn't really know that there is an LSP thing behind the implementation.

PR #1852, opened 2021-08-04
Make record_* use consistent

pbiggar:
I don't know if this is correct, but it looks like a copy/paste that didn't get fixed.


nojaf (approved):
Thanks!

PR #1858, opened 2021-08-08
Style guide lambda

nojaf:
Hi @dsyme, in order to follow up on https://github.com/dotnet/docs/issues/25207, I've implemented the change in Fantomas to have a better idea of what the impact would be.

I'll explain the high-level gist of the implementation and perhaps you confirm if I understood the guidance correctly.
So, firstly I believe the guidance was focused around function applications where a lambda (be it SynExpr.Lambda or SynExpr.MatchLambda) is the last argument.

I've captured the nested SynExpr.App in an active pattern `(|AppWithLambda|_|)` and applied new logic to that exact shape of SynExpr. 
First, if the expression is short, we try to put it in one line.

```fsharp
functionName arg1 arg2 arg3 (fun arg4 -> bodyExpr)
```

If the `max_line_length` is crossed or the body is multiline, we apply alternative formatting.
There, we first check if the function name and arguments until the arrow of the lambda fit on the remainder of the line.

So imagine something like:
```fsharp
functionName arg1 arg2 arg3 (fun arg4 ->
```

If this all still fits on the current line, we write it as such and indent the body one level on the next line.
```fsharp
functionName arg1 arg2 arg3 (fun arg4 ->
    bodyExpr)
```

In case this doesn't fit all argument go on the next line with one level of indent:
```fsharp
functionName 
    arg1 
    arg2 
    arg3 
    (fun arg4 ->
        bodyExpr)
```
This makes a lot of sense to me, especially when `arg1, arg2 or arg3` are multiline themselves.

In case the last argument is `SynExpr.MatchLambda` we try the same thing:
```fsharp
functionName arg1 arg2 arg3 (function
    | Choice1of2 x
    | Choice2of2 y)
```

Or if the max_line_length is crossed:
```fsharp
functionName 
    arg1 
    arg2 
    arg3 
    (function
     | Choice1of2 x
     | Choice2of2 y)
```

Notice there is no extra indent on the line after the `function` keyword, the clauses do align with the `function` keyword.

If you look at the implementation, you will see the `MultiLineLambdaClosingNewline` setting.
This is to cope with the [GR style guide](https://github.com/G-Research/fsharp-formatting-conventions#formatting-function-parameter-application), as they do things differently with lambdas.
So, don't mind that 😊.

Let me know what you think, it would be a great help if you could take a look at the expected results in the unit tests.
Thanks in advance!

### DIFF
--- a/src/Fantomas.Tests/LambdaTests.fs
+++ b/src/Fantomas.Tests/LambdaTests.fs
@@ -143,15 +143,13 @@ Target.create "Clean" (fun _ ->
     |> should
         equal
         """
-Target.create
-    "Clean"
-    (fun _ ->
-        [ "bin"
-          "src/Fantomas/bin"
-          "src/Fantomas/obj"
-          "src/Fantomas.CoreGlobalTool/bin"
-          "src/Fantomas.CoreGlobalTool/obj" ]
-        |> List.iter Shell.cleanDir)
+Target.create "Clean" (fun _ ->
+    [ "bin"
+      "src/Fantomas/bin"
+      "src/Fantomas/obj"
+      "src/Fantomas.CoreGlobalTool/bin"
+      "src/Fantomas.CoreGlobalTool/obj" ]
+    |> List.iter Shell.cleanDir)
 """
 
 [<Test>]
@@ -955,4 +953,28 @@ module Foo =
         { Foo =
               blah
               |> Struct.map (fun _ (a, _, _) -> filterBackings a) }
+"""
+
+[<Test>]
+let ``multiline SynExpr.MatchLambda`` () =
+    formatSourceString
+        false
+        """
+module Foo =
+    let bar =
+        []
+        |> List.choose
+            (function
+             | _ -> "")
+"""
+        config
+    |> prepend newline
+    |> should
+        equal
+        """
+module Foo =
+    let bar =
+        []
+        |> List.choose (function
+            | _ -> "")
 """
--- a/src/Fantomas.Tests/LambdaTests.fs
+++ b/src/Fantomas.Tests/LambdaTests.fs
@@ -991,3 +991,56 @@ let foobar =
 
 let somethingElse = ()
 """
+
+[<Test>]
+let ``multiline non lambda argument`` () =
+    formatSourceString
+        false
+        """
+let argExpr =
+    col sepNln es (fun e ->
+        let genLambda
+            (pats: Context -> Context)
+            (bodyExpr: SynExpr)
+            (lpr: Range)
+            (rpr: Range option)
+            (arrowRange: Range)
+            (pr: Range)
+            : Context -> Context =
+            leadingExpressionIsMultiline (sepOpenTFor lpr -- "fun "
+                                            +> pats
+                                            +> genArrowWithTrivia
+                                                (genExprKeepIndentInBranch astContext bodyExpr)
+                                                arrowRange) (fun isMultiline ->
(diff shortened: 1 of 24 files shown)


src/Fantomas.Tests/LambdaTests.fs
```diff
@@ -561,12 +548,9 @@ Target.create "Install" (fun _ ->
     |> should
         equal
         """
-Target.create
-    "Install"
-    (fun _ ->
-        Yarn.install (fun o -> { o with WorkingDirectory = clientDir })
-        // Paket restore will already happen when the build.fsx dependencies are restored
-        )
+Target.create "Install" (fun _ -> Yarn.install (fun o -> { o with WorkingDirectory = clientDir })
```
dsyme, line 551: This one looks odd

dsyme: Feels like it should not be considered single-line expression because of the following comment

src/Fantomas.Tests/LetBindingTests.fs
```diff
@@ -1316,10 +1305,9 @@ let x =
        | _ ->
            if tcref.IsILTycon then
                tcref.ILTyconRawMetadata.CustomAttrs.AsArray
-               |> Array.exists
-                   (fun attr ->
-                       attr.Method.DeclaringType.TypeSpec.Name = typeof<TypeProviderEditorHideMethodsAttribute>
-                           .FullName)
+               |> Array.exists (fun attr ->
+                   attr.Method.DeclaringType.TypeSpec.Name = typeof<TypeProviderEditorHideMethodsAttribute>
+                       .FullName)
```
dsyme, line 1310: This break of `.FullName` looks wrong to me.  Though it's an orthogonal problem

dsyme (commented):
I looked through the test cases and in every case it looks like an improvement to me.

src/Fantomas.Tests/LetBindingTests.fs
nojaf: This is unrelated to the changes in this PR, created https://github.com/fsprojects/fantomas/issues/1863 to follow up.

src/Fantomas.Tests/LambdaTests.fs
nojaf: This is unrelated to the changes in this PR, created #1864 to follow up.

nojaf:
Thanks Don!

dsyme:
@nojaf Should we merge this?  Thanks

nojaf:
Yes, I've been on holiday for some time so that is the reason why this is has been lingering a bit.
The goal is to release this soon as alpha and update the online tool to target the `4.6` branch for the prerelease mode.

One thing that somewhat puzzles me is the disappearance of the [FCS 40 prerelease](https://www.nuget.org/packages/FSharp.Compiler.Service/40.0.1-preview.21352.5). It is unlisted on NuGet.
Any ideas on that @dsyme? See https://github.com/dotnet/fsharp/issues/11926

dsyme:
@nojaf Am asking @KevinRansom about the disappearence of the pre-release

KevinRansom:
Sorry, I unlisted it by mistake, there are some we mean to list, and some we don't meam to list and because it's hard to tell, which set something belongs to I unlisted it.

However, we have decided to eliminate the need to ship inlisted FSharp.Compiler.Service packages, so from now now on the listing status shouldn't change.

Sorry about the inconvenience, and thanks for pointing it out.

Kevin

PR #1885, opened 2021-09-08
Update max_infix_operator_expression documentation

### LINKED ISSUE #1884: Documentation contains reference to infix multiline formatter even though it's removed
In the [documentation](https://github.com/fsprojects/fantomas/blob/master/docs/Documentation.md#fsharp_max_infix_operator_expression) under the description to `fsharp_max_infix_operator_expression` the property `fsharp_infix_operator_expression_multiline_formatter` is mentioned. This property no longer exists though.

That option sounds really interesting to us, so I would be interested into the reasons behind the change. I couldn't find any info in the respective PR (https://github.com/fsprojects/fantomas/pull/1232).

But as I'm sure that there were valid reasons for this change, I suggest removing the note:
```
Requires fsharp_infix_operator_expression_multiline_formatter to be character_width to take effect.
```

panmau:
Remove mention of fsharp_infix_operator_expression_multiline_formatter as it was removed (#1232)

Resolves #1884

### DIFF
--- a/docs/Documentation.md
+++ b/docs/Documentation.md
@@ -487,8 +487,7 @@ else
 ### fsharp_max_infix_operator_expression
 
 Control the maximum length for which infix expression can be on one line.
-Default = 50. Requires `fsharp_infix_operator_expression_multiline_formatter` to
-be `character_width` to take effect.
+Default = 50.
 
 `defaultConfig`
 


nojaf:
Thanks!

Issue #1934, opened 2021-10-31
Extra space in `val` and `member` bindings in signature files


dsyme:
Default settings are adding a space after the name in
```fsharp
val x: int -> int
```
giving

```fsharp
val x : int -> int
```

Style guide examples for `val` don't have this space, e.g. https://docs.microsoft.com/en-us/dotnet/fsharp/style-guide/formatting#formatting-types-in-signatures

Also for `member`. A space is only required if there are explicit generic type parameters to prevent `f<'T>:` 

Relevant lines
* https://github.com/fsprojects/fantomas/blob/master/src/Fantomas/CodePrinter.fs#L890
* https://github.com/fsprojects/fantomas/blob/master/src/Fantomas/CodePrinter.fs#L4066
May need to be something more like this: https://github.com/fsprojects/fantomas/blob/master/src/Fantomas/CodePrinter.fs#L5325

[...]

PR #1935, opened 2021-10-31
fix spacing for val and member in signatures

dsyme:
Fixes https://github.com/fsprojects/fantomas/issues/1934


nojaf (approved):
Thanks!

dsyme:
Thank you! SO fast!

Issue #1945, opened 2021-11-02
Formatting power operator in code quotation pattern match fails


cgravill:
[...]

PR #1952, opened 2021-11-05
Fix power operator in code quotation pattern match

cgravill:
This fixes the issue with power operators in #1945

There is a bit of concerning duplication and divergence between OpName, OpNameFullInPattern, and OpNameFull before and after my change:

https://github.com/fsprojects/fantomas/blob/10f599c2d2cdeddcfb2c00b915aaf45b3a5f9be1/src/Fantomas/SourceParser.fs#L120-L175

that might need a bit of restructuring to avoid this popping up again in a slightly different place but this PR helps move the right way.

### DIFF
--- a/src/Fantomas.Tests/ActivePatternTests.fs
+++ b/src/Fantomas.Tests/ActivePatternTests.fs
@@ -84,3 +84,18 @@ let (|ParseRegex|_|) regex str =
     else
         None
 """
+
+
+[<Test>]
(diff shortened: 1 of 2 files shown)


nojaf:
Hmm, interesting. I wonder what if we really need all three of them.
I'll check that out before I merge this.
Thanks for this fix, Colin!

nojaf (approved):

Issue #1969, opened 2021-11-10
Fantomas is unable to format valid F# (.net 6.0) program


KarolBajkowski:
Hi.

I ran into some issue while formatting one of my programs. Below are details:

[...]

PR #1971, opened 2021-11-17
 Fix duplicated trivias in combination with double nested try-with statements #1969 

KarolBajkowski:


src/Fantomas.Tests/TriviaTests.fs
```diff
@@ -541,3 +542,33 @@ let x =
     | [ { Type = TriviaNodeType.MainNode SynConst_Unit
           ContentBefore = [ Directive "#if DEBUG\n\n#endif" ] } ] -> pass ()
     | _ -> Assert.Fail(sprintf "Unexpected trivia: %A" trivia)
+
+[<Test>]
+let ``double try-with, comment before inner 'with' not duplicated, 1969`` () =
+    // This test fails because of [Expected: "\r\ntr" But was:  "\ntry\n"] at start position of the string
```
KarolBajkowski, line 548: @nojaf  Do you have any hints why this test might not pass here at 'TriviaTests.fs' but identical passes in 'CompilerDirectivesTests.fs?' Are there any different configurations for different test files?

src/Fantomas.Tests/CompilerDirectivesTests.fs
```diff
@@ -2659,3 +2659,58 @@ let ``indented #if directive inside another non-indented #if directive should fo
...
+        correctlyFormatedSrc
+        config
+    |> prepend newline
+    |> should
+        equal
+        correctlyFormatedSrc
+
+[<Test>]
+let ``(copy) double try-with, inner #if directive should not throw error, 1969`` () =
+    // copied from TriviaTests.fs - here it works... No reason why...
```
KarolBajkowski, line 2690: Test copied from a different file. See the comment below...

src/Fantomas.Tests/CompilerDirectivesTests.fs
```diff
@@ -2659,3 +2659,58 @@ let ``indented #if directive inside another non-indented #if directive should fo
 #endif
 #endif
 """
+
+[<Test>]
+let ``double try-with, inner #if directive should not throw error, 1969`` () =
+    let correctlyFormatedSrc =
```
nojaf, line 2665: Please use the unit test template as listed in the [contribution guidelines](https://github.com/fsprojects/fantomas/blob/master/CONTRIBUTING.md#pull-request-ground-rules).
Even if the input and output are the same.
Just for consistency sake.

src/Fantomas.Tests/TriviaTests.fs
nojaf: Not entirely sure this will help but please format all your changed files.
Run `dotnet fantomas src -r`, you will need to do this anyway in order to pass the CI build.

src/Fantomas.Tests/TriviaTests.fs
```diff
@@ -541,3 +542,33 @@ let x =
     | [ { Type = TriviaNodeType.MainNode SynConst_Unit
           ContentBefore = [ Directive "#if DEBUG\n\n#endif" ] } ] -> pass ()
     | _ -> Assert.Fail(sprintf "Unexpected trivia: %A" trivia)
+
+[<Test>]
```
nojaf, line 546: This isn't really the correct place for this unit test, please move it to: `CommentTests.fs`

src/Fantomas/CodePrinter.fs
```diff
@@ -2030,7 +2030,13 @@ and genExpr astContext synExpr ctx =
                 +> sepNln
                 +> genExpr astContext e
                 +> unindent
-                +> kw WITH !+~ "with"
+                +> (fun ctx ->
+                    let lookupRange =
+                        ctx.MkRangeWith
```
nojaf, line 2035: This check is a bit too vague I'm afraid.
It will fail if the with keyword is higher up:
```fsharp
[<Test>]
let ``comment is still lost`` () =
    formatSourceString
        false
        """
try
    a
// comment
with


| b -> c
"""
        config
    |> prepend newline
    |> should
        equal
        """
try
    a
// comment
with


| b -> c
"""
```

I would add `mWithToLast` to the active pattern:
([parser source](https://github.com/dotnet/fsharp/blob/db2c9da8d1e76d11217d6da53a64253fd0df0246/src/fsharp/pars.fsy#L3520-L3528))

```fsharp
let (|TryWith|_|) =
    function
    | SynExpr.TryWith (e, _, cs, mWithToLast, _, _, _) -> Some(e, mWithToLast, cs)
    | _ -> None
```

and extract the with range as:
```fsharp
| TryWith (e, mWithToLast, cs) ->
 ....
                    let lookupRange =
                        ctx.MkRangeWith
                            (mWithToLast.StartLine, mWithToLast.StartColumn)
                            (mWithToLast.StartLine, mWithToLast.StartColumn + 4)
```

nojaf (changes requested):
Hello @KarolBajkowski, you are on the right track here.
Keep it up!

src/Fantomas/CodePrinter.fs
KarolBajkowski: I updated the code with your suggestion. Thank you for the help :)

src/Fantomas.Tests/TriviaTests.fs
KarolBajkowski: It should be better now. Please double check! :)

src/Fantomas.Tests/CompilerDirectivesTests.fs
KarolBajkowski: Yes I'll do that as soon as possible 👍

src/Fantomas.Tests/TriviaTests.fs
KarolBajkowski: Yes I'll do that as soon as possible 👍

src/Fantomas.Tests/CompilerDirectivesTests.fs
KarolBajkowski: Got it. I changed that.

However, I'm not sure how often that patterns `fantomas(src) = src` occur in tests but maybe it would be beneficial to introduce some "test helper" method that would simplify that kind of 'identity tests'? 🤔 I'm just loudly thinking, because maybe here it just doesn't make sense...

src/Fantomas/CodePrinter.fs
```diff
@@ -2023,14 +2023,14 @@ and genExpr astContext synExpr ctx =
...
-        | TryWith (e, cs) ->
+        | TryWith (e, mWithToLast, cs) ->
             atCurrentColumn (
                 kw TRY !- "try "
                 +> indent
                 +> sepNln
                 +> genExpr astContext e
                 +> unindent
-                +> kw WITH !+~ "with"
+                +> tokN mWithToLast WITH !+~ "with"
```
nojaf, line 2033: Close, but this still doesn't cover all cases:
```fsharp
[<Test>]
let ``nested try/with with comment on with`` () =
    formatSourceString
        false
        """
try
    a
with
| b ->
    try c
    // inner comment
    with
    | d -> ()
"""
        config
    |> prepend newline
    |> should
        equal
        """
try
    a
with
| b ->
    try
        c
    // inner comment
    with
    | d -> ()
"""
```

This will still fail, unfortunately. ([online tool](https://fsprojects.github.io/fantomas-tools/#/fantomas/preview?data=N4KABGBEDGD2AmBTSAuKAXATgTwDqdwDsIIBDfIgdwEt0ALCwgHzACMwBaAPkZLC2xhovEgHpRYaoUKJMQ2AFsFiQuhEQa9dWBbxOXMAAoAlJAA04KAGdE6dFIDmV1GADaliKD4RIAEnTYAA7IaJBS6OYeJH4AbqQANgCuIW5RfAAMFt7RAJKESKoAytQAXshZ2WAALGlgALpRAL4VnrV%2BAcEuYaqR2bEJyS7ulWAAjC3ekACypAAeADJSiPMqDvS9laMATOm1DXzNUV6T-kEpkKywsPEbfP1JKcOVWxN3hYgK1HDxsIQAgugAKL5ADyADNFjJbt4wQkbHsmhNjndTp1QpdrtCfL44g8hrUIABmV7RQqBUjQRAAIUQYNgmEQAAVSJhSMp0LIsSQsMkEQckW1UecMTcSfdBmgntkqiSfGSKdTafTlrBKLJoKQbHkYrANfZflyIDzEHySIc%2BMjokKuiKueLHgSwABWWVQeWUml0hkAVUCwUwGq1hB1euoBtdsPi8Oy%2BzNAr61vRV1FgtxEtSIwAbK7IO7FV7EABheKaqyF35WHnQdD0w1gSPR7yxiDmkiW7EdYXJu04gYOkYAdhzec9yqmH1YnIjcJNMcRR0FnZt3bFvbxksdAA5h%2BSPUqGeWfoQ6w3Z035xbF2dl5jV2n%2B5UAJw7hV-MEczDlpSkOvG00t%2BMTiXJNb1TPt8RGUZMkdXNd3zZV3k%2Bb5w0dU9-zAVtWgTYCoFtO9wI3SDxhgvM3w-RCvmuFCRj-Od%2BQXbDrxAlME3vCDNheGC8gKdAQUIAAVHAAHVaDoE8Z3QzCwHbKBE1wlcwPXDNNmJEi4L%2BTBYESfIABFEHiahPg-X9MF5Oi4wYoCmKgcIezYwjNhlGCZlmHIwX4ugVEBKNEEKOh6XQET4HWV0ql2cyAMslEcO6CJ8KUqVvFGF1nLmPIwWoWYQX9Uga0wQFZkCBkrCsMNj1dJ1wvPejL0YtEbJ6eL00SvhRmzVLZgAJUQOBMHgIKQsdMLJMA6LrNiuyCOU7JRiHDrut6%2BAADlEgUSdMHBHIOQUZxXVGEaoqtGKFESeJ7AMmQADF6QUXKP3468mofGbtxghb6XgKZTvOpZrswW67CnGDoDoFkKQ-AB9GhgrEg7aqs%2BqJqe9iZufDqNNZbAQUwRZKwGsTQqqvhmww0ajvG2zkYc7IdhzFyMdILGceoSsVrW2RNu23bHX2iLScOjtxpOs7qAuxA-oB%2B7HsU5rHS2YiRkgBmmdx9AvpFsWJbuoHFZBsHq1kKHqBhyA4bbK9EcpmXnu8LZOMVlyADU%2BypKR4EcfG603ImSBJqSZPaCnGutlHbdUh25ku7TqzK138g943BpGYa%2Bf9i3zit1ippakgticiPZh02B0AAcVsAqisQEqys9iqfYgP2ycFxG8JD6nbZSh3vtFpYqR%2BaAAGsqVZQfbCsPjCjZItrlW8rUIk1Om9kmLW6zhK5faxWlsQSgxZpdA1RUB7gj0jLCFoMq-nycd2cwbmRjQxeBeX8bV5OezpttubFYAaUQRBAhuQ8ioPIk9lCQnKPPHyZssIIwzsHNessRhbFegXbynwrB0FrkNeu9QLzmzql2UCiCbZ8C2GjRWxRCAOHiIgDSDhVoqECogVgUwECQIfgvaqFl4ZjRbgpEhoc%2BCEmgorP4BkHCECjoQGOvxiiSNyokBk-FYDcSYblMq4loFP14eTfhxD37Z0dISBWlQlZnVkIQDRMQVTUJvutU%2BUgL4Vi0Y2Ym%2BDYF8KISxQx68RiEntmY9W9gIHzDZKweApBiywFKtQ7eu8liuLPO4mqBC4E3h8SiD%2BOciThzMTpFmpBWC0PQSzOghRsCqDmEkmB0l05dBUPAcEEDCgBFoZNPxlRCT5zMcCJpEJEk5mgJgeIYJTY6LSV4jJHSkFdM7mYv%2BAC1GqDyMPUgMiCZQLcb7DxdTCHTKpp-YRm8zF93WQPCBVgNJaXyNvSsiBPrdzFhXYqpUXGulotwyKujm7eJmaQkghJv6nJZKOBkBSrDDMMlIXKDzvTn1%2BHpaAJZWT6jnpw7RXz%2BY-JfvozJVpsnGNQWY1pmAvhq3YTU1Olg6gVDCFYS6pUXBoUaEAA))

nojaf (changes requested):
Hi Karol, don't be so hard on yourself.
This is a good first PR!
Fantomas is just one of those codebases where things can be delicate from time to time.

src/Fantomas/CodePrinter.fs
KarolBajkowski: I had to miss your suggested code :/ I'm wondering why this code works even with `+3` like `(mWithToLast.StartLine, mWithToLast.StartColumn + 3)`

src/Fantomas/CodePrinter.fs
KarolBajkowski: I updated the code and test cases. Please take a look when you find some free time 🙏

src/Fantomas.Tests/TriviaTests.fs
nojaf: Thanks

src/Fantomas.Tests/CompilerDirectivesTests.fs
nojaf: There is a helper function for this, but I do not recommend it.
It can be difficult to reason about these things when the style guide changes.
So, in short, Fantomas follows https://docs.microsoft.com/en-us/dotnet/fsharp/style-guide/formatting.
If this ever changes (which spoiler it often does), then your source input should not change and only the output needs to be updated accordingly.

nojaf:
@KarolBajkowski I think this is good to go.
I noticed one typo in a unit test name but I can change that myself.
Can we convert the draft to ready PR?

KarolBajkowski:
> 
> 
> @KarolBajkowski I think this is good to go. I noticed one typo in a unit test name but I can change that myself. Can we convert the draft to ready PR?

Yes. I also thought about test naming convention, because I was not sure if the test names are readable. I'm used to `Given When Then` form so it's for me natural to first express all preconditions, then action (sometimes it's obvious so it can be omitted), and then expected outcome.

src/Fantomas.Tests/CompilerDirectivesTests.fs
KarolBajkowski: It make sense. Thanks for the explanation. I didn't think about scenario when style guide change :)

src/Fantomas/CodePrinter.fs
KarolBajkowski: Thanks for explanation! It really helps. But I must admit for me it was too much rush. I have to slowly finish fantomas youtube videos and look again at the source code and assembly all those puzzles together. I think I intuitively understand what's going on, but I need better understanding... TBH this is my first contact with compiler source code especially in F# (which I think basics I know very well but still I'm wrapping my head around some F# features/patterns which are not common in OOP)

nojaf (approved):
Many thanks, @KarolBajkowski for this first contribution!

src/Fantomas/CodePrinter.fs
nojaf: Yes, it took me a lot of time to understand this project when I first started out.
That is why I made these videos, to speed up this process for newcomers.
But I'm well aware not everything immediately clicks. Especially all those custom operators.
At first you sorta conceptually understand what they do, and later you really start understanding the events they add to the context.
I'm in the same boot that before this project, I never saw a compiler up close.
That was also a very overwhelming brave new world, but with all things, eventually, you get the hang of it.

Thanks again for all your patience and for engaging in conversation.

KarolBajkowski:
@nojaf I tried to give my best! I thank you for your patience and help :)

PR #1977, opened 2021-11-22
Restart fantomas-tool when process has exited.

nojaf:
In this PR, I'm adding a check to see if the process is still running.
Theoretically, this can only happen when the user kills the process by itself or something major unforeseen kills the process.
In any case, it worked on my machine.

I've also changed the Dispose implementation slightly as I noticed the processes weren't really killed.
I'm not sure how useful the `isCancellationRequested` check is.

@baronfel would you mind reviewing this one, please?
Thanks!

### DIFF
--- a/src/Fantomas.Client/LSPFantomasService.fs
+++ b/src/Fantomas.Client/LSPFantomasService.fs
@@ -9,9 +9,6 @@ open Fantomas.Client.Contracts
 open Fantomas.Client.LSPFantomasServiceTypes
 open Fantomas.Client.FantomasToolLocator
 
-let private orDefaultCancellationToken =
-    Option.defaultValue CancellationToken.None
-
 type Msg =
     | GetDaemon of Folder * AsyncReplyChannel<JsonRpc option>
     | Reset of AsyncReplyChannel<unit>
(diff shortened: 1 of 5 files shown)


src/Fantomas.Client/LSPFantomasService.fs
```diff
@@ -150,24 +174,33 @@ let private daemonNotFoundResponse filePath : Task<FantomasResponse> =
...
         member this.Dispose() =
-            agent.Post Reset
-            cts.Cancel()
+            if not cts.IsCancellationRequested then
+                agent.PostAndReply Reset
+                cts.Cancel()
 
         member _.VersionAsync(filePath, ?cancellationToken: CancellationToken) : Task<FantomasResponse> =
-            getFolderFor filePath
+            isCancellationRequested cts.IsCancellationRequested
```
baronfel, line 202: in addition to these checks at the start of the methods, I'd also change the `cancellationToken = orDefaultCancellationToken cancellationToken` logic in them to use `cts.Token` as the token if none is provided.

nojaf: Good catch.

Issue #1982, opened 2021-11-26
"The JSON-RPC connection with the remote party was lost before the request could complete."


pbiggar:
Using the latest version of ionide in VSCode (5.10.1), I am not able to save+format files. I get the following error in the F# output tab in VSCode when saving:

```
[Error - 9:28:33 PM] Request textDocument/formatting failed.
  Message: One or more errors occurred. (One or more errors occurred. (The JSON-RPC connection with the remote party was lost before the request could complete.))
  Code: -32603 
[21:28:33.122 WRN] [Fantomas] Errors while formatting file, defaulting to previous content. Error message was One or more errors occurred. (One or more errors occurred. (The JSON-RPC connection with the remote party was lost before the request could complete.))
System.AggregateException: One or more errors occurred. (One or more errors occurred. (The JSON-RPC connection with the remote party was lost before the request could complete.))
 ---> System.AggregateException: One or more errors occurred. (The JSON-RPC connection with the remote party was lost before the request could complete.)
 ---> StreamJsonRpc.ConnectionLostException: The JSON-RPC connection with the remote party was lost before the request could complete.
   at StreamJsonRpc.JsonRpc.InvokeCoreAsync(JsonRpcRequest request, Type expectedResultType, CancellationToken cancellationToken)
   at StreamJsonRpc.JsonRpc.InvokeCoreAsync[TResult](RequestId id, String targetName, IReadOnlyList`1 arguments, IReadOnlyList`1 positionalArgumentDeclaredTypes, IReadOnlyDictionary`2 namedArgumentDeclaredTypes, CancellationToken cancellationToken, Boolean isParameterObject)
   --- End of inner exception stack trace ---
   at System.Threading.Tasks.Task.ThrowIfExceptional(Boolean includeTaskCanceledExceptions)
   at System.Threading.Tasks.Task`1.GetResultCore(Boolean waitCompletionNotification)
   at Fantomas.Client.LSPFantomasService.Fantomas-Client-Contracts-FantomasService-FormatDocumentAsync@200-1.Invoke(Task`1 t) in C:\Users\fverdonck\Projects\fantomas\src\Fantomas.Client\LSPFantomasService.fs:line 200
   at System.Threading.Tasks.ContinuationResultTaskFromResultTask`2.InnerInvoke()
   at System.Threading.ExecutionContext.RunFromThreadPoolDispatchLoop(Thread threadPoolThread, ExecutionContext executionContext, ContextCallback callback, Object state)
[...]

PR #1983, opened 2021-11-28
Add more detailed information when the daemon fails to start.

nojaf:
In preparation for the call discussed in https://github.com/fsprojects/fantomas/issues/1982, I'd like to add some additional error handling to be able to pinpoint what step exactly is going wrong.

@baronfel would you mind taking a look at this? Just as a sanity review kinda thing.

### DIFF
--- a/src/Fantomas.Client/FantomasToolLocator.fs
+++ b/src/Fantomas.Client/FantomasToolLocator.fs
@@ -1,6 +1,7 @@
 module Fantomas.Client.FantomasToolLocator
 
 open System
+open System.ComponentModel
 open System.Diagnostics
 open System.IO
 open System.Text.RegularExpressions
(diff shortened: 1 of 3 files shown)


src/Fantomas.Client/FantomasToolLocator.fs
```diff
@@ -48,7 +49,18 @@ let private readOutputStreamAsLines (outputStream: StreamReader) : string list =
...
-let private runToolListCmd (Folder workingDir) (globalFlag: bool) =
+let private startProcess (ps: ProcessStartInfo) : Result<Process, ProcessStartError> =
+    try
+        Ok(Process.Start ps)
+    with
+    | :? Win32Exception as win32ex ->
+        let pathEnv =
+            Environment.GetEnvironmentVariable "PATH"
+
+        Error(ProcessStartError.ExecutableFileNotFound(ps.FileName, ps.Arguments, pathEnv, win32ex.Message))
```
baronfel, line 60: the working directory would be useful information for these errors as well, that would be the most complete representation of the execution environment.

nojaf: Good catch

PR #1987, opened 2021-12-01
Improve Fantomas Daemon detection

nojaf:
This PR means to address #1982.
It starts global tool using the absolute path.
And adds support for just having Fantomas on your PATH.
I should do more testing but could use a review by @baronfel if that works.


src/Fantomas.Client/FantomasToolLocator.fs
```diff
@@ -141,33 +133,124 @@ let private (|CompatibleTool|_|) lines =
...
+
+let createFor (startInfo: FantomasToolStartInfo) : Result<RunningFantomasTool, ProcessStartError> =
+    let processStart =
+        match startInfo with
+        | FantomasToolStartInfo.LocalTool (Folder workingDirectory) ->
+            let ps = ProcessStartInfo("dotnet")
+            ps.WorkingDirectory <- workingDirectory
+            ps.Arguments <- "fantomas --daemon"
+            ps
+        | FantomasToolStartInfo.GlobalTool ->
```
baronfel, line 232: in both of these new cases I think the working directory needs to be set as well, doesn't it? since the daemon will probe for configs based on that file structure?

nojaf: Actual both `.editorconfig` and `.fantomasignore` are detected from the file's point of view.
See https://github.com/fsprojects/fantomas/blob/4.6/src/Fantomas.Extras/IgnoreFile.fs
For `.editorconfig` that just works out of the box as the spec is based on all of this.

baronfel: oh great 👍

PR #1990, opened 2021-12-02
Fix 1984: Don't rewrite unchanged files

### LINKED ISSUE #1984: Fantomas writes even when not necessary
When calling fantomas with a list of files, it will write each of those files, even if no change is required.

This causes my build watcher to start a new build. I suspect it is also much slower than it could be.

I would recommend that if fantomas formats a file, and the input matches the output, that it skip writing the file.

Seen with many different versions, including v4.6.0-alpha-008+5343e280a18d679b36226f1374c3f2e2d079a029.

dawedawe:
Fixes #1984


nojaf (approved):
Thanks @dawedawe!

PR #1991, opened 2021-12-04
Put stars in long tuple in long signature at the end.

nojaf:
WIP, related to https://github.com/fsharp/fslang-design/issues/644.
Style guide would need an update before we can merge this.
@dsyme and I did discuss this in person though.


baronfel:
Re: style guide change, would that also require FSAC changes to align with this formatting?

nojaf:
That would be nice, but I'm really jumping the gun here by already having an implementation for this.
One benefit is that we discovered when you need the extra indent (after tuples) for example.

dsyme:
@nojaf I guess given our conversation at NDC we can consider the stile guide change pre-approved. I'll work on a PR next week

nojaf:
Yes, I was thinking along the same lines.

PR #2018, opened 2022-01-12
WIP: ifdef removed when used on inline keyword in function

### LINKED ISSUE #2017: ifdef removed when used on inline keyword in function
Issue created from [fantomas-online](https://fsprojects.github.io/fantomas-tools/#/fantomas/preview?data=N4KABGBEDGD2AmBTSAuKAdAdliEA2iALmFgMQCWAZmAIQAiAogEICqA4jrl%2BZnj4mUSZ4VTlwgBbAIYAHMNSkAnMAF4wAQQDOAT0zQAdNLkAKAEqJNAVzyFDs%2BQEowSyABpwUTUUI8A5ptQwAG0PCFBxKAASQm0ZZDRIHkI3UNxISIA3KTxLeODUrgAGdwiISABJYSFCAGVyAC9kEtKAFgKwAF1UgF9msPb0mLjAxMxkvrTM7NzAkNKwAEYJrkgAWSkADwAZfi2hX0IACxT5hYAmQvaurl7U8PFB2LzIIXgAeUodzEQamIITh5THJ5OalM7LNIMYQfL5NdplPCUSBXHoTe4raJPEYAI1gsDwAIxWWBs3hYAAzBCypUvIpCAAxHjZAByiAA7nxvoTxIRFLkUTc0QNMcMErj8dyykCZmhQREWlSoDVEBJyHA8LBMOpCFD3p9%2BJLcJRsl4Bbhblx0ZMhs9xQSIeliTL8vMAKyKyA1GRSaCIJiISiwRSIAAKSikEiIiEUhogvP5EWu5qFEUeoqgdsNjumILJADYPV6fX6A0HEFtYGzo9ApF5KhlYDWfJrY2B44gzRALbgrVKbTi8fbhU7c-MAOyF72%2B-2B4MsGRxRQ1uuYBtN8gtxXGvCmxOou7C-tiwdZ6Wj0oADknxZnZYAwnha5o75rNPHoIQg63t7vxEmuymgJHhmJ4OmepLzAAnNe06lsGqwqti0bfiaHZ7oKB6piKtqgcOOYQaUCzFGSnpTiWs6IC%2BGqYChO5oX%2B%2B6WoeWLHhKYEjgREQLEsJFFr66iUIQ0YvhI0itu2nZgN2-RYcBkCZux%2BGymS5wweRZbKqq6qbmSP70Vw-5SYBGJyQpeEksppyUrxZECUJiiaWq%2BI6fMEnocmmFASxIFseZzpyuICwKiRlRIGMbyYAAKoo2gAOrkEctG-gZjE9sx6bybhWEcZZhHujZxbqIosCWMIdCIHwqr2eJfL6bghnSWAvZRHJSSnjlLqEQWJHrBs5SUJFhxCAwdE1IcQaEPF8CJYqLSXO5AGeSZ3mjOMfnnlxE49ZslSUOQGxvIuUifooDAbDIwaaJoG40YqrrzQxGFMbJK1tYpFmdVxV7bRs5hwIo8BTTNZJzZJjXNWmzxvetnGBdBP1-UG8DMpYEhIYoHzlEJEgBIqCxg8Z1orRI1g%2BJyiD0kG0iEPZkVYu9-lkhcHqIwDqyk%2BQ5OU4o1PVR60CHOGH7RgA%2Bmy5DTccBNLUTGXQ9lSmfeIZw8fMaybEVihSNobyKDsb5A8cs0PSlT1pS9ctjO1isBVwZzgj9mva7r%2BuECjaPRpj2O4yp0vPV5GUkzYnP8NzvPRnTwwMxtyvWWrTs63r5Bvuzwdc1Tx18yRAtC-ZYsS4lfvmwHUNW9HsN28Fau9QAajmTA8CImC%2BIbrYXib9WpTJJcjPLgIdbbuBnPl1ebPSpUfjdDfCH4rfG0X3fLZba0Kx9g8QGc3WjxsdCwIQbBEGdF0WNdmpz2S90L016U4b5q%2BM-MZxbdXHPk0wGrQAA1kwWtf0QmgRRqBGSi%2BJUa3V0qhK%2BENsIDjvv3G2TNvpq1ZByfg-pCBViEJHRA5U9qYASjddQwgEIe0UD7eYekoE31gUOe%2BMc7bwzVgAaUQIgGQ-VBpCEqEAyMsIkp1QgA1QmfZXplxhrlCI5JiLbxGqqTQhxz7zFBgtIyMsREZTMnQiuuBySq1KJ6PwBAiq%2BFRtUWKiBsSrAQHCChkCVHg2oaxWh8C15knJA7eOfBfCYHHnoZsmA6jeOOpYYMkVYChWqMdG6-CqEW1vs4okCD5jkjjvo9QNhoyYCiRkcsmpfAkPRrgngBDXwxPscIlqK1NEuIfqUckVd9Gpx8LCLYEZsTwCkA%2BWA11m4oPJmUx6Hl-ZL3idbVxySR76LoMnKQ2ICCyOTocGouhCCbAGabIZxcRk0LGbUyRW99EsLYREsYlQf5SD0EbCBdFYk9ycbs%2BhOjn76Pfhcz%2BsJNBFRKsIVkb5EDwCaSHb4R9Lqn0wOQ0oblBmLWGbLUZ5cJHiHJEgl5ShbzBmmZoaAihyCqiyUJeALB8GanKtAR8Wt-HrM7mbRecKdkIqVlwckjD9G-BxR%2BKxSAqWCJ6B4DozREiaHpNdQIlDuhAA)

#### Code

```fsharp


    let 
#if !DEBUG
        inline
#endif
        map f ar = Async.map (Result.map f) ar
```
            

#### Result

```fsharp
let inline map f ar = Async.map (Result.map f) ar

```
            
#### Problem description

I'm adding AltCover to FsToolkit.ErrorHandling, however there is a [known issue](https://github.com/SteveGilham/altcover/issues/128) with regards to `inline` keyword and getting correct code coverage.  So the suggest fix is to add an `ifdef` around the `inline` keyword.  However, when running code formatting Fantomas removes the change. 

Yes, I hate this workaround for coverage as much as you probably do.

#### Extra information

- [ ] The formatted result breaks by code.

#### Options

Fantomas 4.6 branch at 1/1/1990

Default Fantomas configuration

<sub>Did you know that you can ignore files when formatting from fantomas-tool or the FAKE targets by using a [.fantomasignore file](https://github.com/fsprojects/fantomas/blob/master/docs/Documentation.md#ignore-files-fantomasignore)?</sub>

TheAngryByrd:
Closes #2017 

Currently WIP since I don't have this working yet and wanted to share what I have.

### DIFF
--- a/src/Fantomas.Tests/TokenParserTests.fs
+++ b/src/Fantomas.Tests/TokenParserTests.fs
@@ -658,3 +658,29 @@ let isUnix =
 
     getDefines source
     == [ "NETSTANDARD1_6"; "NETSTANDARD2_0" ]
+
+
+
+[<Test>]
+[<Category("Oracle")>]
+let ``ifdef for inline keyword`` () =
+
+    formatSourceString
+        false
+        """
+    let 
+#if !DEBUG
+        inline
+#endif
+        map f ar = Async.map (Result.map f) ar
+"""
+        config
+    |> should
+        equal
+        """
+    let 
+#if !DEBUG
+        inline
+#endif
+        map f ar = Async.map (Result.map f) ar
+"""
--- a/src/Fantomas/CodePrinter.fs
+++ b/src/Fantomas/CodePrinter.fs
@@ -5136,8 +5136,17 @@ and genSynBindingFunction
             (pref +> genOnelinerAttributes astContext ats)
 
     let afterLetKeyword =
+        let estimateRange = 
+            
+            let foo = ctx.TriviaTokenNodes |> Seq.tryHead
+            foo
+            |> Option.map(fun x -> x.Value)
+            |> Option.bind Seq.tryHead
+            |> Option.map(fun x -> x.Range)
+            |> Option.defaultValue Range.Zero
+        let xyz = tokN estimateRange INLINE (!- "inline ")
         ifElse isMutable (!- "mutable ") sepNone
-        +> ifElse isInline (!- "inline ") sepNone
+        +> ifElse isInline xyz sepNone
         +> opt sepSpace ao genAccess
 
     let genFunctionName =
--- a/src/Fantomas/TokenParser.fs
+++ b/src/Fantomas/TokenParser.fs
@@ -1200,7 +1200,8 @@ let private tokenNames =
       "WITH"
       "MEMBER"
       "AND_BANG"
-      "IN" ]
+      "IN"
+      "INLINE" ]
 
 let private tokenKinds = [ FSharpTokenCharKind.Operator ]
 
@@ -1234,6 +1235,7 @@ let internal getFsToken tokenName =
     | "INFIX_COMPARE_OP" -> INFIX_COMPARE_OP
     | "INFIX_STAR_DIV_MOD_OP" -> INFIX_STAR_DIV_MOD_OP
     | "INFIX_STAR_STAR_OP" -> INFIX_STAR_STAR_OP
+    | "INLINE" -> INLINE
     | "INT32_DOT_DOT" -> INT32_DOT_DOT
     | "LBRACE" -> LBRACE
     | "LBRACK" -> LBRACK
--- a/src/Fantomas/TriviaTypes.fs
+++ b/src/Fantomas/TriviaTypes.fs
@@ -32,6 +32,7 @@ type FsTokenType =
     | INFIX_COMPARE_OP
     | INFIX_STAR_DIV_MOD_OP
     | INFIX_STAR_STAR_OP
+    | INLINE
     | INT32_DOT_DOT
     | LBRACE
     | LBRACK


src/Fantomas.Tests/TokenParserTests.fs
```diff
@@ -658,3 +658,29 @@ let isUnix =
 
     getDefines source
     == [ "NETSTANDARD1_6"; "NETSTANDARD2_0" ]
+
+
+
+[<Test>]
```
nojaf, line 664: Please move this test to CompilerDirectiveTests.fs, it fits better over there.
Would you also mind adding the issue number in the test name? See [guidelines](https://github.com/fsprojects/fantomas/blob/master/CONTRIBUTING.md#guidelines).

src/Fantomas.Tests/TokenParserTests.fs
```diff
@@ -658,3 +658,29 @@ let isUnix =
...
+[<Category("Oracle")>]
+let ``ifdef for inline keyword`` () =
+
+    formatSourceString
```
nojaf, line 668: Fantomas will format this code twice: once with no defines and once with `DEBUG`.
It might be easier to solve one case at a time by using
```fsharp
[<Test>]
[<Category("Oracle")>]
let ``ifdef for inline keyword, no defines`` () =
    formatSourceStringWithDefines
        []
        """
    let 
#if !DEBUG
        inline
#endif
        map f ar = Async.map (Result.map f) ar
"""
        config
    |> should
        equal
        """
    let 
#if !DEBUG
        inline
#endif
        map f ar = Async.map (Result.map f) ar
"""
```

src/Fantomas/CodePrinter.fs
```diff
@@ -5136,8 +5136,17 @@ and genSynBindingFunction
             (pref +> genOnelinerAttributes astContext ats)
 
     let afterLetKeyword =
+        let estimateRange =
```
nojaf, line 5139: This proves to be more challenging than I anticipated.
In essence, you want to use the start position of the entire top-level let binding (a range that includes the `let` keyword) and the start position of the function name `map`.
You know the `inline` keyword is somewhere in that range.

image

And then you end up with something like:
```fsharp
    let afterLetKeyword =
        let genInline (ctx: Context) =
            let letBindingRange : Range = failwith "information not present at this point"
            let estimateRange = ctx.MkRange letBindingRange.Start patRange.Start
            tokN estimateRange INLINE (!- "inline ")

        ifElse isMutable (!- "mutable ") sepNone
        +> ifElse isInline genInline sepNone
        +> opt sepSpace ao genAccess
```

The main problem is that inside that function you don't have any notion of the start of the prefix.
You could be inside of a member and then instead of `let` you started with `member`.
So, you would need to pass down the start position of the parent node of SynBinding.

As there are multiple places where `genSynBindingFunction` is eventually called this would be quite tricky.

An easier route would be if the range of the inline keyword was added to the Syntax tree in the compiler.
Because now we really have to hack our way around it to find an approximate range.
This sounds like a good candidate for SyntaxTrivia, https://github.com/dotnet/fsharp/issues/12418.

nojaf (commented):
If you want to solve this one purely on Fantomas side, you will be going down the rabbit hole I'm afraid. If you are willing, we can explore it but I feel like tackling this on the compiler side seems more plausible.

TheAngryByrd:
> If you want to solve this one purely on Fantomas side, you will be going down the rabbit hole I'm afraid. If you are willing, we can explore it but I feel like tackling this on the compiler side seems more plausible.

Yeah this was a weird edge case to work around another issue in the core so I’m not feeling strong about fixing this. Fixing the original issue of F# being able to point to inline functions correctly in pdbs would be where spending the effort should go. 

Thanks for all the help and pointers with this issue! I really appreciate the coaching effort to help me learn how I can contribute. 

I’m going to go ahead and close these.

PR #2040, opened 2022-01-25
Remove LPAREN and RPAREN from TokenParser.

nojaf:
Similar to https://github.com/fsprojects/fantomas/pull/2035 and https://github.com/fsprojects/fantomas/pull/2033.

### DIFF
--- a/src/Fantomas.Tests/TypeDeclarationTests.fs
+++ b/src/Fantomas.Tests/TypeDeclarationTests.fs
@@ -2879,3 +2879,18 @@ let leBarry =
     let formatted = formatSourceString false sourceCode config
 
     formatted |> should not' (equal EmptyString)
+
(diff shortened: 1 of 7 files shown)


baronfel:
it's really great to see how the syntax tree additions are reducing the amount of token-based workarounds you had before. much clearer this way, in my opinion!

nojaf:
Yup, I really want to get rid of https://github.com/fsprojects/fantomas/blob/5cf0d016813ba942e6582b40b735ca0f398ebdcf/src/Fantomas/TokenParser.fs#L1122-L1130

These are all the trivia anchors found in the tokens, each time we want to use one of those, we need to construct an estimated range and wing it CodePrinter. So much easier if they are in the tree.

`Equals` and `With` are tackled in the next FCS.

PR #2036, opened 2022-01-22
Add space after parenthesis function call with lambda argument.

nojaf:
Fixes #2015.

@jindraivanek let me know if you want a release with this fix.

Issue #1998, opened 2021-12-21
Line of code including a backward composition operator causes error


thomasorton-i2o:
The following line within a block of code causes Fantomas to throw an error:

`(SomeModule.doSomething << SomeModule.doSomethingElse) (fun x -> x)`

[...]

PR #2060, opened 2022-02-02
Add regression test for #1998

dawedawe:
@nojaf I think you are right, #1998 was fixed by #2036
Here's the regression test for #1998


nojaf (approved):
Many thanks @dawedawe!

Issue #2067, opened 2022-02-07
Comments in anonymous record values are deleted


ArthurHNL:
[...]

PR #2076, opened 2022-02-10
Fix 2067: Comments in anonymous record values are deleted

ribeirotomas1904:
Fixes: #2067

### DIFF
--- a/src/Fantomas.Tests/RecordTests.fs
+++ b/src/Fantomas.Tests/RecordTests.fs
@@ -1894,3 +1894,27 @@ let parse (checker: FSharpChecker) (parsingOptions: FSharpParsingOptions) { File
         })
     |> Async.Parallel
 """
+
+[<Test>]
+let ``anonymous records with comments on members, 2067`` () =
+    formatSourceString
+        false
+        """
+{|
+    // The foo value.
+    FooValue = fooValue
+    // The bar value.
+    BarValue = barValue
+|}
+"""
+        config
+    |> prepend newline
+    |> should
+        equal
+        """
+{|
+// The foo value.
+FooValue = fooValue
+   // The bar value.
+BarValue = barValue |}
+"""
--- a/src/Fantomas.Tests/RecordTests.fs
+++ b/src/Fantomas.Tests/RecordTests.fs
@@ -1913,8 +1913,8 @@ let ``anonymous records with comments on record fields, 2067`` () =
         equal
         """
 {|
-// The foo value.
-FooValue = fooValue
+   // The foo value.
+   FooValue = fooValue
    // The bar value.
-BarValue = barValue |}
+   BarValue = barValue |}
 """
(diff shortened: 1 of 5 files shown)


src/Fantomas.Tests/RecordTests.fs
```diff
@@ -1894,3 +1894,27 @@ let parse (checker: FSharpChecker) (parsingOptions: FSharpParsingOptions) { File
...
+    BarValue = barValue
+|}
+"""
+        config
+    |> prepend newline
+    |> should
+        equal
+        """
+{|
+// The foo value.
```
ribeirotomas1904, line 1916: I'm having an unexpected output regarding the styling, I would guess that I have to change something around [here](https://github.com/fsprojects/fantomas/blob/master/src/Fantomas/CodePrinter.fs#L3053), can you confirm if that's the case? Also, what would be the expected styling/output in this case?

ribeirotomas1904:
When possible, please give me some feedback regarding naming (test, variables), as well as any other change necessary to make the PR acceptable

PS: Thanks for all the effort to make it easier for newcomers to contribute

src/Fantomas.Tests/RecordTests.fs
nojaf: Yes, I was expecting it more to be something like
```fsharp
{| // The foo value.
   FooValue = fooValue
   // The bar value.
   BarValue = barValue |}
```
I'll need to look into why this is happening, I'll get back to you.

src/Fantomas.Tests/RecordTests.fs
```diff
@@ -1894,3 +1894,27 @@ let parse (checker: FSharpChecker) (parsingOptions: FSharpParsingOptions) { File
         })
     |> Async.Parallel
 """
+
+[<Test>]
+let ``anonymous records with comments on members, 2067`` () =
```
nojaf, line 1899: Maybe change to `with comments on record fields`, as they are not really members.

nojaf:
Hey, you are off to a good start here! Thanks for this PR.
I'll need to look into why the result is a bit weird.

src/Fantomas.Tests/RecordTests.fs
ribeirotomas1904: Fixed on b2550b46813c5f7daece9245b1468abbee7f3307

nojaf:
Could you add this test to `MultilineBlockBracketsOnSameColumnRecordTests.fs`
```fsharp
[<Test>]
let ``anonymous records with comments on record fields`` () =
    formatSourceString
        false
        """
{|
    // The foo value.
    FooValue = fooValue
    // The bar value.
    BarValue = barValue
|}
"""
        config
    |> prepend newline
    |> should
        equal
        """
{|
    // The foo value.
    FooValue = fooValue
    // The bar value.
    BarValue = barValue
|}
"""
```
This is more likely the behaviour the user is looking for and you also fixed this in this PR.

ribeirotomas1904:
> Could you add this test to `MultilineBlockBracketsOnSameColumnRecordTests.fs`
> 
> This is more likely the behaviour the user is looking for and you also fixed this in this PR.

Yes, tonight I will take some time to add this and fix the regular one

nojaf (approved):
Many thanks for this PR @ribeirotomas1904

PR #2077, opened 2022-02-10
Always add a space for multiple curried args invocation

janus:
Addresses https://github.com/dotnet/docs/pull/28141

### DIFF
--- a/src/Fantomas/CodePrinter.fs
+++ b/src/Fantomas/CodePrinter.fs
@@ -3215,6 +3215,10 @@ and genApp astContext e es ctx =
                 (fun ctx ->
                     match es with
                     | [] -> false
+                    | Paren _ :: rest when es.Length > 1 ->
+                        match rest with
+                        | Paren _ :: _ -> true
+                        | _ -> addSpaceBeforeParensInFunCall e es.Head ctx
                     | [ h ]
                     | h :: _ -> addSpaceBeforeParensInFunCall e h ctx)
                 sepSpace
--- a/src/Fantomas.Tests/SingleExprTests.fs
+++ b/src/Fantomas.Tests/SingleExprTests.fs
@@ -213,3 +213,22 @@ let ``comment after address of tokens`` () =
 && // comment
     foobar
 """
+
+[<Test>]
+let ``function invocation with multiple curried parameters`` () =
+    formatSourceString
+        false
+        """
+module Foo =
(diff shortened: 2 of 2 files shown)


nojaf:
Hello, thank you for this contribution.

Could you please first open an issue in Fantomas instead of adding a link to a PR to the docs.
This helps for the release notes later on.

Next, please target the `4.7` branch instead of the master.
As you could not have known this, I've added some additional notes in https://github.com/fsprojects/fantomas/pull/2078.

src/Fantomas/CodePrinter.fs
```diff
@@ -3215,6 +3215,10 @@ and genApp astContext e es ctx =
                 (fun ctx ->
                     match es with
                     | [] -> false
+                    | Paren _ :: rest when es.Length > 1 ->
```
nojaf, line 3218: This is wrong, as soon as there are more than elements in `es` it should return true.
Unit test ```maintain indent if when condition is multiline``` should be updated as well.

nojaf (changes requested):
```fsharp 
       | AppWithLambda (e, es, lpr, lambda, rpr, pr) ->
```
should be updated as well to reflect these changes.
Something like:

```fsharp
            let sepSpaceAfterFunctionName =
                let sepSpaceBasedOnSetting e =
                    match e with
                    | Paren _ -> sepSpace
                    | UppercaseSynExpr -> (fun ctx -> onlyIf ctx.Config.SpaceBeforeUppercaseInvocation sepSpace ctx)
                    | LowercaseSynExpr -> (fun ctx -> onlyIf ctx.Config.SpaceBeforeLowercaseInvocation sepSpace ctx)

                match es with
                | [] -> sepSpaceBasedOnSetting e
                | _ -> sepSpace
```

src/Fantomas/CodePrinter.fs
janus: @nojaf 
It returns true if there are at least two curried args. It this fails, it continues with old logic.

nojaf: There is no need for the old logic, that is my point.

knocte:
Addressed comments in PR2077 (https://github.com/fsprojects/fantomas/pull/2077) that supersedes this one.

knocte:
err, PR2078

Issue #2075, opened 2022-02-10
type declaration loses 'when' qualifier


russellw:
[...]

PR #2080, opened 2022-02-11
Fix 2075: type declaration loses 'when' qualifier

ribeirotomas1904:
Fixes: #2075


nojaf (approved):
Many thanks @ribeirotomas1904!

PR #2082, opened 2022-02-11
Fix FSharp compiler docs link for SynExpr

saerosV:
Fixed the link for the [SynExpr Type](https://fsharp.github.io/fsharp-compiler-docs/reference/fsharp-compiler-syntax-synexpr.html) in the FSharp compiler docs. The [current link](https://fsharp.github.io/FSharp.Compiler.Service/reference/fsharp-compiler-syntaxtree-synexpr.html) is broken.


nojaf (approved):
Great catch, thank you!

PR #2085, opened 2022-02-11
Pre push hook to ensure code is formatted on upload

baronfel:
Feel free to not take this, this is a little experiment.

I added a git hook that fires before commits are pushed to a repo that invokes the `CheckFormat` target.  If that target fails, the push fails.

In order for this to work, by default the hook script would have to live in the `.git/hooks` directory, but this directory isn't sharable/tracked in git. Therefore, I made a new directory to hold the hooks, and added a build script target that can be used to set up this kind of repo-local git config (since the .git/config file can't be shared for the same reason!). One other use of this target might be to set the `.git-blame-ignore-revs` file configuration, for example.

What do you think? I would imagine if you like this I'd need to update the README as well.

### DIFF
--- a/.githooks/pre-push
+++ b/.githooks/pre-push
@@ -1,0 +1,4 @@
+#!/bin/sh
+
+# ensure that formatting is correct before pushing to the remote repository
+dotnet fake build -t CheckFormat
--- a/.githooks/pre-push
+++ b/.githooks/pre-push
@@ -1,4 +1,20 @@
 #!/bin/sh
 
+remote_name="$1"
+remote_url="$2"
+# Information about the commits which are being pushed is supplied as lines to
+# the standard input in the form:
+#
+#   <local ref> <local oid> <remote ref> <remote oid>
+# TODO: read these lines and use the list of 'refs to push' to inspect the 
+# content of those refs and only format the changed files.
+
+# sample read loop:
+# while read local_ref local_oid remote_ref remote_oid
+# do
+#   // use local_ref, local_oid, remote_ref, remote_oid to compute file changes
+# done
+
 # ensure that formatting is correct before pushing to the remote repository
+dotnet tool restore
 dotnet fake build -t CheckFormat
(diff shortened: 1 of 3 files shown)


.githooks/pre-push
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+
+# ensure that formatting is correct before pushing to the remote repository
+dotnet fake build -t CheckFormat
```
nojaf, line 4: Prepend with `dotnet tool restore` maybe?
I know very unlikely.

build.fsx
```diff
@@ -476,7 +476,10 @@ Target.create "CheckFormat" (fun _ ->
     elif result.ExitCode = 99 then
         failwith "Some files need formatting, run `dotnet fake build -t Format` to format them"
     else
-        Trace.logf "Errors while formatting: %A" result.Errors)
+        Trace.logf "Errors while formatting: %A" result.Errors
+        failwith "Unknown errors while formatting")
+
+Target.create "EnsureRepoConfig" (fun _ -> Fake.Tools.Git.CommandHelper.gitCommand "" "config core.hooksPath .githooks")
```
nojaf, line 482: We should add this to the Contribution Guidelines, section https://github.com/fsprojects/fantomas/blob/master/CONTRIBUTING.md#getting-started-with-fantomas.

nojaf (changes requested):
Thanks Chet! I'm ok with having this. Some minor nits.

.githooks/pre-push
baronfel: Good call, I've added that. I also added documentation/specs for the 'api' of this git hook, as well as noted a potential performance optimization for the future.

build.fsx
baronfel: Added, as well as documented in the build.fsx the purpose of this particular git call.

nojaf (approved):

Issue #2087, opened 2022-02-13
Space missing in curried function invocation of multiple params before 1st param


knocte:
[...]

PR #2088, opened 2022-02-13
Always add a space for multiple curried args invocation

knocte:
Fixes https://github.com/fsprojects/fantomas/issues/2087


src/Fantomas/CodePrinter.fs
```diff
@@ -3215,8 +3212,8 @@ and genApp astContext e es ctx =
                 (fun ctx ->
                     match es with
                     | [] -> false
-                    | [ h ]
-                    | h :: _ -> addSpaceBeforeParensInFunCall e h ctx)
+                    | [ h ] -> addSpaceBeforeParensInFunCall e h ctx
+                    | _ (* es.Length > 1 *)  -> true)
```
nojaf, line 3216: Nitpick but please remove the comment `(* es.Length > 1 *)`, in the context of the pattern match it is pretty obvious that  `es` contains more than 1 element.

nojaf (approved):

nojaf:
Thank you.

src/Fantomas/CodePrinter.fs
knocte: Comment removed.

PR #2092, opened 2022-02-13
Explain what's needed to close issues, that aren't reproducable anymore.

dawedawe:
Might be worse it, to explain how we handle stale issues.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -144,6 +144,9 @@ Filename: `MyNewConceptTests.fs`.
 
 When developing a new feature, add new tests to cover all code paths.
 
+If you come across an issue, which can't be reproduced with the latest version of Fantomas but is still open, please submit a regression test.
+That way, we can ensure the issue stays fixed after closing it.
+
 ## Guidelines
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -182,8 +185,8 @@ It's often better to check *before* contributing that you're setting off on the
 ## Fixing style guide inconsistencies
 
 Fantomas tries to keep up with the style guides, but as these are living documents, it can occur that something is listed in the style that Fantomas is not respecting.
-In this case, please create an issue using our [online tool](https://fsprojects.github.io/fantomas-tools/#/).
```
nojaf, line 185: I think you've removed a couple of these in other PR's (perhaps automatically via tooling), but please don't remove these spaces.
They introduce a line break (`<br />` HTML element).

nojaf (changes requested):

CONTRIBUTING.md
dawedawe: Oh, sorry. That's a setting in my VS Code. I'll keep that in mind.

nojaf (approved):
Thanks

PR #2096, opened 2022-02-14
Fix 1749 Creating anonymous record based on a function call with a list arg fails

### LINKED ISSUE #1749: Creating anonymous record based on a function call with a list arg fails
Using Fantomas v4.5.0-alpha-019. Unfortunately I can't reproduce this in the tool - I suspect because I've changed some options (.editorconfig below), but I've narrowed it down to the following:

```fsharp
module MyModule

type Foo = static member Create a = {| Name = "Isaac" |}
type Bar() = member _.Create (a,b) = ()
let bar = Bar()
type Thing = static member Stuff = 123
let x = {| Foo.Create ([ bar.Create(Thing.Stuff, Thing.Stuff) ]) with Age = 41 |}
```

I've tried removing various elements from this, some of which make the code valid e.g. removing the parens around the list in the last line.

```ini
root = true

[*]
indent_style = space
indent_size = 4
charset = utf-8
trim_trailing_whitespace = true
insert_final_newline = false

[*.fs]
fsharp_space_before_lowercase_invocation = true
fsharp_space_before_uppercase_invocation = true
fsharp_space_before_colon = true
fsharp_multiline_block_brackets_on_same_column = true

[src/Tests/*.fs]
fsharp_single_argument_web_mode = true
```

dawedawe:
Fixes #1749

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -135,6 +135,9 @@ Filename: `MyNewConceptTests.fs`.
 
 When developing a new feature, add new tests to cover all code paths.
 
+If you come across an issue, which can't be reproduced with the latest version of Fantomas but is still open, please submit a regression test.
+That way, we can ensure the issue stays fixed after closing it.
+
 ## Guidelines
 
 - Please always rebase your code on the targeted branch.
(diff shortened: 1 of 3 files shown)


nojaf (approved):
Thanks, David!

PR #2097, opened 2022-02-14
Only load the .fantomasignore file once, except for the daemon

Smaug123:
A simpler attempt at https://github.com/fsprojects/fantomas/pull/2086.

Note that this is a breaking change in behaviour; previously we would traverse the filesystem to find a `.fantomasignore`, and now we do not.


src/Fantomas.Extras/IgnoreFile.fs
```diff
@@ -31,17 +43,30 @@ module IgnoreFile =
         else
             path
 
-    let isIgnoredFile (file: string) =
-        let fullPath = Path.GetFullPath(file)
+    /// When executed from the command line, Fantomas only supports a `.fantomasignore` file in the
+    /// current working directory. This is that `.fantomasignore` file.
+    let current: Lazy<IgnoreFile option> =
+        lazy
+            let path =
```
nojaf, line 50: Perhaps we traverse upwards here as well?
That would cover the scenario where you are inside a subfolder and `.fantomasignore` at the root of the repository is still found.

nojaf (approved):
Hi again Patrick, thanks for this second attempt. It is a bit less impactful I think.
You are right that the functionality is somewhat breaking, but I think I changed it without much thought in `4.6`. I'm not sure anyone will really notice, to be honest.

src/Fantomas.Extras/IgnoreFile.fs
nojaf: I think you will actually need to do this in order to fix the failing test.

Smaug123:
(Got some yaks to shave here, sorry - FAKE's logic for finding reference assemblies is borked if your `dotnet` is in a custom location, as it is when installed through Nix.)

nojaf:
That one rings a bell actually, I got around a problem some time ago by setting `FAKE_SDK_RESOLVER_CUSTOM_DOTNET_PATH`.

Smaug123:
That looks plausible as a workaround, thanks; I've submitted a fix for the underlying issue as https://github.com/fsprojects/FAKE/pull/2662

Smaug123:
(The bump to .NET 6 has fixed a bunch of the places where I needed to do strange things to get Fantomas building, but sadly it appears to have introduced some new ones :/ )

Smaug123:
I'm going to park working on this until https://github.com/dotnet/runtime/pull/64266 is released (probably .NET 6.0.3) and FAKE starts working again on darwin-aarch64.

nojaf:
Ok, no worries. I might take a look at this myself when I find some time.

Smaug123:
@nojaf I found a Windows machine! So I fixed up the test; do you want to have a quick look again? The fix was https://github.com/fsprojects/fantomas/pull/2097/commits/11ae382f73ea77247e6294e1ab53136e0311a02a .

src/Fantomas/paket.references
```diff
@@ -1,4 +1,6 @@
 FSharp.Compiler.Service
 FSharp.Core
 Ionide.KeepAChangelog.Tasks copy_local: true
-Dotnet.ReproducibleBuilds copy_local: true
\ No newline at end of file
+Dotnet.ReproducibleBuilds copy_local: true
+System.IO.Abstractions
```
nojaf, line 5: Are these references required in `Fantomas`?
I would assume you need them in `Fantomas.Extras` instead.

paket.dependencies
```diff
@@ -7,6 +7,9 @@ nuget FSharp.Core content: none
 nuget FSharp.Compiler.Service == 41.0.3
 nuget FsUnit
 nuget FsCheck
+nuget System.IO.Abstractions 16.1.10
```
nojaf, line 10: `System.IO.Abstractions` is defined twice.
Any reason why you are locking the version?

src/Fantomas/paket.references
nojaf: image

This introduces unnecessary dependencies in `Fantomas`.

`System.IO.Abstractions` should be used in `Fantomas.Extras` and `System.IO.Abstractions.Tests` in `Fantomas.Tests` I think.

nojaf (changes requested):
This looks very good! A have a couple of concerns but nothing major.
Many thanks for picking this one up!

src/Fantomas/paket.references
Smaug123: Oh, this is simple accident - just me not being used to Paket, sorry.

nojaf (approved):
Thank you!

nojaf:
Published as https://www.nuget.org/packages/fantomas-tool/4.7.5

PR #2102, opened 2022-02-17
Use a single place to specify the MSBuild Version property

knocte:
This way changing it when releasing a new version is easier.


knocte:
Oops, I thought this would be a simple change but turns out for some reason doesn't work. Meh, I guess it's worth digging.

nojaf:
This will change anyway in https://github.com/fsprojects/fantomas/pull/2095.

PR #2117, opened 2022-02-27
Pin Paket version

Smaug123:
I did this because of FAKE's instruction (added in https://github.com/fsprojects/FAKE/issues/2193 where FAKE was changed to print out a warning):

```
Hint: Could not find a version in your paket.dependencies file, consider adding 'version 6.2.1' at the top of your dependencies file (/Users/patrick/Documents/GitHub/fantomas/paket.dependencies).
Read https://github.com/fsharp/FAKE/issues/2193 for details.
```

### DIFF
--- a/paket.dependencies
+++ b/paket.dependencies
@@ -1,3 +1,4 @@
+version 6.2.1
 source https://api.nuget.org/v3/index.json
 
 framework: netstandard2.0, net6.0, netcoreapp3.1


baronfel:
I'm not sure that this should be the guidance anymore, now that dotnet tools and tool manifests are the established way to install tools. I'd consider raising an issue in paket to detect this case and not issue the warning.

nojaf:
Yes, I don't think this is really necessary anymore.

PR #2119, opened 2022-03-02
Lock FCS version in Fantomas nuspec

### LINKED ISSUE #2118: Ping exact FCS version
In the latest release we claim that we need FCS 41.0.1 or higher:
image

This is not the case though, Fantomas will only support one exact FCS version (being `41.0.1` in this case).

Do you agree with pinning this version @baronfel?

nojaf:
Fixes #2118 

image


baronfel (approved):
LGTM!

PR #2131, opened 2022-03-07
Add  copy_local: true for Ionide.KeepAChangelog.Tasks and Dotnet.ReproducibleBuilds.

nojaf:
image

I noticed `Ionide.KeepAChangelog.Tasks` and `Dotnet.ReproducibleBuilds` are listed as dependencies, this should not be the case.
This PR tries to rectify that.


baronfel:
you should be able to upload the local package to `nuget.info` and check the dependencies before merging to make sure they're all gone, but the changes look fine to me

baronfel (approved):

nojaf:
Thanks, I verified it indeed locally using NuGet package explorer.

Issue #2135, opened 2022-03-09
IDE should be set in Release mode


nojaf:
There are a number of unit tests that require the code to run in Release mode due to some tail-recursive optimizations that only happen in Release. We should document this in the contribution guidelines.

nojaf:
//cc @josh-degraw

baronfel:
You could also add `<Configuration>Release</Configuration>` to a Directory.Build.props file at the repo root in order to have that applied to every project in the repository automatically.

nojaf:
Brilliant, thanks for the pointer!

baronfel:
It should be enough just to add it to the existing one.

These files also can be scoped 'closer' to the files/projects they modify, MSBuild will walk the file heirarchy to find the closest on to the project being built, but they don't inherit like .editorconfig does and so you have to manually encode that inheritance if you go this way.

nojaf:
Thanks @auduchinok, I like the suggestion.

PR #2137, opened 2022-03-09
Enable tailcall optimization even in debug mode

josh-degraw:
Fixes issue discussed in #2135

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -22,6 +22,7 @@ Some common use cases include:
         <PackageIcon>fantomas_logo.png</PackageIcon>
         <PackageLicenseExpression>Apache-2.0</PackageLicenseExpression>
         <PackageReadmeFile>README.md</PackageReadmeFile>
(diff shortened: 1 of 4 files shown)


Directory.Build.props
```diff
@@ -22,6 +22,7 @@ Some common use cases include:
         <PackageIcon>fantomas_logo.png</PackageIcon>
         <PackageLicenseExpression>Apache-2.0</PackageLicenseExpression>
         <PackageReadmeFile>README.md</PackageReadmeFile>
+        <Tailcalls>true</Tailcalls>
```
nojaf, line 25: Could you add it to `Fantomas.fsproj` only? It is really only there that we have code that needs this.

nojaf (changes requested):

nojaf (approved):
Thanks! Were you able to run the tests now in your IDE?
The CI build will run everything in Release mode so I can't really tell if this is working.

nojaf:
Wonderful, thanks for confirming.

Issue #2180, opened 2022-04-02
Revisit Elmish code


nojaf:
## Introduction

We currently have multiple settings to control a subset of function applications with array/lists as arguments.
The rationale behind this is described in [Formatting Elmish style guide](https://github.com/fsprojects/fantomas/blob/master/docs/Formatting-Elmish-code.md) and has a certain overlap with https://github.com/fsprojects/fantomas/issues/1408. https://github.com/fsprojects/fantomas/issues/1275 is always related in a way and it would be nice if we can consolidate everything in a general fashion.

## Proposal

As part of #1408, I propose to move and expand the current Elmish settings. To generalize and apply an alternative style for function applications that end with a list/array.

### Scope

Every (nested) [SynExp.App](https://fsharp.github.io/fsharp-compiler-docs/reference/fsharp-compiler-syntax-synexpr.html#App) that ends with one or two [SynExpr.ArrayOrListComputed](https://fsharp.github.io/fsharp-compiler-docs/reference/fsharp-compiler-syntax-synexpr.html#ArrayOrListComputed) (or [SynExpr.ArrayOrList](https://fsharp.github.io/fsharp-compiler-docs/reference/fsharp-compiler-syntax-synexpr.html#ArrayOrList)).

Examples:
```fsharp
f1 a1 [ b1 ; b2 ; b3 ]

f2 [ a1; a2; a3 ] [ b1; b2; b3 ]

f3 a b c [ d1; d2 ]
```

Arguments before the final list arguments can be of any [SynExpr](https://fsharp.github.io/fsharp-compiler-docs/reference/fsharp-compiler-syntax-synexpr.html).

### Rules

First of, try and fit the expression on one line:

```fsharp
fn a [ b1; b1 ]
```

If the expression is longer than a threshold (be it [max_line_length](https://github.com/fsprojects/fantomas/blob/master/docs/Documentation.md#max_line_length) or [fsharp_max_elmish_width](https://github.com/fsprojects/fantomas/blob/master/docs/Documentation.md#fsharp_max_elmish_width))

It should put the elements of the list on the next line indented:

```fsharp
fn a [
    b1
    b2
]
```

Assuming the function name and additional arguments (`fn a`) fit on the same line, otherwise the default formatting rules should apply.

```fsharp
fn
   a
   [
        b1
        b2
   ]
```

When there are two lists at the end, there would be slight variation where the same rules are applied.

```fsharp
// try on single line first
[...]

PR #2200, opened 2022-04-14
WIP: 2180 revisit elmish code

TheAngryByrd:
This is work in progress for #2180.


nojaf:
Hello Jimmy, thanks again for picking this up.
I'm going to write some general feedback and attention points, these are in no particular order:
- I would remove any real mention of Expecto in the unit test names. There is no desire to bend the knee for any F# project that advocates a certain style. So, I don't want to give any impressions that we are trying to fix Expecto support. The goal is broader and is about applying Stroustrup to function applications.
- Unit tests names are normally lowercased, https://github.com/fsprojects/fantomas/blob/master/CONTRIBUTING.md#unit-test
- I would build up the unit tests at a slow pace. Add a test for example where there are elements in the last list argument but everything still fits on one line. Only then move on to the next case. 
Something like:
```fsharp
functionName a b [ c; d ]
```
Going slow in variations will make sure that you approach everything test-driven and will benefit the implementation.
- `sepOpenLFixed` will not restore any trivia assigned to the `[` token.
Write a test first that covers this scenario:
```fsharp
functionName a b [ // comment
                                c; d ]
```
the result should be something like:
```fsharp
functionName a b [ // comment
    c
    d
]
```
`genTriviaFor SynExpr_ArrayOrList_OpeningDelimiter openingTokenRange (ifElse isArray sepOpenAFixed sepOpenLFixed)`
Will print the trivia, you can use [the online tool](https://fsprojects.github.io/fantomas-tools/#/trivia?data=N4KABGBEDGD2AmBTSAuKAzArgO2gFwEtZsA5AQwFtEwywAjMAbTAHoWw4Krs8AdAJ17YIEaAKEiw8cdgC6kADTgoSdAWyIAzqiiLlkApoBimgjvRkANpsQgAvkA) to figure out that you need `SynExpr_ArrayOrList_OpeningDelimiter`.
You will need to do this for both `short` and `long` expressions. The `short` needs it to become the `long` expr if you catch my drift.
- We currently have [fsharp_max_elmish_width](https://github.com/fsprojects/fantomas/blob/master/docs/Documentation.md#fsharp_max_elmish_width), to control the single line versus multiline. I think it makes sense to re-use and rename this setting. You can use something like `isShortExpression ctx.Config.MaxElmishWidth smallExpression multilineExpression ctx`, take the original code as reference.
- Add tests from both lists and arrays.
- There is also [fsharp_max_array_or_list_number_of_items](https://github.com/fsprojects/fantomas/blob/master/docs/Documentation.md#fsharp_max_array_or_list_number_of_items), which allows you to control single line vs multiline by number of elements. We have this setting, it sucks but we need to respect it as well I guess. Check the original code again and use `getListOrArrayExprSize`.
- For the active pattern, I think the logic in
https://github.com/fsprojects/fantomas/blob/7b35c8912c8a039bc443705019c7f15f5e6614c1/src/Fantomas/SourceParser.fs#L1387-L1404
can we re-used to capture what you need more efficiently.

That is all I can think of right now. I understand it can be very overwhelming, just take it in step by step, unit test by unit test.
Don't hesitate to reach out if you have any questions. 

The best of luck!

TheAngryByrd:
Do you think any formatting that was originally relying on the elmish style (such as the [ElmishTests](https://github.com/fsprojects/fantomas/blob/get-back/src/Fantomas.Tests/ElmishTests.fs)) should use `Ragnarok` config and/or should I move them to under the Ragnarok folder?

src/Fantomas/SourceParser.fs
```diff
@@ -1763,13 +1763,28 @@ let (|IndexWithoutDotExpr|ElmishReactWithoutChildren|ElmishReactWithChildren|Non
...
+    //     FunctionApplicationSingleList(ident, sr, isArray, children, er)
+    // | SynExpr.App (_,
+    //                false,
+    //                SynExpr.App (_, false, OptVar ident, (ArrayOrList _ as attributes), _),
+    //                ArrayOrList (sr, isArray, children, er, r),
+    //                _) -> FunctionApplicationDualList(ident, attributes, (isArray, sr, children, er))
+    | App(expr, exprs) ->
+        match exprs |> List.rev with
+
+        | (ArrayOrList (sr, isArray, children, er, _))::(ArrayOrList (sr2, isArray2, children2, er2, _))::tail ->
```
TheAngryByrd, line 1778: This about this, should we be hard coding single/dual array? Would it make sense to keep applying the formatting if the last args are all `ArrayOrList`? Such as

```fsharp
someFoo "This thing" [
  a
  b
] [
  a
  b
] [ 
  a
  b
]
```

src/Fantomas/CodePrinter.fs
```diff
@@ -1089,14 +1090,14 @@ and genExpr astContext synExpr ctx =
                     +> sepSpace
                     +> col sepSpace args (genExpr astContext)
                     +> sepSpace
-                    +> genArrayOrList astContext isArray openingTokenRange children closingTokenRange
+                    +> genArrayOrList astContext isArray openingTokenRange children closingTokenRange ctx.Config.MaxElmishWidth
```
nojaf, line 1093: Interesting to see that this just works. Good thinking.

src/Fantomas/SourceParser.fs
nojaf: Perhaps if you captured all arrays/lists inside a single active pattern, similar to `FunctionApplicationSingleList`, `FunctionApplicationDualList` this idea might work.

nojaf:
> Do you think any formatting that was originally relying on the elmish style (such as the [ElmishTests](https://github.com/fsprojects/fantomas/blob/get-back/src/Fantomas.Tests/ElmishTests.fs)) should use Ragnarok config and/or should I move them to under the Ragnarok folder?

Yeah, it makes sense to move this file and address those tests to the new option.

nojaf (commented):
You've made good progress, keep it up!

src/Fantomas/SourceParser.fs
TheAngryByrd: Yeah, just seeing how this is playing out it seems like a pattern that could emerge, but I'd also be fine with saying "two is the limit" at least for now.

src/Fantomas/CodePrinter.fs
TheAngryByrd: Yeah we'll see how far I get re-using that function.  Might end up diverging in the end.

PR #2243, opened 2022-05-10
Option to reduce multiple blank lines from source code

### LINKED ISSUE #2198: Reduce multiple blank lines to a single one.
I propose we add a new setting to reduce multiple blank lines to one single blank line.

The existing way of Fantomas deals with this problem is always restoring all the found blank lines.

## Pros and Cons 

The advantages of making this adjustment to Fantomas are that the number of blank lines is also more consistent. If the source text contains a multitude of blank lines, this will stay there after formatting. This is undesirable as Fantomas already has [some heuristics](https://github.com/fsprojects/fantomas/blob/get-back/docs/FormattingConventions.md#2020-revision) in place for placing blank lines. To remain consistent, the user should not be in control of this.

The disadvantage of making this adjustment to Fantomas is introducing yet another setting. As this proposal changes the default behaviour, I'm sure that some users will prefer that all blank lines are preserved.

## Examples

```fsharp
let a =
    // ...
    // ...
    42



let b = 17
```

would be formatted to
```fsharp
let a =
    // ...
    // ...
    42

let b = 17
```

## Technical remarks

In the past, I recall an experiment where no blank lines were preserved as trivia. This didn't work out in all scenarios.

```fsharp
module X

// foo
let a = 0
```

would have lead to

```fsharp
module X
// foo
let a = 0
```

Due to the way `colWithNlnWhenItemIsMultiline` works.

It seems viable to reduce the number of blank lines assigned to a trivia node, but not to remove them completely.
This reduction should be done in `Trivia` module to ensure everything in `CodePrinter` works as expected.

## Extra information

When the design is complete, this should be added to the next major branch (v5, [get-back](https://github.com/fsprojects/fantomas/tree/get-back)).

Estimated cost (XS, S, M, L, XL, XXL): S

Related suggestions: I'm pretty sure this has been requested in the past, but I could not find the issue anymore.

## Affidavit (please submit!)

* [x] I have read the [Contribution Guidelines](https://github.com/fsprojects/fantomas/blob/master/CONTRIBUTING.md).

Please tick all that apply:
* [ ] I or my company would be willing to help implement and/or test this

//cc @jindraivanek @JonCanning

jindraivanek:
Fix #2198.

Introduce new config option `KeepMaxBlankLines` that set maximal number of consecutive blank lines from source code that is kept in formatted code.

I'm open to suggestion to better name for new option.

Option is now implemented as a number, with big value as default. Is it OK, or it should be `Num option`?

`KeepMaxBlankLines=0` reveals multiple cases where Fantomas should add blank lines - I think this should be solved as separate issue(s).

Checklist:
- [x] agree on name of new option
- [x] new option: `Num` or `Num option`?
- [ ] ignored tests with issues for `KeepMaxBlankLines=0`


docs/Documentation.md
```diff
@@ -147,6 +147,7 @@ fsharp_disable_elmish_syntax=false
 fsharp_keep_indent_in_branch=false
 fsharp_blank_lines_around_nested_multiline_expressions=true
 fsharp_bar_before_discriminated_union_declaration=false
+fsharp_keep_max_blank_lines=100
```
nojaf, line 150: Maybe go for `fsharp_keep_max_number_of_blank_lines`.
We used `number_of_` before. But I don't have a strong opinion on it.
`fsharp_keep_max_blank_lines` would work for me as well.

src/Fantomas.Core.Tests/TestHelpers.fs
```diff
@@ -133,10 +133,12 @@ let tryFormatAST ast sourceCode config =
 
 let formatConfig = { FormatConfig.Default with StrictMode = true }
 
-let shouldNotChangeAfterFormat source =
+let checkFormat config source expected =
```
nojaf, line 136: Could you move this to `KeepMaxEmptyLinesTests`? It is only used there and I typically want to keep all the tests having the same structure.
This is ok to have, just make it clear that it is only used for `KeepMaxEmptyLines`.

nojaf, line 136: Could you move this to `KeepMaxEmptyLinesTests`? It is only used there and I typically want to keep all the tests having the same structure.
This is ok to have, just make it clear that it is only used for `KeepMaxEmptyLines`.

src/Fantomas.Core.Tests/KeepMaxEmptyLinesTests.fs
```diff
@@ -0,0 +1,103 @@
+﻿module Fantomas.Core.Tests.KeepMaxEmptyLines
```
nojaf, line 1: Rename to `module Fantomas.Core.Tests.KeepMaxEmptyLinesTests`

nojaf, line 1: Rename to `module Fantomas.Core.Tests.KeepMaxEmptyLinesTests`

src/Fantomas.Core/FormatConfig.fs
```diff
@@ -222,6 +222,10 @@ type FormatConfig =
       [<DisplayName("Ragnarok")>]
       Ragnarok: bool
 
+      [<Category("Convention")>]
+      [<DisplayName("Maximum number of consecutive blank lines to keep")>]
+      KeepMaxBlankLines: Num
```
nojaf, line 227: I would go for `Num` here, `Num option` was never used and looks unnecessary.

nojaf, line 227: I would go for `Num` here, `Num option` was never used and looks unnecessary.

src/Fantomas.Core/Trivia.fs
```diff
@@ -207,16 +213,32 @@ let private addTriviaToTriviaNode (startOfSourceCode: int) (triviaNodes: TriviaN
 
     // Newlines are only relevant if they occur after the first line of source code
     | { Item = Newline; Range = range } when (range.StartLine > startOfSourceCode) ->
+        let canAddNewline triviaContent =
+            let numberOfNewlinesAtEndOfContent triviaContent =
```
nojaf, line 217: I wonder if we could not collect the newlines in the first place in `collectTriviaFromBlankLines` if we know there is a limit.

If we detected a newline at lines `[ 2; 3; 5]` and the limit is 1, we could potentially only create two Newline trivia `[ Newline at 3; Newline at 5 ]`.
I'm not sure what would be more performant.

nojaf, line 217: I wonder if we could not collect the newlines in the first place in `collectTriviaFromBlankLines` if we know there is a limit.

If we detected a newline at lines `[ 2; 3; 5]` and the limit is 1, we could potentially only create two Newline trivia `[ Newline at 3; Newline at 5 ]`.
I'm not sure what would be more performant.

nojaf (changes requested):
This looks good! I've added one remark in terms of creating the Newline trivia, that might be an interesting one to consider. Otherwise, a couple of nits but nothing serious.

I would also like to see the edge cases here. I wasn't able to see the limitations that `fsharp_keep_max_blank_lines=0` has.
We could have this as an ignored test to highlight we are at least aware of when this will do weird things.

Well done!

jindraivanek:
> I would also like to see the edge cases here. I wasn't able to see the limitations that `fsharp_keep_max_blank_lines=0` has.
> We could have this as an ignored test to highlight we are at least aware of when this will do weird things.

It can be seen from my experiment here https://github.com/fsprojects/fantomas/issues/2198#issuecomment-1097272001 (it was POC, but result should be same - I try to reproduce same thing with this implementation later).

I will create ignored test highlighting problematic case from this.

nojaf:
> but result should be same

I'm not entirely sure about that, because this time, less trivia was assigned to a trivia node and I believe in the original POC the newlines were swallowed in CodePrinter.
So I'm not sure if you can still have the weird behaviour.

jindraivanek:
I made requested changes.

Fantomas.Core formatted with fsharp_keep_max_number_of_blank_lines=0 is here: https://github.com/jindraivanek/fantomas/commit/ece950fb35d2ae7544c16de0c778cc72e8a1dbde

Only thing missing is ignored test(s) with problematic cases from there.

nojaf (approved):
Spot on! Thank you.

PR #2252, opened 2022-05-13
Consider wider default format for record declarations if any XML doc comments are present on fields

### LINKED ISSUE #1879: Consider wider default format for record declarations if any XML doc comments are present on fields
From https://github.com/dotnet/docs/pull/25972#issuecomment-914249102

I propose we use wide formatting of records type declarations in implementation or signature file when any of the fields of the record type have XML comments.

### Proposal 

The proposal is to have 
* record type definitions without XML doc be formatted compactly (record definition braces on same column = false) 
* record type definitions with XML doc be formatted widely (record definition braces on same column = true).

This is roughly what was implemented for signature file generation in https://github.com/dotnet/fsharp/issues/12072.

So default formats are as today if no XML doc:
```fsharp
type PostalAddress =
    { Address: string
      City: string
      Zip: string }
```
and wide if XML doc:
```fsharp
type PostalAddress =
    {
      /// The address
      Address: string

      /// The city
      City: string

      /// The zip code
      Zip: string
    }
```

I'd also propose that wide format be used even if only one member is XML doc commented:

```fsharp
type PostalAddress =
    {
      Address: string
      City: string

      /// The zip code
      Zip: string
    }
```

###' Background

Fantomas defaults to compact format for records
```fsharp
// Declaring additional members on PostalAddress
type PostalAddress =
    { Address: string
      City: string
      Zip: string }
```

This seems a good default. However, as soon as XML doc comments are added it sufers. That is, if we follow the style guide rule that the `///` has a blank line before it, except the first, we get this:
```fsharp
// Declaring additional members on PostalAddress
type PostalAddress =
    { /// The address
      Address: string

      /// The city
      City: string

      /// The zip code
      Zip: string }
```
The only setting available to adjust this is the very widely affecting `fsharp_multiline_block_brackets_on_same_column`.   However that would be too broadly applicable - record type definitions are the only place in the syntax where XML doc comments interact with block brackets.

### Pros and cons

The main disadvantage is that this is a discontinuous approach that people may find confusing. 

The main advantage is that it supports the principal that the "destination" of code be XML-doc commented. 

### Alternatives

Make wide formatting the default for all record expressions.

nojaf:
Fixes #1879.
//cc @dsyme

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 
 * Update FCS to More cleanup, e063dd2a7005faf953619ab4f232d1e3606c7ed2
 * Rename `fsharp_ragnarok` to `fsharp_experimental_stroupstrup_style` and `fsharp_keep_indent_in_branch` to `fsharp_experimental_keep_indent_in_branch`. [#2251](https://github.com/fsprojects/fantomas/pull/2251)
+* Consider wider default format for record declarations if any XML doc comments are present on fields. [#1879](https://github.com/fsprojects/fantomas/issues/1879)
 
 ## [5.0.0-alpha-005] - 2022-05-07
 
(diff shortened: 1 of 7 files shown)


dsyme:
👍

Issue #2271, opened 2022-05-20
Refactor list of TriviaNodeAssigner to tree structure


nojaf:
Our trivia story has become a simpler one, due to the recent changes.
We now only need to puzzle blank lines, code comments and conditional directives to trivia nodes.

In `ASTTransformers` we traverse the tree and flatten it out to a list of `TriviaNodeAssigner`.
To assign the trivia to a node from this list we mostly go over all items in the list.
This actually doesn't make that much sense anymore.
Consider:

```fsharp
let a = 
    0
let b = 1
let c = 
    ignore "meh" // comment
    2
let d = 3
```

We don't need to check the children of the `SynBinding` `let a = 0` to find a matching node.
We can be smarter about this.
If we have a tree structure, we can easily rule out that we don't need to try certain branches if the top-level node doesn't remotely contain the trivia. This would speed up the trivia assignment process significantly.

The downside here is that includes touching everything in `ASTTransformer` and `Trivia`.
Different traversal strategies would probably be required for the different trivia items.

PR #2272, opened 2022-05-20
Format selection

nojaf:
I'm trying to re-enable Format selection in Fantomas.
The old code used the tokenizer we won't don't expose anymore in Fantomas.FCS so I took a different approach.

The idea is to find an AST node (of a predefined selected set of types) and format it using an anonymous module. We exact the closest selected node and format from AST to get the result.
This isn't perfect but will work for 80 - 90% of the cases.

As a side remark, I'm strongly considering https://github.com/fsprojects/fantomas/issues/2271 as well, but in a separate PR.

Initially thoughts @baronfel, @jindraivanek or @dawedawe?

TODO:
- [x] Feedback David
- [x] Add documentation to manage expectations
- [x] Format selection in signature files
- [x] Update Daemon
- [x] Sign off on public API


src/Fantomas.Core.Tests/FormattingSelectionOnlyTests.fs
```diff
@@ -1,9 +1,113 @@
...
+    |> should
+        equal
+        "
+try
+    something ()
+with ex ->
+    printfn \"caught\""
+
+/// The point of this test is to format the function application as multiline.
+/// because that is what would happen if the entire file was formatted.
```
dawedawe, line 83: because -> Because (Start with uppercase after full stop)

src/Fantomas.Core/Selection.fs
```diff
@@ -0,0 +1,123 @@
...
+        [],
+        None,
+        Range.Zero,
+        { ModuleKeyword = None
+          NamespaceKeyword = None }
+    )
+
+let private mkSynModuleDecl (expr: SynExpr) : SynModuleDecl = SynModuleDecl.Expr(expr, expr.Range)
+
+/// Wrap the selected node inside an anonymous module
```
dawedawe, line 27: Full stop after module

src/Fantomas.Core/Selection.fs
```diff
@@ -0,0 +1,123 @@
...
+            let treeWithSelection =
+                triviaNodes
+                |> List.filter (fun tna ->
+                    tna.HasFSharpASTNode
+                    && RangeHelpers.``range contains`` selection tna.Range)
+                |> List.sortByDescending (fun tna ->
+                    if tna.Range.StartLine = tna.Range.EndLine then
+                        tna.Range.EndColumn - tna.Range.StartColumn
+                    else
+                        // Calculate an artificial surface of positions they range consume.
```
dawedawe, line 88: I think you meant "the range consumes."

src/Fantomas.Core/Selection.fs
```diff
@@ -0,0 +1,123 @@
...
+                        // Take the max_line_length as size for a blank line
```
dawedawe, line 89: Missing full stop

dawedawe (commented):
Yeah, this approach makes sense to me. Just some nitpicks in the comments.

nojaf:
Thanks for taking a look @dawedawe!
@baronfel any thoughts on the return type of `FormatSelection`?
There are multiple things that could go wrong: invalid selection, unsupported AST node, a selection with compiler directives which might not end well and probably some other stuff. This is a rather complex thing to get right. The daemon could pass along some detailed info on what went wrong, not sure how far we want to go there. Thoughts?

nojaf:
Note for me: the restored trivia needs to be inside the selection. This will currently not be the case.

nojaf:
Note to self: larger selections than the range of a node should not overlap with overlap other nodes.
For example: `let myFunction() = ()`, `le`` myFunction()`` = ()`.
Keep it simple, support exact ranges and ranges with leading and or trailing whitespaces.

nojaf:
Hello @dawedawe, this is a lot to review so I won't ask that you review every file.
However, could you please maybe take a look at `formatSelection` in `Selection.fs`?
Follow the code from there on and see if it still makes sense to you.
Maybe some places could use an extra comment. Just skim through it and let me know if something is unclear.

dawedawe (commented):
Seems very well structured to me.
I hope people appreciate how much work you put into this feature.

src/Fantomas.Core.Tests/FormattingSelectionOnlyTests.fs
```diff
@@ -1,9 +1,200 @@
...
+    | ex ->
+        printfn "caught"
+
+    ()
+"""
+        config
+    |> prepend newline
+    |> should
+        equal
+        "
```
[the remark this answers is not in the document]
nojaf: Because the string ends on a `"` character.
image

nojaf:
Thanks for the review @dawedawe! I'm going to merge this one in, update the online tool and ask you to test this against your code base to detect any regressions.
I already did this against Fantomas, the F# compiler and FsAutocomplete so it should be fairly stable.
But there still could be some edge cases where a comment is lost.

dawedawe (approved):

Issue #2267, opened 2022-05-18
Long function argument with constraints is not respecting max_line_length


nojaf:
[...]

Issue #2266, opened 2022-05-18
Long type argument with constraints is not respecting max_line_length


nojaf:
[...]

PR #2263, opened 2022-05-17
Put long multiline constrained SRTP on separate lines.

nojaf:
Could you please review @dsyme if this is in line with the style guide?
I do notice that Fantomas doesn't put a space before the colon in `when ^T :`.
Is this by design for constraints?

Fixes #2230.

Issue #1133, opened 2020-09-11
long constraints really ugly!


jmagaram:
[...]

PR #2273, opened 2022-05-20
Long generic constraints

nojaf:
Hello @dsyme, this is an attempt to address https://github.com/fsprojects/fantomas/issues/2267 and https://github.com/fsprojects/fantomas/issues/2266.

I went with the original proposals to start somewhere.
At the verify least, the areas that need a change in Fantomas are found.

All in all, I don't think this is that bad and it looks consistent to me with https://github.com/fsprojects/fantomas/pull/2263.


knocte:
This seems related to https://github.com/fsprojects/fantomas/issues/1133 too, maybe it covers it as well?

dsyme (approved):
Looks good!

Issue #2207, opened 2022-04-20
Parentheses are lost in SynMeasure


sonicbhoc:
[...]

PR #2282, opened 2022-05-28
add space around binary operators in units of measure

dawedawe:
Fixes still open part of #2207

As Don has already [agreed](https://github.com/fsharp/fslang-design/issues/660#issuecomment-1138847319) to do it like this and there is a [PR](https://github.com/dotnet/docs/pull/29657) pending for the style guide clarification, I think we can move on with this.

I changed the test to use `*` instead of `/` because we have another test which already covers SynMeasure.Divide.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 * Superfluous space between negate and RationalConst. [#2265](https://github.com/fsprojects/fantomas/issues/2265)
 * Idempotency problem with ParsedHashDirective in signature file. [#2258](https://github.com/fsprojects/fantomas/issues/2258)
 * Keyword 'override' gets changed to 'member'. [#2221](https://github.com/fsprojects/fantomas/issues/2221)
+* Spaces around binary operators in units of measure. [#2207](https://github.com/fsprojects/fantomas/issues/2207)
 
 ### Changed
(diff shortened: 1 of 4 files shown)


nojaf:
Hey David, amazing that you are seeing this through.
Each individual piece of the process for a style change doesn't take that much.
But putting them together always seems a hurdle for most people, thanks for going the extra mile here.

In the spirit of leading by example, I'd like to merge this PR after the style guide gets published.
There have been some situations in the past where a link of a comment on a thread about the style guide was deemed significant to move on here on Fantomas' side.
However, I'm not a fan of this at all. It's in the guide or it isn't, how it got there is of no significance to Fantomas.

nojaf (approved):

dawedawe:
> In the spirit of leading by example, I'd like to merge this PR after the style guide gets published. There have been some situations in the past where a link of a comment on a thread about the style guide was deemed significant to move on here on Fantomas' side. However, I'm not a fan of this at all. It's in the guide or it isn't, how it got there is of no significance to Fantomas.

Sure, I'm totally okay with waiting. I was just a bit eager with this, as I'm going on vacation and I'll be without a good coding setup for quite a while.

nojaf:
Great, thanks for understanding. I'll take care of this as soon as the guide is published.

nojaf:
I just learned about the [style-guide-decided](https://github.com/fsharp/fslang-design/issues/660#issuecomment-1138847319), which is good enough for me.

Issue #2103, opened 2022-02-17
Trivia not restored after long identifier in SynField


pbiggar:
[...]

PR #2290, opened 2022-06-05
Add Regression test for variants with modules lose comments

edgarfgp:
This PR adds a Regression test for #2103 as this is fixed in the upcoming v5

### DIFF
--- a/src/Fantomas.Core.Tests/UnionTests.fs
+++ b/src/Fantomas.Core.Tests/UnionTests.fs
@@ -880,6 +880,25 @@ type Foo = // comment
 """
 
 [<Test>]
+let ``comment after variants with modules`` () =
+    formatSourceString
+        false
+        """
+type X = 
+  | A of int // This comment survives
+  | B of C.D // This comment survives
+"""
+        config
+    |> prepend newline
+    |> should
+        equal
+        """
+type X =
+    | A of int // This comment survives
+    | B of C.D // This comment survives
+"""
+
+[<Test>]
 let ``comment after equals in union`` () =
     formatSourceString
         false


src/Fantomas.Core.Tests/UnionTests.fs
```diff
@@ -879,6 +879,25 @@ type Foo = // comment
         1
 """
 
+[<Test>]
+let ``comment after variants with modules`` () =
```
nojaf, line 883: Could you add the issue number, please?
`comment after variants with modules, 2103`

nojaf (changes requested):
Hey @edgarfgp, thank you for this contribution.
Could you please add a changelog entry:
https://github.com/fsprojects/fantomas/blob/master/CONTRIBUTING.md#changelog
And could you please try to link the PR to the issue, see:
https://github.com/fsprojects/fantomas/blob/master/CONTRIBUTING.md#pull-request-title

src/Fantomas.Core.Tests/UnionTests.fs
nojaf: Actually, that test name isn't really good.
This has nothing to do with modules.
Maybe go for `restore trivia after long identifier in SynField`.
I'll update the original issue title as well.

edgarfgp: I have updated this PR following your suggestions :)

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## [Unreleased]
 
+## Changed
+* Trivia not restored after long identifier in SynField. [#2290](https://github.com/fsprojects/fantomas/pull/2290)
```
nojaf, line 6: Sorry to be picky here, but it makes more sense to put this under `Fixed`.
Changed would more be applicable if the style is changed.

edgarfgp:
My pleasure . Let me know if there is anything else I can help ;)

Issue #2264, opened 2022-05-18
Idempotency problem for unit of measure


dsyme:
[...]

PR #2302, opened 2022-06-21
Unit of measure should keep idempotency after formatting

edgarfgp:
Fixes :#2264


CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Changelog
```
nojaf, line 2: You need to add a `## [Unreleased]` header section first.

src/Fantomas.Core/CodePrinter.fs
```diff
@@ -4377,8 +4377,11 @@ and genType astContext outerBracket t =
         | [] -> sepNone
         | [ (_, t) ] -> loop t
         | (isDivide, t) :: ts ->
-            loop t -- (if isDivide then " / " else " * ")
-            +> loopTTupleList ts
+            match t with
```
nojaf, line 4380: I'm afraid this problem cannot be solved purely on Fantomas' side.
Consider https://fsprojects.github.io/fantomas-tools/#/ast?data=N4KABGBEAmCmBmBLAdrAzpAXFSAacUiaAYmolmPAIYA2as%2BEkAxgPZwWQDaAPALKwqaAK4AnWAD4AumAAuATwAOsMAAtYo2arABeMAHow9NsmgAdUWeT7DAZSoBbFQEFbAFQtXeAoWMkyFZTUNLV0wAEYDI1gTaEgQAF8gA

The AST is the same of `1 / second`, so that example would not restore correctly anymore.
As Don [mentioned](https://github.com/fsprojects/fantomas/issues/2264#issuecomment-1130263521), we should tackle this first on the AST level at the compiler side.

If you are interested in picking that up, I can help you out there if you like.

nojaf (changes requested):
Sorry, I didn't see this earlier but this cannot be fixed on Fantomas' side.

src/Fantomas.Core/CodePrinter.fs
edgarfgp: Thanks for the review :) . I will definitely would lobe to give a try fixing that in the compiler. But my SyntaxTree is very limited :( . Im sure with your help I have a better chance to fix that :) .

nojaf: I created https://github.com/dotnet/fsharp/issues/13348 to follow up on this.
I'm not quite sure what the ideal solution would be here, so I asked it there.

nojaf (approved):
Wonderful 🎉🥰

Issue #1975, opened 2021-11-20
Idempotency problem when comments are in odd places in type definition


dsyme:
Low priority isue

[...]

PR #2308, opened 2022-06-27
Add regression test for 1975 and fix double spaces

dawedawe:
While adding the regression test for #1975 I noticed double spaces in some cases.
I moved the wordXXX functions to keep them together and make use of sepSpace like @nojaf suggested.
The existing tests which expected the double spaces were adjusted.


nojaf (approved):
Thanks @dawedawe!

nojaf:
@yisusalanpng please make a note that `sepSpace` would be an interesting function to have in the documentation.

PR #2331, opened 2022-07-01
Add a quote from Chet

baronfel:
I tried to test this locally with `dotnet fsdocs watch`, but some of the styling assets wouldn't load. So I can't be sure that the carousel works 100% correctly.


nojaf (approved):
Many thanks @baronfel

Issue #2299, opened 2022-06-20
[Feature request]


dsyme:
I propose we add `fsharp_max_if_then_short_width`.  Style guide update TBD

Slack conversation with @nojaf:

> @dsyme: I'm wondering if there should be two settings `fsharp_max_if_then_short_width`  and `fsharp_max_if_then_else_short_width`. We kind of want short-if-then-else to only apply to functional code, not imperative code. Setting large `fsharp_max_if_then_else_short_width` gives things like
> 
> ```fsharp
>     if something then variable <- expr
> ```
> 
> The style guide should really be explicit that such imperative code should be multi-line, and `if .. then ... ` is always imperative. Equally people sometimes prefer one line `if condition then fail()`
> 
> @nojaf: Should that always be on multiple lines if there is no else branch?
> 
> @dsyme: I think having `fsharp_max_if_then_short_width`  with default 0 sounds right to me
> 
> 
> @nojaf: I'll take a look at that.
> 
> @nojaf: So, a couple of things come to mind here:
> 
> * You could have a nested structure without else branch `if a then b elif c`
> 
> * The code printing the if/then/else is a bit complex and there are some obscured parts that take F# 5 offset rules into account. It might be good to revisit maybe a bit more than just the split of the setting.
> 
> * Can we detect imperative code from an AST point of view and should we act upon that?
> 

## Pros and Cons 

The advantages of making this adjustment to Fantomas are more explicit formatting of some imperative code

The disadvantages of making this adjustment to Fantomas are not all cases of imperative code are covered

## Examples

Please provide multiple examples (before and after scenarios) and explain what rules did apply to achieve the outcome.

## Extra information

Estimated cost (XS, S, M, L, XL, XXL): S

Related suggestions: (put links to related suggestions here)

## Affidavit (please submit!)

Please tick this by placing a cross in the box:
* [x] I have read the [Contribution Guidelines](https://github.com/fsprojects/fantomas/blob/master/CONTRIBUTING.md).
* [x] I have [searched both open and closed suggestions on this site](https://github.com/fsprojects/fantomas/issues) and believe this is not a duplicate

Please tick all that apply:
* [x] This is not a breaking change to Fantomas
* [x] I or my company would be willing to help implement and/or test this
* [ ] This suggestion is part of the [Microsoft style guide](https://docs.microsoft.com/en-us/dotnet/fsharp/style-guide/formatting) (please add a link to section if so)
[...]

PR #2334, opened 2022-07-04
Revisit if/then/else

nojaf:
This PR introduces the `fsharp_max_if_then_short_width` setting from https://github.com/fsprojects/fantomas/issues/2299 and revisits long `ifExpr` and `matchExpr` (See https://github.com/fsharp/fslang-design/issues/646#issuecomment-1138808449).

My personal interest is more in the latter, as it is quite satisfying how much more consistently we deal with these situations.

The flow is the following:

```fsharp
if x then y
```

First `if x then` is printed, if it fits on the rest of the line it stays in one line.
If it cross the `max_line_length` boundary or is multiline it changes to

```fsharp
if
   x
then
   y
```

If the entire expression is lower than the `fsharp_max_if_then_short_width`, it stays in one line.
Otherwise, it will put the `thenExpr` on the next line:

```fsharp
if x then
   y
```

The default value for this is 0, so out of the box, it will always format this on multiple lines.

---

The same flow applies when there is an `elseExpr`, how there the existing `fsharp_max_if_then_else_short_width` setting is used there.

```fsharp
if x then y else z

// or

if x then
   y
else
   z

// or

if
   x
then
   y
else
   z
```

--- 

In case there are any `elif` branches, we check if every line is short. If so, we apply the short style otherwise the long one. Again, what happens inside the `thenExpr` will be respected.

```fsharp
if a then b
elif c then d
else if e then f

// Not that we use `fsharp_max_if_then_short_width` to check if they all fit on one line

// if not ...

if a then
   b
elif
    c // in case c is too long or multiline
then
   d
else if e then
   f
```

The same flow applies if there is an `elseExpr`, however, in there we use the `fsharp_max_if_then_else_short_width` length.

--- 

The same principle applies to `match` and `match!` expressions:

```fsharp
match x with
| _ -> ()

// if x is too long...

match 
    x
with
| _ -> ()
```

The nice thing is that `x` can be formatted the exact same way it would be as a standalone expression. This is because of the relaxations in F# 6 and sparks a lot of joy for this maintainer.

The story overall makes a lot more sense when we do the multiline thing for `ifExpr` or `matchExpr`.

@dsyme what do you think about this?
@dawedawe would you mind reviewing this?

### DIFF
--- a/docs-old/Documentation.md
+++ b/docs-old/Documentation.md
@@ -122,6 +122,7 @@ fsharp_space_after_comma=true
 fsharp_space_before_semicolon=false
 fsharp_space_after_semicolon=true
 fsharp_space_around_delimiter=true
+fsharp_max_if_then_short_width=0
 fsharp_max_if_then_else_short_width=40
 fsharp_max_infix_operator_expression=50
 fsharp_max_record_width=40
@@ -440,6 +441,26 @@ let b = [| 4;5;6 |]
 ```fsharp
 let a = [1;2;3]
 let b = [|4;5;6|]
+```
+
+### fsharp_max_if_then_short_width
+
+Control the maximum length for which if/then expression without an else expression can be on one line.  
+The [Microsoft F# style guide](https://docs.microsoft.com/en-us/dotnet/fsharp/style-guide/formatting#formatting-if-expressions) recommends to never write such an expression in one line.
+> If the else expression is absent, it is recommended to never to write the entire expression in one line.
+Default = 0.
+
+`defaultConfig`
+
+```fsharp
+if a then 
+    ()
+```
+
+`{ defaultConfig with MaxIfThenShortWidth = 15 }`
+
+```fsharp
+if a then ()
 ```
 
 ### fsharp_max_if_then_else_short_width
--- a/src/Fantomas.Core/CodePrinter.fs
+++ b/src/Fantomas.Core/CodePrinter.fs
@@ -1390,30 +1390,10 @@ and genExpr astContext synExpr ctx =
             +> sepNln
             +> genClauses astContext cs
         | Match (matchRange, e, withRange, cs) ->
-            let genMatchExpr =
-                genTriviaFor SynExpr_Match_Match matchRange !- "match "
-                +> autoIndentAndNlnWhenWriteBeforeNewlineNotEmpty (
-                    expressionFitsOnRestOfLine
-                        (genExpr astContext e
-                         +> genWithAfterMatch SynExpr_Match_With withRange)
-                        (genExprInIfOrMatch astContext e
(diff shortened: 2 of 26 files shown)


src/Fantomas.Core/CodePrinter.fs
```diff
@@ -2093,150 +2072,165 @@ and genExpr astContext synExpr ctx =
...
-                +> genExpr astContext e2
-                +> genShortElse elseOpt elseKw
+                    if isMultiline then
+                        (indent
+                         +> sepNln
+                         +> genExpr astContext thenExpr
+                         +> unindent)
+                            ctx
+                    else
+                        // Check if the entire expression is will still fit on one line, respecting MaxIfThenShortWidth
```
dawedawe, line 2093:
```suggestion
                        // Check if the entire expression will still fit on one line, respecting MaxIfThenShortWidth.
```

src/Fantomas.Core/CodePrinter.fs
```diff
@@ -2093,150 +2072,165 @@ and genExpr astContext synExpr ctx =
...
-                    exprs
-                    |> List.indexed
-                    |> List.fold
-                        (fun (acc, allLinesShort) (idx, expr) ->
-                            if allLinesShort then
-                                let lastLine, lastColumn = acc.WriterModel.Lines.Length, acc.Column
+                    if isMultiline || thenExprIsIfThenElse then
+                        long ctx
+                    else
+                        // Check if the entire expression is will still fit on one line, respecting MaxIfThenShortWidth
```
dawedawe, line 2142:
```suggestion
                        // Check if the entire expression will still fit on one line, respecting MaxIfThenShortWidth.
```

dawedawe (approved):
Nice!

dsyme:
Looks great (from PR description)

Should `fsharp_max_match_short_width` be added to allow single line `match x with A -> true | _ -> false` and make it all symmetric?

nojaf:
> Should fsharp_max_match_short_width be added to allow single line match x with A -> true | _ -> false and make it all symmetric?

This never came up in Fantomas or the style guide so I don't think the community is really waiting for that setting. If enough interest is expressed in the style guide we could have this, but for now, it feels like we can do without it.

I'll start raising a PR for the style guide to capture what is implemented here. Once that lands I'll finish up this PR.

docs-old/Documentation.md
```diff
@@ -442,6 +443,26 @@ let a = [1;2;3]
 let b = [|4;5;6|]
 ```
 
+### fsharp_max_if_then_short_width
```
nojaf, line 446: @yisusalanpng keep in mind that I added this documentation bit to the old docs.
Please move it as well in https://github.com/fsprojects/fantomas/pull/2336

Issue #2327, opened 2022-07-01
Reduce ASTContext


nojaf:
The usage of the record `ASTContext` should be avoided in `CodePrinter`:
https://github.com/fsprojects/fantomas/blob/45d068c8cfa1a4a0db19b39a2a165cbb7255377e/src/Fantomas.Core/CodePrinter.fs#L16-L39

Any PRs that remove a field of this record would be most welcome.

- [x] InterfaceRange
- [x] IsCStylePattern
- [x] IsNakedRange
- [x] IsUnionField
- [x] IsFirstTypeParam
- [x] IsInsideMatchClausePattern

This record was initially introduced to add some context from where in the AST structure the code was coming. 
For example:
```
        /// Inside a SynPat of MatchClause
        IsInsideMatchClausePattern: bool
```
tells us that we are printing a pattern where the parent node is `SynMatchClause`.
Knowing this, we have a different view on how a `SynPat.Or` should be formatted.

For example:

```fsharp
match x with
| Y _
| Z_ -> ()
// versus
| Something(moreNested = (Y _ | Z _)) -> ()
| _ -> () 
```

We want the `Or` pattern in the first case, (the sole pattern of SynMatchClause) to be formatted on multiple lines.
While the `Or` in `Something(moreNested = (Y _ | Z _))` should be tackled differently.

Instead of using `ASTContext` to distinguish these cases, we should better introduce a particular function that is called from the parent node.

The whole problem with using `ASTContext` to differentiate between the cases, is that you need to reset that flag at the right time or it leads to very confusing bugs.

Some fields can probably be easily refactored, others might need some more detailed changes.

PR #2344, opened 2022-07-07
Remove the IsNakedRange field from the ASTContext type.

dawedawe:
It was only set once but never checked.

First stab at #2327


nojaf:
Wunderbar! The world is a better place with this.

nojaf (approved):

PR #2350, opened 2022-07-08
Mark config as optional in public API.

nojaf:
The main motivation for this is the `FormatASTAsync` function, which is used primarily by projects that generate code. They do not have any source input and typically don't care about the output so use they use the default configuration.


dawedawe (approved):
lgtm

Issue #1186, opened 2020-10-07
Using a compiler directive copies the previous(?) line in fsi files


vsinha:
[...]

Issue #801, opened 2020-05-01
Content of #if block removed - Breaking Code on reformat


sadgit:
[...]

PR #2352, opened 2022-07-08
Add regression tests for 801, 1186

dawedawe:
regression test for #801
regression test for #1186


nojaf (approved):
Thank you! Would you mind adding an entry to the changelog for this?

Issue #1775, opened 2021-06-12
Mutliline if expression with multiline infix operations


nojaf:
[...]

PR #2353, opened 2022-07-09
Add regression test for 1775

dawedawe:
regression test for #1775


nojaf (approved):
Thanks! Could you as well drop a line in the changelog here.

Issue #2374, opened 2022-07-19
FSharp.Core version conflict between Fantomas.Core and Fantomas.FCS


thinkbeforecoding:
I'm trying to use Fantomas.Core to format the FSharp.Data.WsdlGenerator generated code, but I have a problem with the references.

The Fantomas.Core nuget has a reference on FSharp.Core with a >= 6.0.3 version restriction, and it also references Fantomas.FCS that has a ref on FSharp.Core with a == 6.0.1 restriction. (using versoin 5.0.0-beta-001, but it is the same with all version of Fantomas.Core)

These two restrictions are in conflict and cause problem both when adding the nuget, and at runtime.

What is the necessity of the strict restriction in Fantomas.FCS ? If this is necessary, Fantomas.Core should also apply this restriction.

Today, Fantomas.Core uses the dependencies from the main group, while Fantomas.FCS uses dependencies from the compiler group. This explains the different versions.

thinkbeforecoding:
PR submitted 😄

PR #2375, opened 2022-07-19
Fix: use same constraint >= 6.0.1 for Fantomas.Core and Fantomas.FCS

thinkbeforecoding:
This is a fix for #2374


thinkbeforecoding:
@nojaf  that was a small fix 😄

thinkbeforecoding:
This time it should be better

nojaf (approved):
Yup, wonderful!
Thank you.

nojaf:
@thinkbeforecoding I've pushed beta 2 to NuGet, it will be available once the indexing is done.

PR #2377, opened 2022-07-20
New docs page custom theme (WIP)

yisusalanpng:
**TBD plans for documentation:**
- Completely redesigned sidebar
- New navbar with urls, collapsing button for sidebar,and search bar
- New doc sections
- Own custom styles


docs/_template.html
```diff
@@ -94,9 +89,7 @@
     src="https://cdnjs.cloudflare.com/ajax/libs/JavaScript-autoComplete/1.0.4/auto-complete.min.js"></script>
   <script type="text/javascript" src="{{root}}content/fsdocs-search.js"></script>
   <!-- END SEARCH BOX: this adds support for the search box -->
-  <script src="https://cdn.jsdelivr.net/npm/bootstrap@5.0.2/dist/js/bootstrap.bundle.min.js"
-    integrity="sha384-MrcW6ZMFYlzcLA8Nl+NtUVF0sA7MsXsP1UyJoMp4YLEuNSfAP+JcXn/tWtIaxVXM"
-    crossorigin="anonymous"></script>
+  <script src="https://cdn.jsdelivr.net/npm/bootstrap@5.2.0/dist/js/bootstrap.bundle.min.js" integrity="sha384-A3rJD856KowSb7dwlZdYEkO39Gagi7vIsF0jrRAoQmDKKtQBHUuLZ9AsSv4jD4Xa" crossorigin="anonymous"></script>
```
nojaf, line 100: If you really want to bump bootstrap, you would also need to update the local files at https://github.com/fsprojects/fantomas/blob/8593cb20c867699bedc6a9c5c735fa4532db27ed/docs/.style/bootstrap/scss/bootstrap.scss.

I'm ok with updating this version of bootstrap, although you typically won't update this without any real reason? Do you have one in this case?

docs/_template.html
```diff
@@ -2,84 +2,102 @@
...
-<script type="text/javascript" src="{{root}}content/fsdocs-search.js"></script>
-<!-- END SEARCH BOX: this adds support for the search box -->
-<script src="https://cdn.jsdelivr.net/npm/bootstrap@5.0.2/dist/js/bootstrap.bundle.min.js"
-        integrity="sha384-MrcW6ZMFYlzcLA8Nl+NtUVF0sA7MsXsP1UyJoMp4YLEuNSfAP+JcXn/tWtIaxVXM"
-        crossorigin="anonymous"></script>
+    
+  </main>
+  <script type="module" src="{{root}}content/sidebar-generation.js"></script>
+  <script type="text/javascript">
+var sidebar = document.getElementById('sidebar-wrapper')
```
nojaf, line 82: `var` is a thing of the past. You now mostly want to use `const` or `let` (when variable is mutable).

docs/_template.html
```diff
@@ -2,84 +2,102 @@
...
+sidebar.addEventListener('hide.bs.collapse', function (e) {
```
nojaf, line 83: `hide.bs.collapse` is a custom event and you are not listing to the `events` of the sidebar here. You are listening to all of them.

So inside, you want to filter to events from the `sidebar`.
Something like:
```js
sidebar.addEventListener('hide.bs.collapse', function (e) {
  if (e.target && e.target.id === 'sidebar-wrapper') {
    document.getElementById('main-container').classList.add('toggled');
    console.log(e)
  }
})
```

Also, nitpick but use [arrow functions](https://www.w3schools.com/js/js_arrow_function.asp). Browsers support these in today's world.

docs/_template.html
```diff
@@ -2,84 +2,102 @@
...
-        <div id="header">
-            <div class="searchbox" id="fsdocs-searchbox">
-                <label for="search-by">
-                    <i class="fas fa-search"></i>
-                </label>
-                <input data-search-input="" id="search-by" type="search" placeholder="Search..."/>
-                <span data-search-clear="">
-                        <i class="fas fa-times"></i>
-                    </span>
+  <main id="main-container">
```
nojaf, line 26: You probably want to flex things a little bit different on mobile.

image

I think there it makes more sense that the menu is over the main content.

docs/_template.html
```diff
@@ -2,84 +2,102 @@
...
+                <input aria-label="Search" class="form-control  border-start-0 border" data-search-input=""
+                  id="search-by" type="search" placeholder="Search..." />
+                <span class="input-group-append">
+                </span>
+                <span data-search-clear="">
+                  <i class="fas fa-times"></i>
+                </span>
+              </div>
+            </form>
+            <a class="nav-link m-0 p-0" href="https://github.com/fsprojects/fantomas">GitHub<i
```
nojaf, line 65: Use `target="_blank"` here, we want to keep the user on the documentation.
Same for the `tool` link.

nojaf, line 65: I think you can use `{{fsdocs-repository-link}}` instead of the hardcoded link.

docs/_template.html
```diff
@@ -3,32 +3,31 @@
...
   {{fsdocs-watch-script}}
 </head>
 
 <body>
-  <main>
-    <div id="sidebar-wrapper" class="bg-primary h-100" >
+  <main >
+    <div id="sidebar-wrapper" class="bg-primary h-100 collapse" >
+      <div id="sidebar-logo" class="d-flex justify-content-center p-5">
       <a href="{{fsdocs-logo-link}}"><img id="fsdocs-logo" src="{{root}}/{{fsdocs-logo-src}}" /></a>
```
nojaf, line 29: This links back to the homepage (onepager), maybe it should be `{{fsdocs-logo-link}}/docs` instead.

nojaf (changes requested):
I'm really happy where this is heading.
Keep it up Alan!

nojaf:
The top menu bar should also scroll down when the page scrolls.

image

`position: sticky` might do the trick there as well.

nojaf:
Looks good @yisusalanpng! I would make the `content` div (`.doc-content .content-wrapper .content`) the same white used in the navbar.

image

---

I would also add a min-height of `100vh` to the `.content-wrapper` div,

image

That way the grey will always go to the end of the page.

---

Maybe we should also somewhat centre the content of the sidebar:

image

---

And on mobile, I think I would drop the padding and box-shadow of 
image

nojaf (changes requested):
Let's complete these last remarks together and merge it in 🎉.

nojaf (approved):
Well done!

Issue #2330, opened 2022-07-01
List what specific features we use from the F# language


nojaf:
When we asked people in the past if they want to fix their own issue, we something got the reply: "No, because I'm new to F#".

That actually shouldn't be a showstopper. We should list the language features that we often use and try to give some pointers for those newcomers on how to get up to speed.

A couple of topics:
- Custom operators: https://docs.microsoft.com/en-us/dotnet/fsharp/language-reference/operator-overloading#creating-new-operators
- Active patterns: https://docs.microsoft.com/en-us/dotnet/fsharp/language-reference/active-patterns
- Signature files: https://docs.microsoft.com/en-us/dotnet/fsharp/language-reference/signature-files

nojaf:
This has been addressed, see https://fsprojects.github.io/fantomas/docs/contributors/FSharp.html

PR #2379, opened 2022-07-20
Add initial F# information.

nojaf:
Related to https://github.com/fsprojects/fantomas/issues/2330

### DIFF
--- a/docs/docs/contributors/FSharp.md
+++ b/docs/docs/contributors/FSharp.md
@@ -1,0 +1,126 @@
+---
+category: Contributors
+categoryindex: 2
+index: 2
+---
+# F#
+
+## New to F#?
+
+If you are truly brand-new to the F# language, you might want to start by reading the [F# documentation of Microsoft](https://dotnet.microsoft.com/en-us/languages/fsharp).  
+Some other great resources (in no particular order) are:
+
+- [Essential F#](https://leanpub.com/essential-fsharp)
+- [F# for Fun and Profit](https://fsharpforfunandprofit.com/)
+- [F# Fundamentals Tutorial | Learn Functional Programming | Step-by-Step Guide](https://www.youtube.com/watch?v=SvOInBxPL30)
+- [F# Foundation Slack](https://fsharp.org/guides/slack/)
+- [F# on Discord](https://discord.com/invite/R6n7c54)
+
+## Used F# features
+
+F# has a lot of nice language features, although not all of them are used in Fantomas.
+We wish to highlight the most important ones that we use before continuing:
+
+- [Partial active pattern](https://docs.microsoft.com/en-us/dotnet/fsharp/language-reference/active-patterns#partial-active-patterns), these heavily used in `SourceParser.fs`.
+  In short, we use the `Untyped Abstract Syntax Tree` created by the F# parser, we don't use all the information in that tree to restore the source code.
+
+For example [SynExpr.For](https://fsharp.github.io/fsharp-compiler-docs/reference/fsharp-compiler-syntax-synexpr.html#For), the definition looks like:
+
+```fsharp
+type SynExpr =
+    ...
+    /// F# syntax: 'for i = ... to ... do ...'
+    | For of
+        forDebugPoint: DebugPointAtFor *
+        toDebugPoint: DebugPointAtInOrTo *
+        ident: Ident *
+        equalsRange: range option *
+        identBody: SynExpr *
+        direction: bool *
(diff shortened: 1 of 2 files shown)


docs/docs/contributors/FSharp.md
```diff
@@ -0,0 +1,128 @@
...
+- [F# Fundamentals Tutorial | Learn Functional Programming | Step-by-Step Guide](https://www.youtube.com/watch?v=SvOInBxPL30)
+- [F# Foundation Slack](https://fsharp.org/guides/slack/)
+- [F# on Discord](https://discord.com/invite/R6n7c54)
+
+## Used F# features
+
+F# has a lot of nice language features, although not all of them are used in Fantomas.
+We wish to highlight the most important ones that we use before continuing:
+
+- [Partial active pattern](https://docs.microsoft.com/en-us/dotnet/fsharp/language-reference/active-patterns#partial-active-patterns), these heavily used in `SourceParser.fs`.
```
dawedawe, line 24:
```suggestion
- [Partial active patterns](https://docs.microsoft.com/en-us/dotnet/fsharp/language-reference/active-patterns#partial-active-patterns), these are heavily used in `SourceParser.fs`.
```

docs/docs/contributors/FSharp.md
```diff
@@ -0,0 +1,128 @@
...
+        +> sepNln
+        +> genExpr astContext e3
+        +> unindent
+    )
+| _ -> ...
+```
+
+- [Custom operators](https://docs.microsoft.com/en-us/dotnet/fsharp/language-reference/operator-overloading#creating-new-operators). In F# there are some special operators like [|>](https://fsharp.github.io/fsharp-core-docs/reference/fsharp-core-operators.html#(|%3E)) and [>>](https://fsharp.github.io/fsharp-core-docs/reference/fsharp-core-operators.html#(%3E%3E)).  
+
+Note that these are just function themselves as well. Instead of specifying all the arguments after the function name, (infix) operators let you specify an argument before the operator and after.
```
dawedawe, line 79:
```suggestion
Note that these are just functions themselves as well. Instead of specifying all the arguments after the function name, (infix) operators let you specify an argument before the operator and after. 
```

docs/docs/contributors/FSharp.md
```diff
@@ -0,0 +1,128 @@
...
+
+Note that these are just function themselves as well. Instead of specifying all the arguments after the function name, (infix) operators let you specify an argument before the operator and after. 
+
+In F#, you are able to create your own operators as well. In Fantomas, the most notable are `!-` and `+>`. We will cover them later, but if you peek in `CodePrinter.fs`, they are heavily used there.
+
+- [Signature files](https://docs.microsoft.com/en-us/dotnet/fsharp/language-reference/signature-files).
+
+In Fantomas we use signature files to define the module boundaries. Everything that is both defined in the implementation file (the `*.fs` file) and in the signature file (the `*.fsi` file) is considered to be visible to other modules.
+If a signature file is present, there is no need to specify `private` in a function you don't want to be visible to other modules. Just don't add `val` entry to the signature file and it will be private automatically.
```
dawedawe, line 86:
```suggestion
If a signature file is present, there is no need to specify `private` in a function you don't want to be visible to other modules. Just don't add a `val` entry to the signature file and it will be private automatically.
```

docs/docs/contributors/FSharp.md
```diff
@@ -0,0 +1,128 @@
...
+              yield! Option.toList memberFlags.Trivia.StaticRange ]
+```
+
+The type `SynMemberFlags` does not expose any range information, but we can extend it to do so.  
+The `.FullRange` naming convention is used to indicate that we are not satisfied by the original range or it is lacking all together.  
+Don't worry just yet about this implementation, so keep in mind that with this feature we can later use `memberFlags.FullRange` on a `SynMemberFlags` instance.
+
+- [Function Values](https://docs.microsoft.com/en-us/dotnet/fsharp/language-reference/functions/#function-values)
+
+This is well-known concept in F# and for completion sake we to mention this. In F#, you can pass a function as an argument to another function.
```
dawedawe, line 113:
```suggestion
This is well-known concept in F# and for completion sake we do mention this. In F#, you can pass a function as an argument to another function.
```

dawedawe (approved):
Feels already preety complete to me. Some details about the custom operators can be added later on.

nojaf:
Thanks for the review @dawedawe!

PR #2384, opened 2022-07-25
Extend InputPath DU to improve error messages

### LINKED ISSUE #2341: Show a better error message when the folder does not exist
When you run `dotnet fantomas` right now, you get the `Input path is missing...` message.
Is that a good message or should the help be shown instead?

Next, if you run `dotnet fantomas foo` where the folder `foo` doesn't exist, you get that same `Input path is missing...` message. This could be improved by `No file or folder "foo" found on disk` or something similar.

I recall an example where a path with Windows slashes was used on a Linux VM using the `--check` flag and it wasn't all that obvious that nothing happened.
The exit code is also `0` when you enter `dotnet fantomas foo --check ; $LASTEXITCODE` (in PowerShell).
This should be `1` as it is without the `--check` flag.

dawedawe:
Fixes #2341 

Instead of showing the complete help in case of InputPath.Unspecified, a hint is shown.
Of course, it's subjective, but I somehow prefer CLI tools which don't slap the whole help in my face without me asking for it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## [Unreleased]
+
+### Changed
+* Show a better error message when the folder does not exist. [#2341](https://github.com/fsprojects/fantomas/issues/2341)
+
 ## [5.0.0-beta-002] - 2022-07-19
 
 ### Fixed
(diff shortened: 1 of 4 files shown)


nojaf (approved):
Looks good, thank you!

Issue #2411, opened 2022-08-06
Underscores in subsections of configuration docs disappear, plus suggestion for table of contents


abelbraaksma:
First of all, great work on Fantomas 5!!! I'm going over the changelog of each alpha and beta (didn't find an all-encompassing list somewhere) and I really like what I find!

I found the docs (it seems like they aren't directly linked from the `readme.md`, I wouldn't mind adding them there), I think these are the ones: https://fsprojects.github.io/fantomas/docs/Configuration.html

If you look at the "Configuration" page, it renders the captions wrongly, because of the `_` in the names of the items. I was trying to search for info on certain settings, but that doesn't work because of the rendering:

![image](https://user-images.githubusercontent.com/16015770/183227399-79bed1bb-3807-4f53-8fdb-c99c65ad8559.png)

Or this:

![image](https://user-images.githubusercontent.com/16015770/183227552-4a6127c7-c386-4ae2-871a-991a504c7c0c.png)

PS: just an idea, but perhaps we can make an TOC of the editor options on top, that way searching through the page won't be so necessary anymore.

PPS: the same file on Github renders just fine: https://github.com/fsprojects/fantomas/blob/master/docs/docs/Configuration.md

nojaf:
Hello,

> First of all, great work on Fantomas 5

Thank you 😊.

> I found the docs
The documentation is still very much [WIP](https://github.com/fsprojects/fantomas/issues?q=is%3Aissue+is%3Aopen+label%3Adocumentation), once we made more progress, the idea is to replace the current `README.md` with a link to the docs website.

> it renders the captions wrongly

Yes, this is a bug in fsdocs (see https://github.com/fsprojects/FSharp.Formatting/issues/389).
Feel free to take a stab at that one 😅.

> TOC of the editor

That is a good idea, I would accept a PR for that.

abelbraaksma:
> That is a good idea, I would accept a PR for that.

Done 😃. See #2414.

> Yes, this is a bug in fsdocs (see https://github.com/fsprojects/FSharp.Formatting/issues/389).

I tried to circumvent the bug. Not 100% it works as I don't know how your CI renders that page.

nojaf:
> I don't know how your CI renders that page

https://github.com/fsprojects/fantomas/blob/8b608f590115d11c0cad599c1df8fdaee4db1ffc/docs/.README.md

nojaf:
Due to the upstream fix of FSharp.Formatting, this is no longer a problem.

PR #2414, opened 2022-08-08
Add a table of contents and fix disappearing underscores in subsections

abelbraaksma:
Fixes: #2411 

As in the title. This adds a ToC to the `Configuration.md`, which is the core reference of all things configuration. I struggled a little with what level the subsections should be at, I ended up making the 2nd level, as opposed to the original 3rd level, as that aligned better with the ToC. Hope that's ok.

This also fixes the issue with underscores in titles being rendered improperly (actually: I **hope** it fixes them, I replaces the underscores with an escaped underscore, which is still legal MD, but hope that `FSharp.Formatting` correctly knows what to do with it). See also: https://github.com/fsprojects/FSharp.Formatting/issues/389.

Note that, because of the way that MD auto-creates anchors of subsections and these anchors contain illegal characters for IDs, I had to add the anchors manually. I did this automated, but for new sections being added, this pattern should be repeated and the ToC should be re-updated.

### DIFF
--- a/docs/docs/Configuration.md
+++ b/docs/docs/Configuration.md
@@ -8,7 +8,58 @@ index: 5
 Fantomas ships with a series of format options.
 These can be stored in an [.editorconfig](https://editorconfig.org/) file and will be picked up automatically by the commandline tool.
 
-A default .editorconfig file would look like
+<!-- 
+    This index can be auto-generated with VS Code's Markdown All in One extension.
+    The ToC will be updated-on-save, or can be generated on command by using
+    Ctrl-Shift-P: "Create table of contents". 
+    More info: https://marketplace.visualstudio.com/items?itemName=yzhang.markdown-all-in-one#table-of-contents 
+-->
+## Index<!-- omit in toc -->
+- [Configuration](#configuration)
+- [Default `.editorconfig` settings](#default-editorconfig-settings)
+- [Each setting explained](#each-setting-explained)
+  - [indent\_size](#indent_size)
+  - [max\_line\_length](#max_line_length)
+  - [end\_of\_line](#end_of_line)
+  - [insert\_final\_newline](#insert_final_newline)
+  - [fsharp\_space\_before\_parameter](#fsharp_space_before_parameter)
+  - [fsharp\_space\_before\_lowercase\_invocation](#fsharp_space_before_lowercase_invocation)
+  - [fsharp\_space\_before\_uppercase\_invocation](#fsharp_space_before_uppercase_invocation)
+  - [fsharp\_space\_before\_class\_constructor](#fsharp_space_before_class_constructor)
+  - [fsharp\_space\_before\_member](#fsharp_space_before_member)
+  - [fsharp\_space\_before\_colon](#fsharp_space_before_colon)
+  - [fsharp\_space\_after\_comma](#fsharp_space_after_comma)
+  - [fsharp\_space\_before\_semicolon](#fsharp_space_before_semicolon)
+  - [fsharp\_space\_after\_semicolon](#fsharp_space_after_semicolon)
+  - [fsharp\_space\_around\_delimiter](#fsharp_space_around_delimiter)
+  - [fsharp\_max\_if\_then\_short\_width](#fsharp_max_if_then_short_width)
+  - [fsharp\_max\_if\_then\_else\_short\_width](#fsharp_max_if_then_else_short_width)
+  - [fsharp\_max\_infix\_operator\_expression](#fsharp_max_infix_operator_expression)
+  - [fsharp\_max\_record\_width](#fsharp_max_record_width)
+  - [fsharp\_max\_record\_number\_of\_items](#fsharp_max_record_number_of_items)
+  - [fsharp\_record\_multiline\_formatter](#fsharp_record_multiline_formatter)
+  - [fsharp\_max\_array\_or\_list\_width](#fsharp_max_array_or_list_width)
+  - [fsharp\_max\_array\_or\_list\_number\_of\_items](#fsharp_max_array_or_list_number_of_items)
+  - [fsharp\_array\_or\_list\_multiline\_formatter](#fsharp_array_or_list_multiline_formatter)
+  - [fsharp\_max\_value\_binding\_width](#fsharp_max_value_binding_width)
+  - [fsharp\_max\_function\_binding\_width](#fsharp_max_function_binding_width)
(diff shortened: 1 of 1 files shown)


docs/docs/Configuration.md
```diff
@@ -8,7 +8,58 @@ index: 5
...
 
-A default .editorconfig file would look like
+<!-- 
+    This index can be auto-generated with VS Code's Markdown All in One extension.
+    The ToC will be updated-on-save, or can be generated on command by using
+    Ctrl-Shift-P: "Create table of contents". 
+    More info: https://marketplace.visualstudio.com/items?itemName=yzhang.markdown-all-in-one#table-of-contents 
+-->
+## Index<!-- omit in toc -->
+- [Configuration](#configuration)
```
nojaf, line 18: This anchor is wrong, should be `#Configuration`.

docs/docs/Configuration.md
```diff
@@ -8,7 +8,58 @@ index: 5
...
+- [Default `.editorconfig` settings](#default-editorconfig-settings)
```
nojaf, line 19: It should be `#Default-settings` please verify all these links locally.

docs/docs/Configuration.md
```diff
@@ -53,9 +104,12 @@ fsharp_strict_mode=false
 Please note that you should only add settings to the `.editorconfig` file when you want to deviate from the default settings.
 Copying the entire list above is unnecessary.
 
-### indent_size
+# Each setting explained
```
nojaf, line 107: I'm not really a fan of having multiple h1 headers on the same page.
We don't do this on the other pages, the h1 should be the title of the page and only occur once.
Please address.

docs/docs/Configuration.md
```diff
@@ -53,9 +104,12 @@ fsharp_strict_mode=false
...
+
+<a id="indent_size"></a>
+## indent\_size
```
nojaf, line 110: I'm on the fence about this one. Escaping work and appears to be legal Markdown, however, it is a missed opportunity to tackle https://github.com/fsprojects/FSharp.Formatting/issues/389.

docs/docs/Configuration.md
```diff
@@ -8,7 +8,58 @@ index: 5
 Fantomas ships with a series of format options.
 These can be stored in an [.editorconfig](https://editorconfig.org/) file and will be picked up automatically by the commandline tool.
 
-A default .editorconfig file would look like
+<!-- 
+    This index can be auto-generated with VS Code's Markdown All in One extension.
+    The ToC will be updated-on-save, or can be generated on command by using
+    Ctrl-Shift-P: "Create table of contents".
```
nojaf, line 14: Not really a fan of using a format because of the convenience of a certain plugin in a certain editor. My editor (Rider) does something different, so this is a bit unfortunate there is no real standard.

nojaf (changes requested):
Hey, thank you for taking the time to propose this TOC.

Looking at the rendered result I'm not sure how big of an improvement this is.
image

There is a long list of items and I feel when you are looking for something specific you would `ctrl + f` your way to it because of the number of items in the list.

Perhaps it is better to make the search box work when you enter a setting name to get to the right location.

docs/docs/Configuration.md
abelbraaksma: Actually, it's `#default-editorconfig-settings`, I did verify, see: https://github.com/fsprojects/fantomas/blob/b11b645a997a0baa2dea51a7c1aba570ce9dceca/docs/docs/Configuration.md#default-editorconfig-settings

docs/docs/Configuration.md
abelbraaksma: Actually, it's `#configuration`, see https://github.com/fsprojects/fantomas/blob/b11b645a997a0baa2dea51a7c1aba570ce9dceca/docs/docs/Configuration.md#configuration

docs/docs/Configuration.md
abelbraaksma: @nojaf, yes, that was my main point here. I didn't really like that either, but it originally rendered better with the ToC. I can remove `Configuration` from the ToC, but as you mentioned above, maybe there won't be a ToC at all. So either way, I'll resolve.

docs/docs/Configuration.md
abelbraaksma: @nojaf I had a quick look at that and that should be solved, sure, but I lack the time and don't know enough (yet) of that lib. Also, if I'm not going to fix it, the bug has been out there for 6 years... hence this approach.

docs/docs/Configuration.md
abelbraaksma: @nojaf I agree. But yeah, there's no standard. And of course, other plugins can be used. I'll remove the comment.

abelbraaksma:
> Looking at the rendered result I'm not sure how big of an improvement this is.

Is there a trivial way to see the rendered result from the repo? I just checked "View file", which showed the links working, but I have the feeling that `FSharp.Format` uses a different rendering of certain anchors. The simple solution would be to just add specific anchors to each title.

> Perhaps it is better to make the search box work when you enter a setting name to get to the right location.

@nojaf  That could work. Or a collapsible ToC. Not sure there's any (standard?) way of doing either, though, as it should be on the page. I do have to say, while working on it, it was nice to be able to click back and forth between the items. The page is rather long, and the ToC is a little bit more readable as a list of available options than the `defaultConfig` section.

abelbraaksma:
Yes, and this PR was the trigger, apparently, for @nojaf to fix it. Which is great!

abelbraaksma:
@nojaf, thanks for that fix, and glad it got merged so quickly! I guess I can undo most changes (glad so, it’s better with proper support!). Would still like to have the ToC in, but if you think it’s to intrusive, we can probably close this with no action.

docs/docs/Configuration.md
abelbraaksma: I’d want to, I just didn’t know how (at least not trivially). Looks like fsdocs does not read the added anchors. They were there to prevent ambiguity in the auto creation of header anchors, which apparently each markdown reader does differently.

So, the code explicitly has the lowercase `c` here, fsdocs just totally ignores it…

nojaf:
Hello @abelbraaksma, following up on this, we would go for a slightly different approach in https://github.com/fsprojects/fantomas/pull/2427.
Thanks again for triggering progress with this PR.

abelbraaksma:
@nojaf no worries, glad I triggered the discussion:)

Issue #2320, opened 2022-07-01
Add contributors documentation


nojaf:
We should have documentation that explains the high-level overview of this project.
And add some detailed information about the code. The high-level stuff is somewhat captured in our [Contribution guidelines](https://github.com/fsprojects/fantomas/blob/master/CONTRIBUTING.md). It should be split up and expanded upon.

### High-level topics

Somewhat in order.

- A first local build
- Project structure (what does each project do)
- Overview of FAKE build
- Parsing of the AST
- Trivia and ASTTransformer
- CodePrinter
- WriteEvents in Context
- Format selection
- Fantomas Daemon and Fantomas.Client
- Formatting with conditional compiler defines
- Online tools and how to use them.
- PR checklist.

### Details

These are in no particular order.

- Explain `Context -> Context`
- Custom operators: `!-`, `+>`
- Helper functions: `col`, `expressionFitsOnRestOfLine`, `isShortExpression`, `sepNln`, `sepSpace`, `onlyIf`, `genTriviaFor`, `indent`, `atCurrentColumn`, `colWithNlnWhenItemIsMultiline`, `leadingExpressionIsMultiline`.
- Explain `Queue` data structure
- Why `indent` only works after `sepNln`
- What `ColMultilineItem` is.
- The `Define` module and how we resolve the different AST trees.
- `AstExtensions` and `.FullRange` extension function.

### Questions that should be answered

Again no particular order.

- Why do we have our own F# compiler API?
- Why all the active patterns in SourceParser?
- Why we restore strings/chars/numbers using `ISourceText.GetContentAt`?
- Why could a code comment disappear after formatting?
- Why is CodePrinter so large and full of recursive functions?

nojaf:
Close enough, https://fsprojects.github.io/fantomas/docs/contributors/Index.html

PR #2415, opened 2022-08-08
Add page about Conditional Compilation Directives.

nojaf:
Related to #2320.

### DIFF
--- a/docs/docs/contributors/Conditional Compilation Directives.md
+++ b/docs/docs/contributors/Conditional Compilation Directives.md
@@ -91,7 +91,7 @@ This tree is almost identical but the constant value is now `Const (Int32 0, ...
 
 As the combination of directives has an influence on the tree, Fantomas first parses the tree without any directives.
 This base tree is then being inspected for [ConditionalDirectiveTrivia](https://fsharp.github.io/fsharp-compiler-docs/reference/fsharp-compiler-syntaxtrivia-conditionaldirectivetrivia.html).  
-We determine the different combinations the `Defines` module.
+We determine the different combinations in the `Defines` module.
 
 <div class="mermaid text-center">
 graph TD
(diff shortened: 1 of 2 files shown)


docs/docs/contributors/Conditional Compilation Directives.md
```diff
@@ -0,0 +1,155 @@
...
+        CodeComments = [] }))
+```
+
+This tree is almost identical but the constant value is now `Const (Int32 0, ...)`.
+
+## Multiple trees
+
+As the combination of directives has an influence on the tree, Fantomas first parses the tree without any directives.
+This base tree is then being inspected for [ConditionalDirectiveTrivia](https://fsharp.github.io/fsharp-compiler-docs/reference/fsharp-compiler-syntaxtrivia-conditionaldirectivetrivia.html).  
+We determine the different combinations the `Defines` module.
```
dawedawe, line 94:
```suggestion
We determine the different combinations in the `Defines` module.
```

docs/docs/contributors/Conditional Compilation Directives.md
```diff
@@ -0,0 +1,155 @@
...
+    0
+    #else
+    #endif
+```
+
+## Merging the trees
+
+Once every tree is formatted, we chop each file into fragments.
+A fragment is everything between a conditional directive `#if | #else | #endif` or an actual directive.  
+Each result should have the same amount of fragments before we can merged them together.
```
dawedawe, line 134:
```suggestion
Each result should have the same amount of fragments before we can merge them together.
```

docs/docs/contributors/Conditional Compilation Directives.md
```diff
@@ -0,0 +1,155 @@
...
+If this is not the case, it means that somewhere a trivia was not properly restored.
+
+If the number of fragments add up in each tree, then we merge two trees by reducing both list and comparing each fragment.
```
dawedawe, line 137:
```suggestion
If the number of fragments add up in each tree, then we merge two trees by reducing both lists and comparing each fragment.  
```

dawedawe (approved):
Another useful piece of information :)

PR #2251, opened 2022-05-13
Rename ragnarok

nojaf:
Related to https://github.com/fsprojects/fantomas/issues/1408.
Renames `fsharp_ragnarok` to `fsharp_experimental_stroupstrup_style`.
Similarly, `fsharp_keep_indent_in_branch` was renamed to `fsharp_experimental_keep_indent_in_branch` to further indicate that the feature is not stable.

PR #2437, opened 2022-08-18
Update leftover references to ragnarok

josh-degraw:
Not sure if these were intentionally left out of #2251, but I figured for the sake of code clarity these references to the ragnarok code name should probably be updated to actually refer to the stroustrup feature.


nojaf (approved):
Yup, totally missed all those! Thanks

Issue #1570, opened 2021-04-01
Comment is unindented with MultilineBlockBracketsOnSameColumn


Smaug123:
[...]

PR #2455, opened 2022-08-29
Pick a better trivia node for line comment.

nojaf:
A faint attempt to solve https://github.com/fsprojects/fantomas/issues/1570.
This wouldn't magically fix any similar bug with code comments shifting, but it might be a step in the right direction.

@jindraivanek I have some questions about the `WriterModel`.
@dawedawe you might find this one interesting as well.


src/Fantomas.Core/CodePrinter.fs
```diff
@@ -2658,7 +2658,16 @@ and genMultiLineArrayOrListAlignBrackets (isArray: bool) xs openingTokenRange cl
         +> indent
         +> sepNlnUnlessLastEventIsNewline
         +> col sepNln xs (genExpr astContext)
-        +> unindent
+        +> (fun (ctx: Context) ->
+            match Queue.tryHead ctx.WriterEvents with
```
nojaf, line 2662: The last `expr` in `xs` had a trivia after in this case.
Leading the last event to be a `WriteLineBecauseOfTrivia`.
This of course messes with the `unindent` call later on, as indentation changes only kick in on the next line.

In this function, I reset the last event and play it again, where I'm adding `unindent` before the  `sepNlnForTrivia`.

Is this in principle somewhat ok to do? If so, we probably want to extract a helper in `Context`.

dawedawe:
> A faint attempt to solve #1570. This wouldn't magically fix any similar bug with code comments shifting, but it might be a step in the right direction.
> 
> @jindraivanek I have some questions about the `WriterModel`. @dawedawe you might find this one interesting as well.

I sure do, thanks for the heads up.

src/Fantomas.Core.Tests/CommentTests.fs
```diff
@@ -2446,3 +2446,51 @@ module A =
...
+        { config with MultilineBlockBracketsOnSameColumn = true }
+    |> prepend newline
+    |> should
+        equal
+        """
+let v =
+    [
+        "thing"
+        "thing2"
+        // no other things
```
nojaf, line 2494: In order for this to work with newlines before or after the comment, they will probably need to be bundled as one trivia group.

src/Fantomas.Core/CodePrinter.fs
jindraivanek: > Is this in principle somewhat ok to do @jindraivanek? If so, we probably want to extract a helper in `Context`.

In general, not exactly, there is more to it than changing `Lines`, see https://github.com/fsprojects/fantomas/blob/master/src/Fantomas.Core/Context.fs#L83 Maybe it doesn't matter in this case, but not sure about that.

Correct way to do it would be to replay changed WriterEvents from some context before `WriteLineBecauseOfTrivia` is added. (As we can't easily reverse events.) I can later describe in more details of what I have in mind, if needed.

nojaf: I guess the main bit that could be off in the current approach is `let currentLine = String.Concat(List.head m.Lines, m.WriteBeforeNewline).TrimEnd()`.

Though getting the context before `WriteLineBecauseOfTrivia` is difficult as it will be in the last `genExpr astContext`. Please elaborate on the `more details` bit, I might need a hint for this one.

jindraivanek: Something like this:

```fsharp
        genTriviaFor SynExpr_ArrayOrList_OpeningDelimiter openingTokenRange sepOpenLFixed
        +> indent
        +> sepNlnUnlessLastEventIsNewline
        +> (fun (ctx: Context) ->
            let ctx2 = (col sepNln xs (genExpr astContext)) ctx
            let newEvs = ctx2.WriterEvents |> Seq.take (ctx2.WriterEvents.Length - ctx.WriterEvents.Length)
            match Seq.tryHead newEvs with
            | Some WriterEvent.WriteLineBecauseOfTrivia ->
                // apply new events on ctx, except the last one
                let resetModel = 
                    (ctx.WriterModel, Seq.rev(Seq.tail newEvs))
                    ||> Seq.fold (fun m e -> WriterModel.update ctx.Config.MaxLineLength e m)
                let resetContext =
                    { ctx2 with
                        WriterEvents = ctx2.WriterEvents.Tail
                        WriterModel = resetModel }
                (unindent +> sepNlnForTrivia) resetContext
            | _ -> unindent ctx2)
        +> sepNlnUnlessLastEventIsNewline
        +> genTriviaFor SynExpr_ArrayOrList_ClosingDelimiter closingTokenRange sepCloseLFixed
```

this code probably don't work (didn't try it), but hopefully main idea is self-evident.

nojaf: Oh, I see now what you mean. Good thinking, this will indeed be the safest thing to do.

src/Fantomas.Core/Context.fs
```diff
@@ -1295,3 +1295,28 @@ let colWithNlnWhenItemIsMultilineUsingConfig (items: ColMultilineItem list) (ctx
         colWithNlnWhenItemIsMultiline items ctx
     else
         col sepNln items (fun (ColMultilineItem (expr, _)) -> expr) ctx
+
```
nojaf, line 1298: @jindraivanek thanks again for the tip, I came up with this solution.
Let me know if you see any room for improvement.

src/Fantomas.Core.Tests/CommentTests.fs
```diff
@@ -2481,3 +2481,81 @@ The same indentation is ensured to be consistent in a source file.
...
+let ``how adding one blank line really complicates things`` () =
+    formatSourceString
+        false
+        """
+let v =
+    [|
+        "thing"
+        "thing2"
+
+        // there is a blank line above this comment
```
nojaf, line 2546: This required the `Newline` and `CommentOnSingleLine` to be assigned together to a fitting node.
That group requires knowledge of the start column of the comment in this case.
I don't think the grouping concept is a bad one, it does ripple through a lot of the existing code. I think some more aspects will need to be revisited before this can be finalized.

src/Fantomas.Core/paket.references
```diff
@@ -1,3 +1,4 @@
 FSharp.Core
+System.Collections.Immutable
```
nojaf, line 2: I'm a bit conflicted about this one. The ImmutableQueue makes sense and is nice, but at the same time, it does introduce an additional dependency.

nojaf:
Meh

PR #2465, opened 2022-08-30
Strip only \r to be able to detect a changed number of newlines

### LINKED ISSUE #2461: --check should fail if the number of newlines is different
When using the `--check` to verify if a file has been formatted correctly, we completely skip all newlines.
The original reasoning for this is that we don't want to fail the check if the newlines are different (`lf` versus `crlf`).

However, given the setting [fsharp_keep_max_number_of_blank_lines](https://fsprojects.github.io/fantomas/docs/end-users/Configuration.html#fantomas-setting-icon-type-green-fantomas-setting-icon-fsharp_keep_max_number_of_blank_lines), the number of newlines can be impacted by this setting. Thus, the check should take the number of newlines into account.

It is probably sufficient to only strip the `\r` characters in
https://github.com/fsprojects/fantomas/blob/7920c43c21739d0b0933abf31495b5411b41fe81/src/Fantomas/Format.fs#L56-L57

dawedawe:
Fixes #2461

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## [Unreleased]
+
+### Fixed
+* --check should fail if the number of newlines is different. [#2461](https://github.com/fsprojects/fantomas/issues/2461)
+
 ## [5.0.0-beta-008] - 2022-08-30
 
 ### Fixed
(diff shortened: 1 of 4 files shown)


nojaf (approved):
Thanks!

PR #2473, opened 2022-09-02
Menu items with boostrap accordion and opening depending on sections.

alanlomeli:
Changed sidebar menu items to use an accordion, and implemented a script to open the menu depending on which section the user is.

  
Note that the file _docs/docs/end-users/VisualStudio.md_ is showing as changed but I can't find any visible change.


docs/_template.html
```diff
@@ -103,8 +103,26 @@
   else{
     document.getElementById(e.target.id+'-wrapper').classList.add('menu-open');
   }
-})
-});
+      })
```
nojaf, line 106: It might be a good time to put all this JS in a separate script.
It is becoming lengthy.

docs/_template.html
```diff
@@ -103,8 +103,26 @@
...
+      let fsdocsFileName = `{{fsdocs-source-filename}}`
```
nojaf, line 107: Always use `const` instead of `let` when possible.

docs/_template.html
```diff
@@ -103,8 +103,26 @@
...
+      let fsdocsSection = fsdocsFileName.substring(fsdocsFileName.indexOf('/') + 1, fsdocsFileName.lastIndexOf('/'))
```
nojaf, line 108: This probably won't work for API doc pages like https://github.com/fsprojects/fantomas/pull/2473/files

nojaf (changes requested):
I think this is an elegant viable solution.
On Chrome I do have a weird extra bullet for each category:

image

And for the "API References" seems to be always open. Maybe you did this by design?

nojaf (approved):
Nice!

PR #2478, opened 2022-09-06
Documentation fixes for mobile

alanlomeli:
- Hide sidebar by default when on mobile
- Fix some overflows with some titles and codesnippets

### DIFF
--- a/docs/.style/fsdocs-custom.sass
+++ b/docs/.style/fsdocs-custom.sass
@@ -5,6 +5,8 @@ $font-family-sans-serif: 'Poppins', sans-serif
 @import "./fsdocs-default"
 @import "./sidebar.sass"
 
+body
+  display: grid
 .sidebar-toggle
   opacity: 0.5
 
--- a/docs/.style/fsdocs-custom.sass
+++ b/docs/.style/fsdocs-custom.sass
@@ -42,6 +42,9 @@ $font-family-sans-serif: 'Poppins', sans-serif
   box-shadow: 0 4px 10px $drop-shadow
   table.pre
     background-color: $pearl
+    display: block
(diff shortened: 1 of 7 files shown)


nojaf:
Good fix on the codesnippets!

I've made some more drastic changes to the mobile menu:
- You really don't want a flicker on your mobile phone. So I tried to hide the menu by default and then show it once you are on the desktop.
- The `collapse` control we currently use doesn't have any support for responsiveness, so your attempt with JavaScript was the only was. But, alas, in either desktop or mobile that would lead to a flicker.
- I then tried to use display helper classes to just hide the menu on mobile and show it on desktop (`<= md`).
- As removing the `collapse`, a lot of the bootstrap stuff did no longer apply, so I remove most of it and fell back on two basic button click handlers. One to open the menu, one to close it.
- A problem would then occur when you go from mobile to desktop and start toggling the menu.
But then it hit me: when would a user on a desktop really close the menu? So, for now, or maybe forever, I removed the toggle button on the desktop.

Please try these changes out locally and let me know what you think about them.
We lost the fancy animation but gained a fast solution that will always work without any flicker.

nojaf:
@alanlomeli another easy way to maybe deal with this would be to use `display: grid`.
It would require some restructuring of the HTML but might also be quite simple.
Example fiddle: https://jsfiddle.net/afk0cux1/14/

nojaf:
In case you are less familiar: https://css-tricks.com/snippets/css/complete-guide-grid/

nojaf (approved):
Hi @alanlomeli, I've finished the menu mechanism.
I've extracted it from the styling and written some comments.
I think we can ship this 🎉

PR #2495, opened 2022-09-09
bugfix/issue-2485-sequentials-stack-overflow

njlr:
* Switches Sequentials to an imperative impl for efficiency and to not blow up stack


src/Fantomas.Core/SourceParser.fs
```diff
@@ -608,10 +608,26 @@ let (|Sequential|_|) =
     | SynExpr.Sequential (_, isSeq, e1, e2, _) -> Some(e1, e2, isSeq)
     | _ -> None
 
-let rec (|Sequentials|_|) =
+let (|Sequentials|_|) =
```
nojaf, line 611: We have solved this type of problem before in the codebase, this is quite the departure from that.

To make it tail recursive you can use the following:

```fsharp
let (|Sequentials|_|) e =
    let rec visit (e: SynExpr) (finalContinuation: SynExpr list -> SynExpr list) : SynExpr list =
        match e with
        | Sequential (e1, e2, _) -> visit e2 (fun xs -> e1 :: xs |> finalContinuation)
        | e -> finalContinuation [ e ]

    match e with
    | Sequential (e1, e2, _) ->
        let xs = visit e2 id
        Some(e1 :: xs)
    | _ -> None
```

nojaf (changes requested):
Hello, thank you for taking the initiative to solve this problem.

I'm going to sound very purist here but I'm not a fan of the solution.
We have solved these problems in the past via tail recursion, I typically use this [blog post](https://www.gresearch.co.uk/blog/article/advanced-recursion-techniques-in-f/) as a reference.

This PR also misses a test case to illustrate the problem is indeed solved.
Could you add a new test file `ASTTransformerTests.fs` (`module Fantomas.Core.Tests.ASTTransformerTests`)

with some unit test like:

```fsharp
#if RELEASE
open FSharp.Compiler.Text
open FSharp.Compiler.Xml
open FSharp.Compiler.Syntax
open FSharp.Compiler.SyntaxTrivia

[<Test>]
let ``avoid stack-overflow in long array/list, 2485`` () =
    let testIdent = Ident("Test", Range.Zero)

    let mkStringExpr () =
        SynExpr.Const(
            SynConst.String((System.Guid.NewGuid().ToString("N"), SynStringKind.Regular, Range.Zero)),
            Range.Zero
        )

    let longArrayExpr: SynExpr =
        let rec mkArray count childExpr =
            if count = 20_000 then
                childExpr
            else
                mkArray
                    (count + 1)
                    (SynExpr.Sequential(
                        DebugPointAtSequential.SuppressNeither,
                        true,
                        mkStringExpr (),
                        childExpr,
                        Range.Zero
                    ))

        SynExpr.ArrayOrListComputed(true, mkArray 0 (mkStringExpr ()), Range.Zero)

    let rootNode =
        Fantomas.Core.AstTransformer.astToNode
            Range.Zero
            []
            [ SynModuleOrNamespace(
                  [ testIdent ],
                  false,
                  SynModuleOrNamespaceKind.AnonModule,
                  [ SynModuleDecl.Expr(longArrayExpr, Range.Zero) ],
                  PreXmlDoc.Empty,
                  [],
                  None,
                  Range.Zero,
                  { ModuleKeyword = None
                    NamespaceKeyword = None }
              ) ]

    Assert.Pass()
#endif
```

Notice that `#if RELEASE` will be required for that test to succeed.

Also please take a look at the [PR checklist](https://github.com/fsprojects/fantomas/blob/master/CONTRIBUTING.md#pull-request-ground-rules)

Many thanks

njlr:
Thanks @nojaf I have followed the steps in the checklist :+1:

nojaf (approved):
Thanks!

Issue #2317, opened 2022-07-01
Optimize mobile experience


nojaf:
There are still a couple of quirks on the mobile website.
Both the landing page and the documentation.

## Landing page

- The typing text is running through the buttons. Maybe we should not display that on mobile.
 
![image](https://user-images.githubusercontent.com/2621499/189299789-6405cf91-e66f-4be6-a871-e7ca3a919f5e.png)

- The community sponsors don't look nice.

![image](https://user-images.githubusercontent.com/2621499/189299896-cdad315b-8924-4b64-a871-eda0fda8c3c2.png)

- Missing space between "Fantomas Daemon" and "World domination, part one"

![image](https://user-images.githubusercontent.com/2621499/189300034-2ed6da0a-26cd-4c81-b45c-41efc57b3064.png)

- Not mobile-related but hide the "News" menu item, it doesn't go anywhere right now.

## Documentation

- There is a long scrollbar when the mobile menu is open.

![image](https://user-images.githubusercontent.com/2621499/189300305-ede91555-8ec4-4475-93ff-4c51e242fc48.png)

PR #2496, opened 2022-09-10
Mobile optimizations

alanlomeli:
(#2317)
Typing text hidden on mobile
Community sponsors optimized on mobile
Removal of news menu item
Minor tweaks
WIP: Still figuring out why the docs sidebar scrolls

### DIFF
--- a/docs/.style/homepage.sass
+++ b/docs/.style/homepage.sass
@@ -9,10 +9,11 @@ html, body
   border: 1px solid red
 
 header
-  height: 70vh
+  height: 500px
   position: relative
-  padding: 4rem 0
+  
   @include media-breakpoint-up(md)
+    padding: 4rem 0
     height: 100vh
 
   video
@@ -92,7 +93,7 @@ header
   #cta-buttons
     z-index: 1
     position: absolute
-    bottom: 9rem
+    bottom: 2rem
     left: 50%
     transform: translateX(-50%)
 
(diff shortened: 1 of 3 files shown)


docs/.style/homepage.sass
```diff
@@ -76,20 +76,25 @@ header
...
   #core-title
+    display: none
     color: $white
     text-align: center
     font-size: 4rem
     position: absolute
     top: 50%
     left: 50%
     transform: translate(-50%, -50%)
+    @include media-breakpoint-up(lg)
```
nojaf, line 87: Use `md` here, we really just don't want it on mobile.

docs/.style/homepage.sass
```diff
@@ -98,6 +103,24 @@ header
       margin: 0 1rem
       margin-bottom: 1rem
 
+.card
```
nojaf, line 106: I don't like this fix, this works by accident, if is was a `tweet` and `card` that were stacked up on mobile this might not work.

You want to change the `col-md-4` div to `col-md-4 mt-4 mt-md-0`. The second and third columns should push on mobile but not for `md` and higher.

nojaf (changes requested):
Now that the text is hidden the first blue screen is rather empty:

image

Could you address that?

The header doesn't need the `1o0vh` height and the buttons would need to be repositioned a bit.

nojaf:
@alanlomeli I found out we can solve the menu height problem by modifying the max height:

image

alanlomeli:
Thanks for the help :)

nojaf (approved):
Thanks!

PR #2509, opened 2022-09-16
Fix spacing for Const and Ident expressions in NamedIndexedPropertySet

dawedawe:
Fixes #2498

PR #2514, opened 2022-09-17
Use sepNone instead of id to improve clarity

dawedawe:
Besides [this](https://github.com/fsprojects/fantomas/pull/2509#discussion_r973543837) I found two other occurences of `id` which imho could be made more clear by using `sepNone`.


nojaf (approved):
Thansk!

PR #2520, opened 2022-09-18
v5.1

nojaf:
Open Pull Request with everything for the v5.1 release.
Currently, there is no fixed scope for this release.
We wish to keep the master branch for bug fixing of `v5.0`.
Everything that has a bigger impact than that should be targeted against the `v5.1` branch.


nojaf:
Because of the large delta between this branch and the main branch, it is more convenient to continue the remaining work on main.
//cc @dawedawe

PR #2522, opened 2022-09-20
Include solution items in .sln

josh-degraw:
This just adds existing files to the `.sln` file for easier access from within Rider/Visual Studio.

I just chose the name `Solution Items` because I think that's a common convention in Visual Studio, but the name is purely arbitrary so if you would prefer a different name that's 100% fine by me haha.

image


fantomas.sln
```diff
@@ -20,15 +20,30 @@ Project("{6EC3EE1D-3C4E-46DD-8F32-0CC8E7565705}") = "Fantomas.Benchmarks", "src\
 EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "docs", "docs", "{29F22904-C9E0-40AA-B971-9AE377C17F59}"
 	ProjectSection(SolutionItems) = preProject
-		docs\Documentation.md = docs\Documentation.md
-		docs\FormattingConventions.md = docs\FormattingConventions.md
```
josh-degraw, line 24: Removed these from the solution since they no longer exist.

nojaf: Thanks!

nojaf (approved):

Issue #2525, opened 2022-09-22
Comments lost from multi-line expressions


SteveGilham:
[...]

PR #2530, opened 2022-09-27
Construct a more accurate range for SynAttribute 

SteveGilham:
Resolve some line comment issues by improving the heuristics used attach them to the source.

* by scanning children of enclosing nodes
* by making a better selection of enclosing node

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## [Unreleased]
+
+### Fixed: *  Comments lost from multi-line expressions [#2525](https://github.com/fsprojects/fantomas/issues/2525)
+
 ## [5.0.2] - 2022-09-22
 
 ### Fixed
--- a/src/Fantomas.Core.Tests/CommentTests.fs
+++ b/src/Fantomas.Core.Tests/CommentTests.fs
@@ -2515,3 +2515,63 @@ type X =
             // The 'Cancel' member
             (System.IAsyncResult -> unit)
 """
+
+[<Test>]
+let ``should keep mid quotation line comments, related to 2525`` () =
+    formatSourceString
+        false
+        """test
+    <@
+      result
+        .Replace('\r', '\u00FF')
+        .Replace('\n', '\u00FF')
+        .Replace("\u00FF\u00FF", "\u00FF")
+        .Replace("8.12", "8.13") // CRAP score rounding
+        .Replace("4.12", "4.13") // CRAP score rounding
+        .Trim([| '\u00FF' |]) = expected
+        .Replace('\r', '\u00FF')
+        .Replace('\n', '\u00FF')
+        .Replace("\u00FF\u00FF", "\u00FF")
+        .Trim([| '\u00FF' |])
+    @>
+"""
+        config
+    |> should
+        equal
+        """test
+    <@
+        result
+            .Replace('\r', '\u00FF')
+            .Replace('\n', '\u00FF')
+            .Replace("\u00FF\u00FF", "\u00FF")
+            .Replace("8.12", "8.13") // CRAP score rounding
+            .Replace("4.12", "4.13") // CRAP score rounding
+            .Trim([| '\u00FF' |]) = expected
+            .Replace('\r', '\u00FF')
+            .Replace('\n', '\u00FF')
+            .Replace("\u00FF\u00FF", "\u00FF")
+            .Trim([| '\u00FF' |])
+    @>
+"""
+
+[<Test>]
+let ``should keep mid attribute line comments, 2525`` () =
+    formatSourceString
+        false
+        """[<assembly: SuppressMessage("Gendarme.Rules.Performance",
+                            "AvoidRepetitiveCallsToPropertiesRule",
+                            Scope = "member",  // MethodDefinition
+                            Target = "AltCover.Recorder.Instance/I/CallTrack::instance()",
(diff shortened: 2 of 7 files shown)


nojaf (changes requested):
Hello Steve,

Thank you for your extended interest in this project. It is exciting to see you on the PR side here!

I'm afraid the proposed changes are way too impactful for the problem you are trying to solve.

The change of looking at all children is my main concern here. The tree structure was refactored in v5 to avoid having to look through all the nodes.
This can be quite impactful for performance, consider this [example](https://fsprojects.github.io/fantomas-tools/#/trivia?data=N4KABGBEDGD2AmBTSAuKAdATugdgG0QBcwBbRACzAF4wBnRARzGC1wggE8BLRPeMAIwBuVjnZhuvfsNHjJfQSOxj286UraqeCmcrnb1srVMVHOB08oC%2BYAPS2wcEmRyFIAGnBQkAMy45EWlQoDy9ILloAMVouYJ8AQzx6ECsgA).
Here the container node is `SynModuleDecl_Let (2, 0 - 11, 1)`, and with your proposal every node in between would be tested.

As I mentioned in my [comment](https://github.com/fsprojects/fantomas/issues/2525#issuecomment-1257983187), the range of SynAttribute is wrong and thus  `let smallestNodeThatContainsTrivia = findNodeWhereRangeFitsIn rootNode trivia.Range` will not lead to a good container node.
I've raised https://github.com/dotnet/fsharp/pull/13998 to address this.
But we could correct the range ahead of the upstream fix.
In `AstExtensions.fs` you could add a type extension `FullRange` which constructs the correct range we are after. This property could then be used in both `ASTTransformer` and `CodePrinter`.
Similar to `SynExprRecordField.FullRange`,
it is used in

https://github.com/fsprojects/fantomas/blob/60fc9026e897ec67118a5f12410de070996173ad/src/Fantomas.Core/AstTransformer.fs#L817-L824
and 
https://github.com/fsprojects/fantomas/blob/dc8b277571f20c3935188549a65fc6a3539585ef/src/Fantomas.Core/CodePrinter.fs#L851-L856

This change would be way less impactful and only address the issue at hand. This is what we want for a bugfix, we don't want to risk a change in behaviour anywhere else.

Thanks for understanding.

SteveGilham:
Trying instead

Screenshot 2022-09-28 104850

with the only plausible definition

Screenshot 2022-09-28 104923

* badly breaks several other unit tests, making this no longer a simple letterbox change when those breaks are also addressed
* due to the recursive nature of `findNodeWhereRangeFitsIn` still ends up doing much of the tree-walking that you are anxious to avoid

Using `FullRange` instead of range for the container node children in lines 273-4 of `lineCommentAfterSourceCodeToTriviaInstruction` does not break any existing tests, but neither does it fix the issue.

At this point, I think I'm blocked.

nojaf:
Could you please re-read my response? I did not at all propose to add an extension method on `TriviaNode`. I very specifically pointed to [SynAttribute](https://fsprojects.github.io/fantomas/reference/fsharp-compiler-syntax-synattribute.html) because that is where your problem is. There is no need to touch anything in `Trivia.fs` here.

nojaf:
> Notice that SynAttribute_ (1, 12 - 1, 27) only covers the word SuppressMessage.
Because of this incorrect range, the children of SynAttribute_ are never considered.

> We could use a different (artificial) range when we create the node in ASTTransformer and when we print its trivia in CodePrinter.

nojaf:
Are you using this `FullRange` in `CodePrinter`?

https://github.com/fsprojects/fantomas/blob/dc8b277571f20c3935188549a65fc6a3539585ef/src/Fantomas.Core/CodePrinter.fs#L464-L475

nojaf:
`|> genTriviaFor SynAttribute_ attr.FullRange`

SteveGilham:
That was the bit I'd missed, while still feeling my way through the code base.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## [Unreleased]
+
+### Fixed: *  Comments lost from multi-line expressions [#2525](https://github.com/fsprojects/fantomas/issues/2525)
```
nojaf, line 5: Please revisit this. Look very closely at how the rest of the file was set up.
Put the entry on the following line, remove unnecessary spaces and add a `.` after the issue title.

src/Fantomas.Core.Tests/CommentTests.fs
```diff
@@ -2515,3 +2515,63 @@ type X =
             // The 'Cancel' member
             (System.IAsyncResult -> unit)
 """
+
+[<Test>]
```
nojaf, line 2519: Please remove this test, it is outside of the scope of your PR.

nojaf (changes requested):
Almost there!

nojaf:
Could you also make sure that issue is probably linked here on GitHub? (see [docs](https://fsprojects.github.io/fantomas/docs/contributors/Pull%20request%20ground%20rules.html#Pull-request-title)).

image

Ref: https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword

SteveGilham:
I think it's simpler for me to just keep backing out the comment-deleting changes rather than continuing to fail to read your minds.

Good day, gentlemen.

PR #2531, opened 2022-09-27
Fake no more

nojaf:
Hey @dawedawe, @baronfel! I would propose this as an alternative for https://github.com/fsprojects/fantomas/pull/2303.
I still need to finish this but the general gist is that:
- `dotnet fsi build.fsx` runs the main build.
- You can run another pipeline by adding `-p`, example: `dotnet fsi build.fsx -p FormatChanged`.

Steps inside stages can be run in parallel (using `paralle`). This makes it more compelling than the MSBuild route. I would also like to add a `Watch` pipeline that does everything to run the docs locally.

TODO:
- [x] Update documentation
- [x] Add `Watch` pipeline
- [x] Verify everything is as expected in the CI.

Let me know what you think!


dawedawe (approved):
Super nice. I would suggest removing the `fake-sample` folder to be complete.

nojaf:
> I would suggest removing the fake-sample folder to be complete.

That is a bit of a separate thing. There is using FAKE in the Fantomas repository, which I want to get rid of in this PR. And then there is using Fantomas inside a FAKE scenario, which the example serves.

Just because we don't want to use FAKE ourselves, doesn't mean we should remove an example of how you could use Fantomas inside FAKE.

On the other hand, less is more 😸

dawedawe:
> > I would suggest removing the fake-sample folder to be complete.
> 
> That is a bit of a separate thing. There is using FAKE in the Fantomas repository, which I want to get rid of in this PR. And then there is using Fantomas inside a FAKE scenario, which the example serves.
> 
> Just because we don't want to use FAKE ourselves, doesn't mean we should remove an example of how you could use Fantomas inside FAKE.
> 
> On the other hand, less is more 😸

Uh, yeah, you're right. I blame jet lag 😁 
I think for the time being, we should keep it then.

Issue #2335, opened 2022-07-06
Add line break before start of argument list 


alanlomeli:
[...]

Issue #2111, opened 2022-02-21
Vanity alignment used inside base ctor call


knocte:
[...]

PR #2555, opened 2022-10-04
Change style of base constructor calls.

nojaf:
Update according to the guidance of https://github.com/fsharp/fslang-design/issues/693.
Fixes https://github.com/fsharp/fslang-design/issues/693.
Fixes https://github.com/fsprojects/fantomas/issues/2111
Fixes https://github.com/fsprojects/fantomas/issues/2335

@Smaug123 there is a minor impact for [MultilineBlockBracketsOnSameColumn](https://fsprojects.github.io/fantomas/docs/end-users/Configuration.html#fsharp_multiline_block_brackets_on_same_column).
I don't think you really use that much. Nevertheless, does it seem acceptable?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,11 @@
 # Changelog
 
-## [Unreleased]
+## [5.1.0-alpha-005] - 2022-10-07
 
 ### Changed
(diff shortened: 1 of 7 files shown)


Smaug123:
Am I right in thinking this doesn't affect `{ new obj() with...` - only `inherit` calls? If so, I think we never use `{ inherit ...` at all, so this should be fine.

nojaf:
Yup, object expressions are not affected by this.

PR #2569, opened 2022-10-09
First stab at a glossary

### LINKED ISSUE #2493: Glossary
We should have a glossary page with all the abbreviations and synonyms we use for different concepts.

Things like:
- PR
- AST, syntax node, untyped tree
- range
- FCS
- ...

dawedawe:
Here's a first stab at a glossary to fix #2493 
I'm not sure about the scope we should cover with this.
Opinions?

### DIFF
--- a/docs/docs/contributors/Glossary.md
+++ b/docs/docs/contributors/Glossary.md
@@ -7,7 +7,7 @@ index: 1000
 
 ## AST
 
-The [Abstract Syntax Tree](https://en.wikipedia.org/wiki/Abstract_syntax_tree) (AST) is a tree data [structure](https://github.com/dotnet/fsharp/blob/main/src/Compiler/SyntaxTree/SyntaxTree.fs) representing a piece of source code.
+The [Abstract Syntax Tree](https://en.wikipedia.org/wiki/Abstract_syntax_tree) (AST) is a tree data [structure](https://github.com/dotnet/fsharp/blob/main/src/Compiler/SyntaxTree/SyntaxTree.fsi) representing a piece of source code.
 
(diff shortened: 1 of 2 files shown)


docs/docs/contributors/Glossary.md
```diff
@@ -0,0 +1,44 @@
...
+
+A node in the AST.
+
+## Trivia
+
+Used to label items (blank lines, code comments, conditional directives) that are not fully captured by the F# compiler in the AST.
+
+## Trivia Node
+
+A node in a special tree structure composed to help Fantomas print code comments.
```
nojaf, line 38: I would go a tad more abstract and general for this definition.
A trivia node can contain trivia, either as content before or content after.
A trivia node is a generalized type that serves as a common denominator for all AST node types. Maybe also mention the tree structure (parent/child relations).

docs/docs/contributors/Glossary.md
```diff
@@ -0,0 +1,44 @@
...
+
+## Untyped Syntax Tree
+
+The AST from the FCS used by Fantomas.
```
nojaf, line 42: I would mention that the untyped tree represents what the user wrote in the source code.
Stress that it does not know what the code does or whether it is valid F# code.
And during the compilation of F# code, the untyped tree is transformed into the typed tree.

We could then add `## Typed tree`, mentioning this has typed information about the source code and is not used in Fantomas.

nojaf (changes requested):
A good first attempt!

nojaf (approved):
There is a writer in you! I can appreciate that 😊.
Well done!

PR #2570, opened 2022-10-10
Allow to disable indentation of DU cases

anchann:
First PR here, sorry for any blunders.

We are in the process of adopting Fantomas into our code base, and need some additional configuration options to make the auto-formatted code compatible with our code guide. From the docs I couldn't quite figure out how strong your stance is on "only official formatting!", but it seems that you allowed some customizations (like stroustrup style, which is awesome!), so hopefully what I'm trying to do is okay.

Question: even though the code update works on the CLI, I could not get the unit test I wrote to pass, and can't figure out why. Please help.

### DIFF
--- a/src/Fantomas.Core.Tests/UnionTests.fs
+++ b/src/Fantomas.Core.Tests/UnionTests.fs
@@ -1132,3 +1132,24 @@ type Foo =
     /// Bar
     | Thing
 """
+
+[<Test>]
+let ``disabling DU case indentation`` () =
+    formatSourceString
+        true
+        """
+type Fruit =
+| Banana
+| Apple
+| Mango
+"""
+        { config with IndentDiscriminatedUnionCases = false }
+    |> prepend newline
+    |> should
+        equal
+        """
+type Fruit =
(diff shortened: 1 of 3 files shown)


nojaf:
Hello,

We are extremely strict on what settings are added to Fantomas, so I will not consider any PR without prior approval via the feature template. And again, [we are not looking for stylistic settings](https://github.com/fsprojects/fantomas/blob/main/CONTRIBUTING.md#what-are-we-not-looking-for).

Stroustrup is an exception to this, as we have an implementation prior to the style guide approval. That being said, the people that are interested in Stroustrup have engaged in dialogue with the style guide (see https://github.com/fsharp/fslang-design/issues/706).
It took Stroupstrup years to get into Fantomas, so we did not tread this lightly.

What you propose here is against the guidance of both style guides, and thus we are not interested in this setting. 
It is not worth it, and you are your team will need to get accustomed [to what Fantomas produces as output](https://fsprojects.github.io/fantomas/docs/end-users/StyleGuide.html#Let-it-go).

PR #2577, opened 2022-10-13
Leading Keywords

nojaf:
@dawedawe a lot of good stuff.
`SynLeadingKeyword` allows us to generalize almost all `SynBinding` nodes, regardless of their parent.

This streamlines and simplifies a lot of things in CodePrinter.


baronfel:
Any impact on perf/memory usage with this change, now that's it's all in the parser?

nojaf:
Not really

Before
```
Benchmark/step-0> | Method |     Mean |   Error |  StdDev | Rank |      Gen 0 |     Gen 1 |    Gen 2 | Allocated |
Benchmark/step-0> |------- |---------:|--------:|--------:|-----:|-----------:|----------:|---------:|----------:|
Benchmark/step-0> | Format | 210.7 ms | 0.76 ms | 0.63 ms |    1 | 14333.3333 | 1333.3333 | 333.3333 |    176 MB |
```
After
```
Benchmark/step-0> | Method |     Mean |   Error |  StdDev | Rank |      Gen 0 |     Gen 1 |    Gen 2 | Allocated |
Benchmark/step-0> |------- |---------:|--------:|--------:|-----:|-----------:|----------:|---------:|----------:|
Benchmark/step-0> | Format | 206.4 ms | 1.63 ms | 1.44 ms |    1 | 13666.6667 | 1333.3333 | 333.3333 |    170 MB |
```

Although I wasn't really expecting much there. The nice thing about this PR is that we can print a `SynBinding` by just looking at the `SynBinding` info itself.
No need to know if it came from a `SynModuleDecl.Let` or `SynMemberDefn.Member`.
This streamlines a lot of things.

PR #2635, opened 2022-12-03
Add target to download FCS files

nojaf:
@baronfel, this worked for me though I can't really tell whether I'm hooking this before the right target. Any thoughts?


baronfel:
That should be fine - CollectPackageReferences is one of the only known Restore prerequisites, so it's as good a target as any (until the SDK does something more useful here).

baronfel (approved):

PR #2639, opened 2022-12-04
call genAttributes for ExplicitCtor node

dawedawe:
fixes #2638

PR #2640, opened 2022-12-04
Print attributes for MemberDefn.ExplicitCtor.

nojaf:
@dawedawe I was going to give you a pointer to see if you found your way around the new code.
But the change was so simple I just went ahead with it.
I deliberately didn't add a changelog entry. Having an unreleased section in both `main` and `v5.2` can be annoying when rebasing.

Steps I used to solve the problem:

Check if the `MemberDefnExplicitCtorNode` does have attributes. It does, so this means the type definition is correct. 

Next, verify if the untyped tree was correctly transformed into the Oak in Fangorn.

image

This is the case as you can see in the online tool.

Lastly, check in `CodePrinter2` if we are printing the attributes. It turned out we weren't.

### DIFF
--- a/src/Fantomas.Core.Tests/DallasTests.fs
+++ b/src/Fantomas.Core.Tests/DallasTests.fs
@@ -1755,3 +1755,35 @@ type TypeDefnUnionNode
     =
     inherit NodeBase(range)
 """
+
+[<Test>]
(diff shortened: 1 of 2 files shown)


nojaf:
Closing in favour of https://github.com/fsprojects/fantomas/pull/2639

PR #2656, opened 2022-12-20
Call ignoreFile.IsIgnored with absolute path.

nojaf:
I discovered that ignored files can potentially be formatted by `Fantomas.Client`.
Steps to reproduce:
- Clone https://github.com/dotnet/fsharp
- Run the following script:

```fsharp
#r "nuget: Fantomas.Client"

open System
open System.IO
open Fantomas.Client.Contracts
open Fantomas.Client.LSPFantomasService

let service = new LSPFantomasService() :> FantomasService

let file = @"C:\Users\nojaf\Projects\fsharp\src\Compiler\Checking\NicePrint.fs"
let content = File.ReadAllText file

let formatReq : FormatDocumentRequest = {
    SourceCode = content
    FilePath = file
    Config = None
}

service.FormatDocumentAsync(formatReq).Result.Code
```
The code will read `1` which means `Formatted`, you should receive `4` (`Ignored`) instead.

I believe I found a fix for this problem but I would like a review from @Smaug123 to be sure.

### DIFF
--- a/src/Fantomas/IgnoreFile.fs
+++ b/src/Fantomas/IgnoreFile.fs
@@ -3,9 +3,20 @@ namespace Fantomas
 open System.IO.Abstractions
 open Ignore
 
+type AbsoluteFilePath =
+    private
+    | AbsoluteFilePath of string
+
+    member x.Path =
+        let (AbsoluteFilePath(path)) = x
+        path
+
+    static member Create (fs: IFileSystem) (filePath: string) =
+        fs.Path.GetFullPath filePath |> AbsoluteFilePath
+
 /// The string argument is taken relative to the location
 /// of the ignore-file.
-type IsPathIgnored = string -> bool
+type IsPathIgnored = AbsoluteFilePath -> bool
 
 type IgnoreFile =
     { Location: IFileInfo
@@ -46,7 +57,7 @@ module IgnoreFile =
             (Ignore(), lines)
             ||> Array.fold (fun (ig: Ignore) (line: string) -> ig.Add(line))
 
-        fun path ->
+        fun (absoluteFilePath: AbsoluteFilePath) ->
             // See https://git-scm.com/docs/gitignore
             // We transform the incoming path relative to the .ignoreFilePath folder.
             // In a cli scenario that is the current directory, for the daemon it is the first found ignore file.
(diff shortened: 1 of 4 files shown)


src/Fantomas/IgnoreFile.fs
```diff
@@ -77,16 +77,7 @@ module IgnoreFile =
...
-                    if fullPath.StartsWith ignoreFile.Location.Directory.FullName then
-                        fullPath.[ignoreFile.Location.Directory.FullName.Length + 1 ..]
-                    else
-                        // This scenario is a bit unexpected - it suggests that we are
-                        // trying to ask an ignorefile whether a file that is outside
-                        // its dependency tree is ignored.
-                        fullPath
-
-                ignoreFile.IsIgnored path
+                ignoreFile.IsIgnored fullPath
```
nojaf, line 80: @Smaug123 I believe the problem of using a relative path here is the following:
In `loadIgnoreList` we process the path with
https://github.com/fsprojects/fantomas/blob/47c6ac6c15abf2830342866987ddd496be326c05/src/Fantomas/IgnoreFile.fs#L54-L58 

Looking at the [docs](https://learn.microsoft.com/en-us/dotnet/api/system.io.path.getrelativepath?view=net-7.0#remarks), they mention `GetFullPath` is invoked before calculating the difference.

In the case of the `cli tool`, the relative will be resolved with the `pwd` because that is where we assume the `ignore file` lives. For the daemon, on the other hand, the `pwd` could be something really random and thus the difference is calculated wrongly.

Does this make sense to you?

Smaug123: Sigh, yeah, that does make sense. But might it be better to only pass fully-qualified paths?

nojaf: > But might it be better to only pass fully-qualified paths?

Something like [this](https://github.com/fsprojects/fantomas/pull/2656/commits/5ccb13a3d328127ed034140ffcd10abf322ea580)?

Smaug123 (approved):
Seems pretty plausible to me, although I can't say I am 100% confident.

nojaf:
Thanks @Smaug123, I think this will be fine to try out.
As I've reported the issue myself, I don't think many people encounter this problem in the wild.

Smaug123:
Yep, sounds sane to me!

PR #2658, opened 2022-12-21
Update config to accept a single option for multiline_bracket_style

### LINKED ISSUE #2425: Consider making `fsharp_experimental_stroustrup_style` apply independently of `fsharp_multiline_block_brackets_on_same_column`
As mentioned in a comment in #2413:

> Using `fsharp_experimental_stroustrup_style` without `fsharp_multiline_block_brackets_on_same_column` does not make any sense. ExperimentalStroustrupStyle is an extension of MultilineBlockBracketsOnSameColumn.

Currently, to use the experimental Stroustrup setting you need to enable two flags. Other than the potential issue of not knowing you need to enable two flags, the `MultilineBlockBracketsOnSameColumn` is technically inaccurate when using the Stroustrup style, as by definition the brackets are on different columns.

I propose that:
- If `fsharp_experimental_stroustrup_style` is enabled, `fsharp_multiline_block_brackets_on_same_column` should be treated as redundant and relevant logic should be applied automatically whether it's present or not.
- Logic that would format differently when using Stroustrup vs `MultilineBlockBracketsOnSameColumn` should only need to check if the Stroustrup setting is enabled, 
- Logic that would apply identically for either Stroustrup or `MultilineBlockBracketsOnSameColumn`, but not otherwise, should check if either flag is present instead of only being applied if `MultilineBlockBracketsOnSameColumn` is true.

Alternatively, it might make sense to have a different flag to encapsulate the possible bracket styles (e.g. `stroustrup`, `allman`, `classic`, or whatever is most accurate). Admittedly a change like that could balloon out in logic handling, but for configuring the code it could make it much more straightforward to only need one flag to represent all potential bracket indentation styles.

Regardless, IMHO the biggest issue with the current api is mainly the fact that `MultilineBlockBracketsOnSameColumn` is factually incorrect when Stroustrup style is applied, so I think an adjustment is at least worth discussing due to the potential confusion it could cause.

josh-degraw:
Updates configuration to accept a single option for multiline bracket style. Names could be bike-shedded if necessary. I made sure to keep it backwards compatible, so existing editorconfig settings using either of the previously existing settings will be respected as before.

One other point I'm not in love with is that it currently requires `open Fantomas.Core.FormatConfig` any time it's used. I think I'd rather just move the type directly into the `Fantomas.Core` namespace, but I just left it for now. If you're open to that I'll move it.

Closes #2425


src/Fantomas.Core/FormatConfig.fs
```diff
@@ -25,6 +25,24 @@ type MultilineFormatterType =
...
+
+    static member ToConfigString(cfg: MultilineBracketStyle) =
+        match cfg with
+        | Classic -> "classic"
+        | Aligned -> "aligned"
+        | ExperimentalStroustrup -> "experimental_stroustrup"
+
+    static member OfConfigString(cfgString: string) =
+        match cfgString with
+        | "classic" -> Some(box Classic)
```
nojaf, line 41: I don't think you need to box here, `EndOfLineStyle` doesn't do it, `MultilineFormatterType` does.
Let's settle on not boxing here and always box in the `EditorConfig.fs` code.

src/Fantomas.Core/FormatConfig.fs
```diff
@@ -25,6 +25,24 @@ type MultilineFormatterType =
         | "number_of_items" -> Some(box MultilineFormatterType.NumberOfItems)
         | _ -> None
 
+type MultilineBracketStyle =
+    | Classic
```
nojaf, line 29: We probably want to go with something other else for `Classic`.
I think Don call this `Cramped` style. `Pico` is another [option I guess](https://en.wikipedia.org/wiki/Indentation_style#Pico_style).

src/Fantomas.Core/Context.fs
```diff
@@ -79,8 +79,11 @@ module WriterModel =
     let update maxPageWidth cmd m =
         let doNewline m =
             let m = { m with Indent = max m.Indent m.AtColumn }
+
```
nojaf, line 82: Please remove these additional newlines. You didn't touch this code.

src/Fantomas.Core/FormatConfig.fs
```diff
@@ -164,9 +182,9 @@ type FormatConfig =
       [<DisplayName("Maximum dot get expression width")>]
       MaxDotGetExpressionWidth: Num
 
-      [<Category("Boundaries")>]
-      [<DisplayName("Multiline-block brackets on same column")>]
-      MultilineBlockBracketsOnSameColumn: bool
+      // [<Category("Boundaries")>]
```
nojaf, line 185: Remove this for the final version of this PR.

src/Fantomas.Core/Context.fs
```diff
@@ -540,10 +545,11 @@ let ifElse b (f1: Context -> Context) f2 (ctx: Context) = if b then f1 ctx else
 
 let ifElseCtx cond (f1: Context -> Context) f2 (ctx: Context) = if cond ctx then f1 ctx else f2 ctx
 
-let ifStroustrupElse = ifElseCtx (fun ctx -> ctx.Config.ExperimentalStroustrupStyle)
+let ifStroustrupElse =
+    ifElseCtx (fun ctx -> ctx.Config.MultilineBracketStyle = ExperimentalStroustrup)
```
nojaf, line 549: Use `ctx.Config.ExperimentalStroustrupStyle` in this case. I like the additional member you created.

src/Fantomas.Core/Context.fs
```diff
@@ -540,10 +545,11 @@ let ifElse b (f1: Context -> Context) f2 (ctx: Context) = if b then f1 ctx else
...
 
 let ifStroustrup (f1: Context -> Context) =
-    ifElseCtx (fun ctx -> ctx.Config.ExperimentalStroustrupStyle) f1 id
+    ifElseCtx (fun ctx -> ctx.Config.MultilineBracketStyle = ExperimentalStroustrup) f1 id
```
nojaf, line 552: I also don't know if `ifStroustrupElse` and `ifStroustrup` are used anywhere. Please remove them if not.

src/Fantomas.Core/Context.fs
```diff
@@ -1027,7 +1043,14 @@ let sepSemi (ctx: Context) =
     <| ctx
 
 let ifAlignBrackets f g =
```
nojaf, line 1045: Maybe rename this to `ifAlignOrStroustupBrackets`?

src/Fantomas.Core/CodePrinter2.fs
```diff
@@ -3351,7 +3349,12 @@ let genTypeDefn (td: TypeDefn) =
                  +> sepNlnTypeAndMembers typeDefnNode
                  +> genMemberDefnList members)
                     ctx
-            else
+
+            match ctx.Config.MultilineBracketStyle with
+            | Aligned
+            | ExperimentalStroustrup -> aligned
+            | Classic when (List.exists (fun (fieldNode: FieldNode) -> fieldNode.XmlDoc.IsSome) node.Fields) -> aligned
```
nojaf, line 3356: The original code was already a bit icky but could you extract `List.exists (fun (fieldNode: FieldNode) -> fieldNode.XmlDoc.IsSome) node.Fields` to some value? Call it `anyFieldHasXmlDoc` maybe.

nojaf (changes requested):
This looks very good, well done!
A lot of nitpicks, but overall nothing major to adress.

We can move`FormatConfig` to the `Fantomas.Core` namespace in the next major release (6.x). It would be breaking if we did it now. Unless we alias it.

josh-degraw:
> We can move`FormatConfig` to the `Fantomas.Core` namespace in the next major release (6.x). It would be breaking if we did it now. Unless we alias it.

FWIW, I was specifically talking about _just_ moving `MultilineBracketStyle` into `Fantomas.Core`, not all of `FormatConfig`.

src/Fantomas.Core/FormatConfig.fs
josh-degraw: I like cramped!

dawedawe (changes requested):
Very, very nice, I'm all in favor of this.
Just one thing:
[Configuration.fsx](https://github.com/fsprojects/fantomas/blob/v5.2/docs/docs/end-users/Configuration.fsx) needs a little bit of adjustment.

nojaf:
> FWIW, I was specifically talking about just moving MultilineBracketStyle into Fantomas.Core, not all of FormatConfig.

Well, I think there is value in having most of the config type in that namespace.
That way you can `open Fantomas.Core` and access the `CodeFormatter` API without any additional open statements.

> [Configuration.fsx](https://github.com/fsprojects/fantomas/blob/v5.2/docs/docs/end-users/Configuration.fsx) needs a little bit of adjustment.

You can only update that once there is a version on NuGet with these changes.

dawedawe:
> > [Configuration.fsx](https://github.com/fsprojects/fantomas/blob/v5.2/docs/docs/end-users/Configuration.fsx) needs a little bit of adjustment.
> 
> You can only update that once there is a version on NuGet with these changes.

Uh, sorry. Nevermind...

dawedawe (approved):

nojaf (approved):
Thanks @josh-degraw! I'm very happy to see movement on this front.

Issue #2449, opened 2022-08-26
`else` followed by `//comment` and then `if` has incorrect indention


nhirschey:
[...]

PR #2663, opened 2022-12-23
Process `else if` using ElseIfNode.

nojaf:
Related to https://github.com/fsprojects/fantomas/issues/2449.
@dawedawe this might seem a bit complex and it definitely is a very creative way to approach the problem space.

In essence, I wrote a special type that deals with the trivia assignment of `else if`.
If a trivia would be assigned to either `else` or `if`, the best possible node is selected so that we can write `else if` in one piece without trivia interference.

This isn't without drawbacks but given the delicate nature this might do the trick.

### DIFF
--- a/src/Fantomas.Core/SyntaxOak.fs
+++ b/src/Fantomas.Core/SyntaxOak.fs
@@ -1469,9 +1469,9 @@ type ElseIfNode(mElse: range, mIf: range, condition: Node, range) as elseIfNode 
     let elseNode =
         { new Node with
             member _.ContentBefore: TriviaNode seq = Seq.empty
-            member x.HasContentBefore: bool = false
+            member _.HasContentBefore: bool = false
             member _.ContentAfter: TriviaNode seq = Seq.empty
-            member x.HasContentAfter: bool = false
+            member _.HasContentAfter: bool = false
             member _.Range = mElse
 
             member _.AddBefore(triviaNode: TriviaNode) =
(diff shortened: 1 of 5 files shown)


src/Fantomas.Core/SyntaxOak.fs
```diff
@@ -1461,11 +1461,89 @@ type ExprTryFinallyNode(tryNode: SingleTextNode, tryExpr: Expr, finallyNode: Sin
...
+
+            member _.AddAfter(triviaNode: TriviaNode) =
+                (elseIfNode :> Node).AddAfter triviaNode
+
+            member _.Children = Array.empty }
+
+    let ifNode =
+        { new Node with
+            member _.ContentBefore: TriviaNode seq = Seq.empty
+            member x.HasContentBefore: bool = false
```
dawedawe, line 1488: Mmh, I might be missing something obvious, but why the switch from `_` to `x` and vice versa in the members?

nojaf: Because this code started out as a copy of `NodeBase` and I changed members afterwards.
We perhaps also should settle on some code convention for this. 
When to use `_`. When `x` vs `this` vs `self`,..

dawedawe: Yes, good idea. But for the moment, I'm totally fine with just being consistent in the scope of a definition.

dawedawe (approved):

PR #2689, opened 2023-01-05
Refactor handling around bracket style in `genTypeDefn`

josh-degraw:
This is mainly an attempt to simplify the code around handling different bracket styles for record types, and to increase consistency and readability in `genTypeDefn` a little bit.

### DIFF
--- a/src/Fantomas.Core/CodePrinter.fs
+++ b/src/Fantomas.Core/CodePrinter.fs
@@ -3267,61 +3267,65 @@ let genTypeDefn (td: TypeDefn) =
             (indentSepNlnUnindent (sepNlnTypeAndMembers typeDefnNode +> genMemberDefnList members))
         |> genNode node
     | TypeDefn.Record node ->
-        let smallExpression =
-            sepSpace
-            +> genAccessOpt node.Accessibility
-            +> sepSpace
-            +> genSingleTextNode node.OpeningBrace
-            +> addSpaceIfSpaceAroundDelimiter
-            +> col sepSemi node.Fields genField
-            +> addSpaceIfSpaceAroundDelimiter
-            +> genSingleTextNode node.ClosingBrace
-
         let multilineExpression (ctx: Context) =
-            let aligned =
-                let msIsEmpty = List.isEmpty members
+            let noMembers = List.isEmpty members
 
-                (ifElseCtx
-                    (fun ctx -> ctx.Config.ExperimentalStroustrupStyle && msIsEmpty)
-                    (genAccessOpt node.Accessibility)
-                    (opt (indent +> sepNln) node.Accessibility genSingleTextNode)
-                 +> genSingleTextNode node.OpeningBrace
-                 +> indentSepNlnUnindent (atCurrentColumn (col sepNln node.Fields genField))
-                 +> sepNln
-                 +> genSingleTextNode node.ClosingBrace
-                 +> onlyIfCtx
-                     (fun ctx -> not (ctx.Config.ExperimentalStroustrupStyle && msIsEmpty))
-                     (optSingle (fun _ -> unindent) node.Accessibility)
-                 +> onlyIf (List.isNotEmpty members) sepNln
-                 +> sepNlnTypeAndMembers typeDefnNode
-                 +> genMemberDefnList members)
-                    ctx
+            let genRecordFields =
+                genSingleTextNode node.OpeningBrace
+                +> indentSepNlnUnindent (atCurrentColumn (col sepNln node.Fields genField))
+                +> sepNln
+                +> genSingleTextNode node.ClosingBrace
+
+            let genMembers =
+                onlyIf (not noMembers) sepNln
+                +> sepNlnTypeAndMembers typeDefnNode
+                +> genMemberDefnList members
+
+            let aligned =
+                opt (indent +> sepNln) node.Accessibility genSingleTextNode
+                +> genRecordFields
+                +> optSingle (fun _ -> unindent) node.Accessibility
+                +> genMembers
 
             let anyFieldHasXmlDoc =
                 List.exists (fun (fieldNode: FieldNode) -> fieldNode.XmlDoc.IsSome) node.Fields
 
-            match ctx.Config.MultilineBracketStyle with
-            | Aligned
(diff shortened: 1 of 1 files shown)


src/Fantomas.Core/CodePrinter.fs
```diff
@@ -3212,6 +3212,8 @@ let genTypeDefn (td: TypeDefn) =
         |> genNode typeName
 
     let members = typeDefnNode.Members
+    let hasMembers = List.isNotEmpty members
```
nojaf, line 3215: I'm not really a fan of this as members can occur twice for certain `TypeDefn`.
[Sample](https://fsprojects.github.io/fantomas-tools/#/oak?data=N4KABGBEDGD2AmBTSAuKAXAngB0WAgmALwA6ATiQHYQTQA2AhgM5PlU0dgC2iXARojJgAHgDoAQsTAAGNtRqJK8ORwDuAS3QALFZwg9%2BgkaIDCUgIyQANOChIAZusqImqKNduR1TAGJN1bvYMdEyINhBeTADK6GSwAK5MsfHYgcGhIAC%2BQA)

src/Fantomas.Core/CodePrinter.fs
```diff
@@ -3262,72 +3264,72 @@ let genTypeDefn (td: TypeDefn) =
...
-                 +> genSingleTextNodeSuffixDelimiter node.OpeningBrace
-                 +> atCurrentColumn (sepNlnWhenWriteBeforeNewlineNotEmpty +> col sepNln node.Fields genField)
-                 +> addSpaceIfSpaceAroundDelimiter
-                 +> genSingleTextNode node.ClosingBrace
-                 +> optSingle (fun _ -> unindent) node.Accessibility
-                 +> onlyIf (List.isNotEmpty members) sepNln
-                 +> sepNlnTypeAndMembers typeDefnNode
-                 +> genMemberDefnList members)
-                    ctx
+                node.Fields |> List.exists (fun fieldNode -> fieldNode.XmlDoc.IsSome)
```
nojaf, line 3289: When a module operation is short (`Module.fn lambda xs`) there really is no reason to turn this into `xs |> Module.fn lambda`.

Readability only benefits from this when the `lambda` is multiline.

src/Fantomas.Core/CodePrinter.fs
```diff
@@ -3262,72 +3264,72 @@ let genTypeDefn (td: TypeDefn) =
...
+
+            let genMultilineExpression =
```
nojaf, line 3291: Don't introduce a function value only to call it immediately as the next statement.

Refactor to
```fsharp
            match ctx.Config.MultilineBracketStyle with
            | ExperimentalStroustrup when hasNoMembers ->
                (genAccessOpt node.Accessibility +> genRecordFields +> genMembers) ctx
            | Aligned
            | ExperimentalStroustrup -> aligned ctx
            | Cramped when anyFieldHasXmlDoc -> aligned ctx
            | Cramped ->
                (sepNlnUnlessLastEventIsNewline
                 +> opt (indent +> sepNln) node.Accessibility genSingleTextNode
                 +> genSingleTextNodeSuffixDelimiter node.OpeningBrace
                 +> atCurrentColumn (sepNlnWhenWriteBeforeNewlineNotEmpty +> col sepNln node.Fields genField)
                 +> addSpaceIfSpaceAroundDelimiter
                 +> genSingleTextNode node.ClosingBrace
                 +> optSingle (fun _ -> unindent) node.Accessibility
                 +> onlyIf hasMembers sepNln
                 +> sepNlnTypeAndMembers typeDefnNode
                 +> genMemberDefnList members)
                    ctx
```
instead.

nojaf (changes requested):

src/Fantomas.Core/CodePrinter.fs
josh-degraw: The main reason I changed this was to remove the need for explicit type annotation

src/Fantomas.Core/CodePrinter.fs
josh-degraw: You'd prefer that even when it makes us have to wrap the more complicated expressions here in parentheses? IMHO it's more clear and readable without the additional parentheses when we can take advantage of partial application like that but I'm totally fine to change it back to that if you'd prefer.

src/Fantomas.Core/CodePrinter.fs
josh-degraw: Are you opposed to the name change here or moving it to the outer scope? Or both 😅?

nojaf: I'm opposed to having the `hasMembers` value on the `genTypeDefn` level (outside the match expression).
That would be confusing as `hasMembers` could be false while the concrete `TypeDefn` still has members. I would move `hasMembers` into the match clauses where you actually use it.
Because of that, there would be some duplication, but at least it would be less ambiguous.

src/Fantomas.Core/CodePrinter.fs
nojaf: I don't really mind explicit type annotations. I find myself adding them way more than is truly required by the compiler.

src/Fantomas.Core/CodePrinter.fs
nojaf: You currently have something like:
```fsharp
let multilineExpression (ctx: Context) =
    let genMultilineExpression = !- "code ..."
    genMultilineExpression ctx
```
which is the equivalent of storing the return value only to return it in the next line. That bothers me.

As a middle ground, I propose to introduce some more intermediate values.
Extract the clauses bodies with parentheses to `stroustrupWithoutMembers` and `cramped`.

End the function with:
```fsharp
match ctx.Config.MultilineBracketStyle with
| ExperimentalStroustrup when hasNoMembers -> stroustrupWithoutMembers ctx
| Aligned
| ExperimentalStroustrup -> aligned ctx
| Cramped when anyFieldHasXmlDoc -> aligned ctx
| Cramped -> cramped ctx
```

nojaf (approved):
Looks good! Thank you for all your patience 😊

PR #2693, opened 2023-01-06
Extend logging and verbosity options

### LINKED ISSUE #2627: [Feature request] Less verbose output
I propose we that allow a less verbose mode. 

The existing way of Fantomas deals with this problem is being very verbose doing `dotnet fantomas --recurse .` where is each file processed is being list..  It would be helpful to being less verbose not print any info about unchanged files.

## Pros and Cons 

The advantages of making this adjustment to Fantomas are less nosiy output allows user to focus on important things.

The disadvantages of making this adjustment to Fantomas are it is a change and someone might be dependent on current behavior of the cli tool. 
It might also might make ignored files harder to debug.

## Extra information

Estimated cost (XS, S, M, L, XL, XXL): S

Related suggestions: (put links to related suggestions here)

## Affidavit (please submit!)

* [X] I have read the [Contribution Guidelines](https://github.com/fsprojects/fantomas/blob/main/CONTRIBUTING.md).
* [X] I have [searched both open and closed suggestions on this site](https://github.com/fsprojects/fantomas/issues) and believe this is not a duplicate

Please tick all that apply:
* [ ] This is not a breaking change to Fantomas
* [X] I or my company would be willing to help implement and/or test this

dawedawe:
To show the general direction and maybe get some feedback, here's an early draft for the logging feature:
`--verbosity d` should be the same as in the past with the addition of the count of processed files in the end
`--verbosity n` should be pretty quiet

Fixes #2627

### DIFF
--- a/src/Fantomas/IgnoreFile.fs
+++ b/src/Fantomas/IgnoreFile.fs
@@ -2,6 +2,7 @@ namespace Fantomas
 
 open System.IO.Abstractions
 open Ignore
+open Fantomas.Logging
 
 type AbsoluteFilePath =
     private
@@ -90,5 +91,5 @@ module IgnoreFile =
             try
                 ignoreFile.IsIgnored fullPath
             with ex ->
-                printfn "%A" ex
+                stdlog "%A" ex
                 false
--- a/src/Fantomas/IgnoreFile.fs
+++ b/src/Fantomas/IgnoreFile.fs
@@ -91,5 +91,5 @@ module IgnoreFile =
             try
                 ignoreFile.IsIgnored fullPath
             with ex ->
-                stdlog "%A" ex
+                sprintf "%A" ex |> stdlog
                 false
--- a/src/Fantomas/IgnoreFile.fs
+++ b/src/Fantomas/IgnoreFile.fs
@@ -91,5 +91,5 @@ module IgnoreFile =
             try
                 ignoreFile.IsIgnored fullPath
             with ex ->
-                sprintf "%A" ex |> stdlog
+                elog $"%A{ex}"
                 false
--- a/src/Fantomas/Program.fs
+++ b/src/Fantomas/Program.fs
@@ -43,7 +43,7 @@ let time f =
     let sw = Diagnostics.Stopwatch.StartNew()
     let res = f ()
     sw.Stop()
-    sprintf "Time taken: %O s" sw.Elapsed |> stdlog
+    stdlog $"Time taken: %O{sw.Elapsed} s"
     res
 
 [<RequireQualifiedAccess>]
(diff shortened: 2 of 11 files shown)


src/Fantomas/IgnoreFile.fs
```diff
@@ -90,5 +91,5 @@ module IgnoreFile =
             try
                 ignoreFile.IsIgnored fullPath
             with ex ->
-                printfn "%A" ex
+                sprintf "%A" ex |> stdlog
```
nojaf, line 94: This probably should be an `elog`.

src/Fantomas/Program.fs
```diff
@@ -35,12 +37,13 @@ type Arguments =
...
                     "Input paths: can be multiple folders or files with %s extension."
                     (Seq.map (fun s -> "*" + s) extensions |> String.concat ",")
+            | Verbosity _ -> "Set the verbosity level. Allowed values are n[ormal] and d[etailed]."
 
 let time f =
     let sw = Diagnostics.Stopwatch.StartNew()
     let res = f ()
     sw.Stop()
-    printfn "Time taken: %O s" sw.Elapsed
+    sprintf "Time taken: %O s" sw.Elapsed |> stdlog
```
nojaf, line 46: I guess all the `stdlog` calls can be interpolated strings instead.
The extra pipe seems redundant.

src/Fantomas/Program.fs
```diff
@@ -166,39 +169,39 @@ let readFromStdin (lineLimit: int) =
         else
             Some(input)
 
-let private reportCheckResults (output: TextWriter) (checkResult: Format.CheckResult) =
+let private reportCheckResults (checkResult: Format.CheckResult) =
     checkResult.Errors
     |> List.map (fun (filename, exn) -> sprintf "error: Failed to format %s: %s" filename (exn.ToString()))
-    |> Seq.iter output.WriteLine
+    |> Seq.iter stdlog
```
nojaf, line 175: error log?

src/Fantomas/Program.fs
```diff
@@ -359,31 +392,37 @@ let main argv =
...
+            | InputPath.Multiple(files, folders), OutputPath.NotKnown ->
+                let n = filesAndFolders force files folders
+                logGrEqDetailed verbosity (sprintf "Processed files: %d" n)
             | InputPath.Multiple _, OutputPath.IO _ ->
-                eprintfn "Multiple input files are not supported with the --out flag."
+                elog "Multiple input files are not supported with the --out flag."
                 exit 1
         with exn ->
-            printfn "%s" exn.Message
+            stdlog exn.Message
```
nojaf, line 425: error log?

nojaf (commented):
This looks good so far!
I would be a little more consistent when using the logger functions. Just always go with an interpolated string. But that is really a nit.

I also think we should keep this for v6. Just to be on the safe side and get some user feedback.

dawedawe:
> This looks good so far! I would be a little more consistent when using the logger functions. Just always go with an interpolated string. But that is really a nit.
> 
> I also think we should keep this for v6. Just to be on the safe side and get some user feedback.

Thanks for the review. Waiting for v6 is fine for me.

nojaf:
I created a https://github.com/fsprojects/fantomas/tree/v6.0 branch which you can use to target this PR.

dawedawe:
If this is okay, I'd like to merge this, so Jimmy and me run into less conflicts with our next steps.
I'm thinking about the changes needed to report in the end how many files failed and how many were good...

nojaf (approved):
I think this is indeed ok to have as is now and revisit later.
Some food for thought:
- Do we not want to show the user which file was formatted? I think this makes sense in normal logging.
- Should we free up `-v` for verbosity? It is current used for version. It might make sense to use it for verbosity instead.
-  We will need to think about the format of logging. Seeing `[15:21:11 INF]` might be a bit misleading when `detailed` is active.

dawedawe:
> I think this is indeed ok to have as is now and revisit later. Some food for thought:
> 
>     * Do we not want to show the user which file was formatted? I think this makes sense in normal logging.
> 
>     * Should we free up `-v` for verbosity? It is current used for version. It might make sense to use it for verbosity instead.
> 
>     * We will need to think about the format of logging. Seeing `[15:21:11 INF]` might be a bit misleading when `detailed` is active.

Thanks a lot, I'll work on these.

PR #2696, opened 2023-01-09
Initial chain proof of concept.

nojaf:
An attempt to address https://github.com/fsharp/fslang-design/issues/688.
This is merely a suggestion on how to provide guidance. Nothing has been decided so far.

The gist of this PR is that everything that is `SynExpr.DotGet` related is being converted to an `Expr.Chain`. A `ChainNode` consists of links. Each link is either a dot or something between a dot.

Examples:
```fsharp
A.B.C().D
E.F(fun g -> g).H
I<j>().K[0].L(m,n,o).P
``` 

The expression between a dot could be multiple things:
- An identifier such as `A` or `A<b>`
- An application with unit `A()`
- An application with a parenthesis expression `A(b)` (including lambdas `A(fun b -> b)`)
- An index expression `.[A]`

Capturing all these different F# AST shapes into one unified model is already a big win.
A lot of very selective AST combinations can now easily be replaced with this. Regardless of the what the style will end up being.

### DIFF
--- a/src/Fantomas.Core.Tests/ChainTests.fs
+++ b/src/Fantomas.Core.Tests/ChainTests.fs
@@ -172,7 +172,7 @@ A().B(fun b -> b)
 """
 
 [<Test>]
-let ``2685 `` () =
+let ``space before lambda should not occur in chain, 2685 `` () =
     formatSourceString
         false
         """
--- a/src/Fantomas.Core.Tests/ChainTests.fs
+++ b/src/Fantomas.Core.Tests/ChainTests.fs
@@ -275,3 +275,40 @@ A.B
     .C(D)
     .E.[0]
 """
+
+[<Test>]
+let ``trivia inside chain, 2686`` () =
+    formatSourceString
+        false
+        """
+builder.
+    FirstThing<X>(fun lambda ->
+        // aaaaaa
+        ()
+    )
+    .SecondThing<Y>(fun next ->
+        // bbbbb
+        next
+    )
+    // ccccc
+    .ThirdThing<Z>().X
+"""
+        { config with
+            MultiLineLambdaClosingNewline = true }
+    |> prepend newline
+    |> should
+        equal
+        """
+builder
+    .FirstThing<X>(fun lambda ->
+        // aaaaaa
+        ()
+    )
+    .SecondThing<Y>(fun next ->
+        // bbbbb
+        next
+    )
+    // ccccc
+    .ThirdThing<Z>()
+    .X
+"""
--- a/src/Fantomas.Core.Tests/ChainTests.fs
+++ b/src/Fantomas.Core.Tests/ChainTests.fs
@@ -1,0 +1,277 @@
+module Fantomas.Core.Tests.ChainTests
+
+open NUnit.Framework
+open FsUnit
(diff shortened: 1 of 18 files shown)


src/Fantomas.Core.Tests/ChainTests.fs
```diff
@@ -0,0 +1,314 @@
...
+        config
+    |> prepend newline
+    |> should
+        equal
+        """
+A().B(fun b -> b)
+"""
+
+[<Test>]
+let ``2685 `` () =
```
dawedawe, line 175:
```suggestion
let ``space before lambda should not occur in chain, 2685 `` () =
```

dawedawe (approved):
Love it

src/Fantomas.Core.Tests/DotGetTests.fs
```diff
@@ -1245,7 +1190,8 @@ m
     .Property(fun p -> p.Name)
     .IsRequired()
     .HasColumnName("ModelName")
-    .HasMaxLength 64
+    .HasMaxLength
+    64
```
knocte, line 1194: @nojaf shouldn't this single line above have one extra level of indentation?

nojaf: No, I don't think so, `64` is the argument of the DotGet application.
Syntactically, this is the equivalent of 
```fsharp
let f =
    m
        .Property(fun p -> p.Name)
        .IsRequired()
        .HasColumnName("ModelName")
        .HasMaxLength

f 64
```

If the application doesn't fit on the rest of the line, the arguments are applied indented on the following line.

```fsharp
f
   64
```

Another extreme example of this would be:
```fsharp
(try
    id
 with ex ->
     id)
    64
```

In practice, I would write the code with parentheses 
```fsharp
    m
        .Property(fun p -> p.Name)
        .IsRequired()
        .HasColumnName("ModelName")
        .HasMaxLength(64)
```

knocte: I don't know, 64 looks more like the argument that is passed to the last func, and as such, would visually be better place there:

```
    .Property(fun p -> p.Name)
    .IsRequired()
    .HasColumnName("ModelName")
    .HasMaxLength 64
    .HasMaxLength
        64
```

But this is just because that 64 is the last one and is not another fluent call (it doesn't start with a dot). If it started with a dot, then I'd agree with you.

nojaf:
```fsharp
m
    .Property(fun p -> p.Name)
    .IsRequired()
    .HasColumnName("ModelName")
    .HasMaxLength 64
    .HasMaxLength
        64
```
This is no longer part of the same chain that starts with `m`.
What you wrote there is:
```fsharp
(m.Property(fun p -> p.Name).IsRequired().HasColumnName("ModelName").HasMaxLength)
     (64.HasMaxLength 64)
```

You can only add that space at the very end, the same reason `"ModelName"` is not applied with a space but has parentheses.

knocte: I didn't mean to write HasMaxLength twice, that was a typo, I edited my message 5 seconds after posting it

PR #2716, opened 2023-01-14
Refactor ExprEndsWithDualListAppNode and ExprEndsWithSingleListAppNode

nojaf:
I don't think the Oak should be affected by any stylistic settings.
So in that regard, `ExprEndsWithDualListAppNode` and `ExprEndsWithSingleListAppNode` were a bit unfortunate choices. Both cases are special groupings of ExprAppNode if experimental Stroustrup is active.

It feels more correct to move this to `CodePrinter` instead of `ASTTransformer`.

If we in the future want to have a grouping of applications that ends with a record, we can more easily facilitate this without the need of having an extra special `Node`.

### DIFF
--- a/src/Fantomas.Core/CodePrinter.fs
+++ b/src/Fantomas.Core/CodePrinter.fs
@@ -4,6 +4,7 @@ open System
 open Fantomas.Core.Context
 open Fantomas.Core.SyntaxOak
 open Fantomas.Core.FormatConfig
+open Microsoft.FSharp.Core.CompilerServices
 
 let noBreakInfixOps = set [| "="; ">"; "<"; "%" |]
 let newLineInfixOps = set [ "|>"; "||>"; "|||>"; ">>"; ">>=" ]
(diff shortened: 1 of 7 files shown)


src/Fantomas.Core/CodePrinter.fs
```diff
@@ -2127,6 +2128,39 @@ let genFunctionNameWithMultilineLids (trailing: Context -> Context) (longIdent:
     | _ -> sepNone
     |> genNode parentNode
 
+let (|EndsWithDualListApp|_|) (config: FormatConfig) (appNode: ExprAppNode) =
+    if not config.ExperimentalStroustrupStyle then
+        None
+    else
+        let mutable otherArgs = ListCollector<Expr>()
```
josh-degraw, line 2135: Does this need to be marked as mutable? Doesn't look like it's being reassigned anywhere

nojaf: Yes, it needs to be mutable for some reason.
It doesn't work without it and is listed as such in the [RFC](https://github.com/fsharp/fslang-design/blob/main/FSharp-6.0/FS-1099-list-collector.md).

josh-degraw: Huh, interesting, never seen that type before

josh-degraw (approved):

dawedawe (approved):

Issue #2624, opened 2022-11-14
[Feature request] Parallel formatting


safesparrow:
I propose we run formatting analysis in parallel, with a config option to control this.
The existing way of Fantomas deals with this problem is running most steps sequentially, I believe.

## Pros and Cons 

The advantages of making this adjustment to Fantomas are: much smaller wall-clock time of running fantomas.

The disadvantages of making this adjustment to Fantomas are:
- Some chance for race conditions
- Higher maximum CPU utilisation
- Higher maximum IO utilisation

## Extra information

Estimated cost (XS, S, M, L, XL, XXL): S?

## Affidavit (please submit!)

Please tick this by placing a cross in the box:
* [x] I have read the [Contribution Guidelines](https://github.com/fsprojects/fantomas/blob/main/CONTRIBUTING.md).
* [x] I have [searched both open and closed suggestions on this site](https://github.com/fsprojects/fantomas/issues) and believe this is not a duplicate

Please tick all that apply:
* [x] This is not a breaking change to Fantomas
* [x] I or my company would be willing to help implement and/or test this
* [ ] This suggestion is part of the [Microsoft style guide](https://docs.microsoft.com/en-us/dotnet/fsharp/style-guide/formatting) (please add a link to section if so)
* [ ] This suggestion is part of the [G-Research style guide](https://github.com/G-Research/fsharp-formatting-conventions) (please add a link to section if so)

nojaf:
Hello, the simplest thing to do would be to take a look at what can be done in parallel in the cli tool.
When formatting multiple files, that could be done in parallel. What happens inside the file (in `Fantomas.Core`) would be trickier and should be attempted in a later phase.

I'm all ears for a proposal 😉

safesparrow:
In the CLI case just formatting all source files in parallel would give more than enough.

I don't know how slow formatting a single file can be in the tooling case, but I imagine it's quite fast, so I don't think there is a need to speed that up.

safesparrow:
I attempted at PoC but for some reason a simple `Array.Parallel.map` for `processFile`/`stringToFile` is bottlenecked and doesn't use most cores on my machine (8C/16T). I don't yet understand what's happening:
<img width="1812" alt="image" src="https://user-images.githubusercontent.com/2478401/202059571-8d70f9b2-9fcc-4e8a-bcf2-33cd96eac43c.png">

nojaf:
[...]

PR #2717, opened 2023-01-14
Adds naive parallel formatting implementation

TheAngryByrd:
Fix for #2624.  Does a rather naive "collect all the async and run at the end" but seems to give really promising numbers.

Test results on my machine

BenchmarkDotNet=v0.13.2, OS=Windows 11 (10.0.22000.1455/21H2)
12th Gen Intel Core i9-12900F, 1 CPU, 24 logical and 16 physical cores
.NET SDK=7.0.100
  [Host]     : .NET 7.0.0 (7.0.22.51805), X64 RyuJIT AVX2 DEBUG
  Job-XAWGZF : .NET 7.0.0 (7.0.22.51805), X64 RyuJIT AVX2

Server=True  IterationCount=1  RunStrategy=ColdStart
|                  Method |             projects |       Mean |     Error |   StdDev |     Median | Ratio | RatioSD | Rank |       Gen0 |       Gen1 |      Gen2 | Allocated | Alloc Ratio |
|------------------------ |--------------------- |-----------:|----------:|---------:|-----------:|------:|--------:|-----:|-----------:|-----------:|----------:|----------:|------------:|
| **FormatCode_Sequentially** |                 **FAKE** | **3,872.3 ms** | **515.20 ms** | **593.3 ms** | **3,731.8 ms** |  **1.00** |    **0.00** |    **2** | **16000.0000** | **10000.0000** | **1000.0000** |   **4.55 GB** |        **1.00** |
|     FormatCode_Parallel |                 FAKE |   753.8 ms | 371.76 ms | 428.1 ms |   602.2 ms |  0.19 |    0.06 |    1 |  4000.0000 |  2000.0000 |         - |   4.55 GB |        1.00 |
|                         |                      |            |           |          |            |       |         |      |            |            |           |           |             |
| **FormatCode_Sequentially** | **FsToo(...)dling [23]** |   **862.6 ms** | **230.28 ms** | **265.2 ms** |   **792.4 ms** |  **1.00** |    **0.00** |    **2** |  **3000.0000** |  **2000.0000** |         **-** |   **1.15 GB** |        **1.00** |
|     FormatCode_Parallel | FsToo(...)dling [23] |   193.2 ms |  97.44 ms | 112.2 ms |   164.2 ms |  0.22 |    0.03 |    1 |  3000.0000 |  2000.0000 |         - |   1.15 GB |        1.00 |
|                         |                      |            |           |          |            |       |         |      |            |            |           |           |             |
| **FormatCode_Sequentially** |           **Plotly.NET** | **1,965.1 ms** | **300.60 ms** | **346.2 ms** | **1,884.8 ms** |  **1.00** |    **0.00** |    **2** |  **6000.0000** |  **3000.0000** |         **-** |   **2.29 GB** |        **1.00** |
|     FormatCode_Parallel |           Plotly.NET |   445.3 ms | 142.50 ms | 164.1 ms |   392.7 ms |  0.23 |    0.07 |    1 | 12000.0000 |  3000.0000 | 1000.0000 |    2.3 GB |        1.00 |
|                         |                      |            |           |          |            |       |         |      |            |            |           |           |             |
| **FormatCode_Sequentially** |               **fsharp** | **6,183.1 ms** | **634.73 ms** | **731.0 ms** | **6,004.4 ms** |  **1.00** |    **0.00** |    **2** | **20000.0000** | **10000.0000** | **2000.0000** |   **9.43 GB** |        **1.00** |
|     FormatCode_Parallel |               fsharp | 2,046.6 ms | 587.11 ms | 676.1 ms | 1,807.8 ms |  0.33 |    0.09 |    1 | 15000.0000 |  6000.0000 | 2000.0000 |   9.44 GB |        1.00 |
---

﻿Please verify your pull request is respecting our [Pull request ground rules](https://fsprojects.github.io/fantomas/docs/contributors/Pull%20request%20ground%20rules.html).
You can find more information in our [Contributors documentation section](https://fsprojects.github.io/fantomas/docs/contributors/Index.html).


src/Fantomas.CLI.Benchmarks/Runners.fs
```diff
@@ -0,0 +1,34 @@
+module Fantomas.CLI.Benchmarks.Runners
```
TheAngryByrd, line 1: We could add this into the current benchmarks project, I was just going off the initial work.

src/Fantomas/Program.fs
```diff
@@ -82,58 +88,64 @@ let rec allFiles isRec path =
 /// Fantomas assumes the input files are UTF-8
 /// As is stated in F# language spec: https://fsharp.org/specs/language-spec/4.1/FSharpSpec-4.1-latest.pdf#page=25
 let private hasByteOrderMark file =
-    if File.Exists(file) then
-        let preamble = Encoding.UTF8.GetPreamble()
+    async {
```
TheAngryByrd, line 91: Converted many functions in this file to `async` and used `Async` variants of all stream/file methods.

src/Fantomas/Program.fs
```diff
@@ -271,58 +283,70 @@ let main argv =
 
     let version = results.TryGetResult <@ Arguments.Version @>
 
+    let inline reraiseAsync (e: exn) =
+        let edi = ExceptionDispatchInfo.Capture e
+        edi.Throw()
```
TheAngryByrd, line 288: `reraise` doesn't work in computation expressions, this is the best work around.

However this is will stop all execution if one fails. I think we discussed that we should try as many files as we can, collect error results at the end. Is that right? @dawedawe @nojaf

src/Fantomas/Program.fs
```diff
@@ -271,58 +283,70 @@ let main argv =
...
-
-            use buffer =
-                if hasByteOrderMark then
-                    new StreamWriter(
-                        new FileStream(outFile, FileMode.OpenOrCreate, FileAccess.ReadWrite),
-                        Encoding.UTF8
-                    )
+        async {
+            try
+                printfn $"Processing %s{inFile}"
```
TheAngryByrd, line 293: All of this `printfn` aren't threadsafe and thus we could be printing garbage on the screen. Possibly we should hold off until @dawedawe merges https://github.com/fsprojects/fantomas/pull/2693?

src/Fantomas/Program.fs
dawedawe: I would prefer to do as many files as we can, yes.

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Changelog
 
+## [Unreleased]
```
nojaf, line 3: As this PR will need to be targeted against the v6 branch, it is easier not to have a changelog entry. Please remove it.

I would add all entries in the v6 branch right before a release.
This avoids git conflicts when rebasing v6 onto main.

src/Fantomas.CLI.Benchmarks/Runners.fs
nojaf: Yeah re-using the current benchmark project make more sense to me.

src/Fantomas/Fantomas.fsproj
```diff
@@ -2,6 +2,7 @@
   <PropertyGroup>
     <OutputType>Exe</OutputType>
     <TargetFramework>net6.0</TargetFramework>
+    <ServerGarbageCollection>true</ServerGarbageCollection>
```
nojaf, line 5: This should already be enabled by https://github.com/fsprojects/fantomas/blob/main/Directory.Build.props#L36

src/Fantomas/Program.fs
```diff
@@ -16,6 +18,7 @@ type Arguments =
     | [<Unique>] Out of string
     | [<Unique>] Check
     | [<Unique>] Daemon
+    | [<Unique>] Parallel
```
nojaf, line 21: I would introduce a `Sequential` flag instead and use parallel by default.

src/Fantomas/Program.fs
nojaf: Makes sense!

src/Fantomas/Program.fs
nojaf: Yup, I think all files should be formatted and at the end, we see which ones failed.

src/Fantomas/Program.fs
TheAngryByrd: ✔️

src/Fantomas/Fantomas.fsproj
TheAngryByrd: ✔️

CHANGELOG.md
TheAngryByrd: ✔️

src/Fantomas/Program.fs
TheAngryByrd: Rebased onto the 6.0 branch

src/Fantomas/Program.fs
TheAngryByrd: Ok this is changed to keep going thru all files. Reporting setup currently to look like:

`[17:42:59 INF] Formatted: 45, Ignored : 0, Unchanged : 72, Errored: 0`

src/Fantomas/Program.fs
```diff
@@ -325,27 +382,54 @@ let main argv =
...
+    let reportOnResults (results: #seq<ProcessResult>) =
+        let oks, ignored, unchanged, errored = partitionResults results
+
+        let summaryMessage =
+            $"Formatted: %d{oks.Length}, Ignored : %d{ignored.Length}, Unchanged : %d{unchanged.Length}, Errored: %d{errored.Length}"
+
+        stdlog summaryMessage
+
+        errored
+        |> Seq.iter (fun (file, error) -> elog $"Failed to format file: {file} : {error}")
```
TheAngryByrd, line 421: Should match on the exception. If it's a fantomas exception then should the error message without a stacktracke. Otherwise if it's verbose then put the the whole exception in, even if it's not from fantomas.

src/Fantomas.CLI.Benchmarks/Runners.fs
```diff
@@ -0,0 +1,35 @@
...
+let (</>) x y = Path.Combine(x, y)
+
+[<MemoryDiagnoser>]
+[<SimpleJob(runStrategy = RunStrategy.ColdStart, targetCount = 3)>]
+type ColdStart() =
+
+    let tmpDir = __SOURCE_DIRECTORY__ </> ".." </> ".." </> "tmp"
+
+    //? Should we download these repositories automatically?
+    [<Params("FsToolkit.ErrorHandling", "FAKE", "fsharp", "Plotly.NET")>]
```
TheAngryByrd, line 19: Should we download these repositories automatically as part of the suite?

nojaf: I have something else in mind for that. This shouldn't be part of this PR I think.

nojaf:
@TheAngryByrd is the `--check` flag also running in parallel?

TheAngryByrd:
> @TheAngryByrd is the `--check` flag also running in parallel?

I didn't touch it so doubtful.

TheAngryByrd:
> > @TheAngryByrd is the `--check` flag also running in parallel?
> 
> I didn't touch it so doubtful.

Seems to already be running parallel: https://github.com/fsprojects/fantomas/blob/main/src/Fantomas/Format.fs#L118

src/Fantomas.Tests/Integration/ConfigTests.fs
```diff
@@ -52,7 +52,7 @@ end_of_line=cr
 """
         )
 
-    let args = sprintf "%s %s" NormalVerbosity fileFixture.Filename
+    let args = sprintf "%s %s" DetailedVerbosity fileFixture.Filename
```
TheAngryByrd, line 55: We discussed we didn't want to show all exception messages unless they came from specific fantomas exceptions. However, fantomas uses `failwith` in many places instead of it's own exceptions.  Should we keep this requirement, always show all the exceptions, or rework all the places with `failwith` to throw a known Fantomas exception?

TheAngryByrd (commented):
1 requirement needed to be redefined but ready to review otherwise.

nojaf (changes requested):
Great work Jimmy! This looks really good!

As always, I had some nitpicks but with some tweaks, we can merge this first version I think.
I think we will need to dogfood this a bit to ensure everything is what we want.

src/Fantomas/Program.fs
```diff
@@ -17,6 +19,7 @@ type Arguments =
     | [<Unique>] Out of string
     | [<Unique>] Check
     | [<Unique>] Daemon
+    | [<Unique>] Sequential
```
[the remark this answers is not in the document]
dawedawe: I'd prefer to go without it.

nojaf:
Hi all, I made some liberal changes to this PR:

Normal logging:
image

Verbose logging:
image

Cliffnotes:
- Use `-v` for `--verbosity instead of `version`.
- Raise a specific `ParseException`.
- Set the Serilog level based on our internal flag.
- Removed the `Sequential` flag.
- Don't show `INF` when you are in `Normal` mode. MSBuild also doesn't show any specific level in their logs.

This by no means needs to be the final proposal for v6, but I think it is a step in the good direction and allows us to merge @TheAngryByrd's work.

What do you think @dawedawe @TheAngryByrd @josh-degraw @edgarfgp?

dawedawe:
I like it a lot. Had something similar in mind but planned to wait for the merge of this one.

nojaf (approved):

PR #2730, opened 2023-01-19
Update member declaration syntax in SyntaxOak.fs to be more consistent

josh-degraw:
There were a lot of members that introduced identifiers that went unused, it was inconsistent (e.g. sometimes `x.Thing`, sometimes `this.Thing`) and while the compiler was happy with that, it felt like additional noise. 

I updated `SyntaxOak.fs` to make things more consistent across node definitions: 
- If possible, use `member val`
- Use discard when identifier is needed but not used
- Otherwise use `x.Thing`, which seemed to be the most commonly used before

### DIFF
--- a/src/Fantomas.Core/SyntaxOak.fs
+++ b/src/Fantomas.Core/SyntaxOak.fs
@@ -97,7 +97,7 @@ type SingleTextNode(idText: string, range: range) =
 
 type MultipleTextsNode(content: SingleTextNode list, range) =
     inherit NodeBase(range)
-    override val Children = [| yield! nodes content |]
+    override val Children: Node array = [| yield! nodes content |]
     member val Content = content
 
 type XmlDocNode(lines: string array, range) =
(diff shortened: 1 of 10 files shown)


nojaf:
I think I would be ok with this.
You did seem to break a lot of tests, and we would need to capture these rules in our documentation.
Maybe create a new section under `Pull request ground rules`.

josh-degraw:
That's really weird, I'm getting all those test failures even if I revert to latest main. It seems to be adding an extra newline all over the place, are you seeing that on main too or is my environment just messed up somehow?

nojaf:
My latest main is passing, is your fork fully synced with everything?
Did you try a `git clean -xdf`?

josh-degraw:
Okay, I figured it out, that was weird! The breakages happened if I changed `member _.HasContentBefore` in the `interface Node` implementation of `NodeBase` to `member val HasContentBefore`. I think it's because `nodesBefore` was mutable, but `val` probably only evaluates once, which meant it would have been reporting that none of the nodes had any content before so it would add a newline. 🤦🏻‍♂️ Changed it so all the members in that implementation use `_.Thing` to keep visual consistency.

src/Fantomas.Core/SyntaxOak.fs
```diff
@@ -32,16 +32,23 @@ type NodeBase(range: range) =
     let nodesBefore = Queue<TriviaNode>(0)
     let nodesAfter = Queue<TriviaNode>(0)
 
+    member _.ContentBefore: TriviaNode seq = nodesBefore
```
nojaf, line 35: @josh-degraw is there any practical use for this?
`NodeBase` should never be used anywhere directly.

josh-degraw: This way you can access those members of any node without explicitly casting to `Node`

nojaf: Sweet!

nojaf (approved):
Thanks, I've added a few more tweaks following up on this.
Feel free to merge this one!

PR #2736, opened 2023-01-19
Move FormatConfig into Fantomas.Core namespace.

nojaf:
As mentioned on Discord, currently when using `Fantomas.Core` you need to open two namespaces for the basic 

```fsharp
#r "nuget: Fantomas.Core, 5.2.0-alpha-012"

open Fantomas.Core
open Fantomas.Core.FormatConfig

let config =
    { FormatConfig.Default with
        MaxLineLength = 90 }

CodeFormatter.FormatDocumentAsync(
    false,
    """
let a =  0
    """,
    config
)
|> Async.RunSynchronously
```

### DIFF
--- a/src/Fantomas.Core.Tests/CodePrinterHelperFunctionsTests.fs
+++ b/src/Fantomas.Core.Tests/CodePrinterHelperFunctionsTests.fs
@@ -3,7 +3,7 @@ module Fantomas.Core.Tests.CodePrinterHelperFunctionsTests
 open NUnit.Framework
 open FsUnit
 open Fantomas.Core.Context
-open Fantomas.Core.FormatConfig
+open Fantomas.Core
 open Fantomas.Core
 open Fantomas.Core.SyntaxOak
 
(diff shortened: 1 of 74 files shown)


src/Fantomas.Core.Tests/CodePrinterHelperFunctionsTests.fs
```diff
@@ -3,7 +3,7 @@ module Fantomas.Core.Tests.CodePrinterHelperFunctionsTests
 open NUnit.Framework
 open FsUnit
 open Fantomas.Core.Context
-open Fantomas.Core.FormatConfig
+open Fantomas.Core
 open Fantomas.Core
```
josh-degraw, line 7: Duplicate import here
```suggestion
open Fantomas.Core
```

nojaf: Good spot, there were a couple of those.

PR #2741, opened 2023-01-22
Let the icons (green, orange, ...) in the explanation be rendered.

dawedawe:
Currently the icons left of the explanation text aren't rendered.
I'm not much of a JS dev, so there's most likely a better fix.


nojaf (approved):
Looks good! Thanks!

PR #2746, opened 2023-01-25
Extract setting for preference of newline placement for computation expressions

### LINKED ISSUE #2276: Splitting ExperimentalStroustrupStyle to separate settings
I love the `ExperimentalStroustrupStyle` setting in records, but I hate it in computation expressions.
Computation expressions are so crucial part of F# that it should be very clear what goes on there.

What we want to optimize? Maintenance.
Let's observe the following code:

```fsharp
let myMethod() =
    async {
        let! x =
           async {
               do! waitSomething()
               return 4
           }
        return {
            Id = 1
            Data1 = "asdf"
            Data2 = "asdf"
            Data3 = x
        }
    }
```

There is no way to create this properly on any settings in Fantomas.

josh-degraw:
I'm very open to bike-shedding the name of the setting here, I just went with what made the most sense in my head at the time but I'm open to alternatives if something makes more sense.

﻿Closes #2276


src/Fantomas.Core/Context.fs
```diff
@@ -905,41 +923,28 @@ let addParenIfAutoNln expr f =
...
-    =
-    let isStroustrup =
-        ctx.Config.ExperimentalStroustrupStyle
-        && e.IsStroustrupStyleExpr
-        && Seq.isEmpty (Expr.Node e).ContentBefore
+let autoIndentAndNlnIfExpressionExceedsPageWidthUnlessSameLinePreferred f (e: Expr) (ctx: Context) =
+    let isStroustrup = isStroustrupApplicable e.IsStroustrupStyleExpr (Expr.Node e) ctx
 
-    if isStroustrup then
+    if isNamedComputationAndPreferSameLine e ctx || isStroustrup then
```
nojaf, line 947: If there is `TriviaBefore` in the `NamedComputation` that should be captured as well.
Example:
```fsharp
[<Test>]
let ``prefer computation expression name on same line, with trivia`` () =
    formatSourceString
        false
        """
let t =
    //
    task {
        let! thing = otherThing ()
        return 5
    }
"""
        config
    |> prepend newline
    |> should
        equal
        """
let t =
    //
    task {
        let! thing = otherThing ()
        return 5
    }
"""
```

src/Fantomas.Core/SyntaxOak.fs
```diff
@@ -1738,10 +1738,6 @@ type Expr =
             match node.CopyInfo with
             | Some _ -> false
             | None -> true
-        | Expr.NamedComputation node ->
```
nojaf, line 1741: Just a thought but I think this member should be extracted into a function in `Context`.
If you have that, you can pass the `FormatConfig` into it and just add an extra `when` expression for the named computation expression.

I think this makes sense as it further removes the coupling between the `Oak` model and any coding style.

By refactoring this, I think the current flow could be preserved slightly easier.

nojaf (changes requested):
Hey @josh-degraw, sorry for the slow response. I've been travelling last week.

From afar this PR looks good. I would explore the option to port some of the `Oak` code to `Context`.

```fsharp
let isStroustrup (config: FormatConfig) (e: Expr) =
    let isMultiRecord = config.MultilineBracketStyle = ExperimentalStroustrup

    match e with
    | Expr.Record node when isMultiRecord ->
        match node.Extra with
        | RecordNodeExtra.Inherit _
        | RecordNodeExtra.With _ -> false
        | RecordNodeExtra.None -> true
    | Expr.AnonRecord node when isMultiRecord ->
        match node.CopyInfo with
        | Some _ -> false
        | None -> true
    | Expr.ArrayOrList _ when isMultiRecord -> true
    | Expr.NamedComputation _ when config.PreferComputationExpressionNameOnSameLine -> true
    | _ -> false
```

I would however like to have another name for this.
The `Prefer` in `PreferComputationExpressionNameOnSameLine` really rubs me the wrong way. The word `prefer` is a bit excessive, given that it is a setting, it already implies a preference. On the top of my head, I would go in the direction of `StroustrupComputationExpression`. I'm not sure that exact name is what we want though, but I think `Stroustrup` is just so well known on the streets that I would somehow incorporate it.

If I understand the code right now, you only need to enable the setting to do something.
Having `MultilineBracketStyle = Stroustrup` is not required to enable the code. If that is the case, I do like that.

Lastly, this would need documentation. Perhaps we need to update the `configuration.fsx` script to use a local version of `Fantomas.Core` instead of a version on NuGet. Otherwise, we always need to play catch-up.

josh-degraw:
> If I understand the code right now, you only need to enable the setting to do something. Having `MultilineBracketStyle = Stroustrup` is not required to enable the code. If that is the case, I do like that.

Yep, that was the goal, this setting should be able to be enabled whether or not the user has enabled `Stroustrup` globally.

> I would however like to have another name for this. The `Prefer` in `PreferComputationExpressionNameOnSameLine` really rubs me the wrong way. The word `prefer` is a bit excessive, given that it is a setting, it already implies a preference. 

I agree the name is pretty chunky, I think removing `Prefer` makes total sense.

> On the top of my head, I would go in the direction of `StroustrupComputationExpression`. I'm not sure that exact name is what we want though, but I think `Stroustrup` is just so well known on the streets that I would somehow incorporate it.

I see what you mean, the only hesitation I have to that is that like you mentioned this is intended to be fully orthogonal to the `Stroustrup` setting, and I worry that it could cause confusion to have the name on this setting. I think `ComputationExpressionNameSameLine` or `SameLineComputationExpressionName` or `ShortComputationExpressions` or `CompressComputationExpressions` might work, but I'm not in love with either of them either ¯\\_(ツ)_/¯.

nojaf:
> I have to that is that like you mentioned this is intended to be fully orthogonal to the Stroustrup setting, and I worry that it could cause confusion to have the name on this setting.

Yeah, I can see why you don't want to attach it to the `Stroustrup` brand. It is less obvious to think about `computation expressions` in that way.

I'm not overly crazy about the `SameLine` wording. That feels a bit clunky still. However, no clue how to improve it. `ComputationExpressionStartOnCurrentLine`, I think `Start` is more general instead of `Name` (technically it could be an expression and not only a name). I'm not sure `CurrentLine` is any better than `SameLine` ¯\(ツ)/¯.

josh-degraw:
An alternative could be to do something like `NewlineBeforeComputationExpression` that defaults to true? I like the name better but would it be confusing at all to have a setting that defaults to true? I originally thought so but now I'm not sure.

nojaf:
Oh, I like that one. Maybe `NewlineBeforeMultilineComputationExpression` to remove all ambiguity.
(Although it becomes long again that way).

josh-degraw:
Hm. Yeah that's longer but I think I'm fine with that. It's long but it still doesn't feel quite as clunky as my original suggested name, IMO. I'm fine with that if you are.

src/Fantomas.Core/Context.fs
```diff
@@ -752,14 +752,31 @@ let sepSpaceOrDoubleIndentAndNlnIfExpressionExceedsPageWidth expr (ctx: Context)
...
+        match node.Extra with
+        | RecordNodeExtra.Inherit _
+        | RecordNodeExtra.With _ -> false
+        | RecordNodeExtra.None -> true
+    | Expr.AnonRecord node when isStroustrupEnabled ->
+        match node.CopyInfo with
+        | Some _ -> false
+        | None -> true
+    | Expr.ArrayOrList _ when isStroustrupEnabled -> true
+    | Expr.NamedComputation _ when config.PreferComputationExpressionNameOnSameLine -> true
```
nojaf: I didn't think about it earlier but we can probably have:
`| Expr.NamedComputation _ -> not config.NewlineBeforeMultilineComputationExpression` here instead.

src/Fantomas.Tests/EditorConfigurationTests.fs
```diff
@@ -503,3 +503,22 @@ fsharp_multiline_bracket_style = cramped
...
+"""
+
+    use configFixture =
+        new ConfigurationFile(defaultConfig, rootDir, content = editorConfig)
+
+    use fsharpFile = new FSharpFile(rootDir)
+
+    let config = EditorConfig.readConfiguration fsharpFile.FSharpFile
+
+    Assert.AreEqual(false, config.NewlineBeforeMultilineComputationExpression)
```
nojaf, line 524: `Assert.IsFalse` please.

src/Fantomas.Core/CodePrinter.fs
```diff
@@ -1914,11 +1916,11 @@ let genClause (isLastItem: bool) (node: MatchClauseNode) =
                      +> ifElse
                          (ctx.Config.ExperimentalKeepIndentInBranch && isLastItem)
                          genKeepIndentInBranch
-                         (autoIndentAndNlnIfExpressionExceedsPageWidthUnlessStroustrup genExpr node.BodyExpr))
+                         (autoIndentAndNlnIfExpressionExceedsPageWidthUnlessSameLinePreferred genExpr node.BodyExpr))
                         ctx)
 
     let genPatAndBody ctx =
-        if ctx.Config.ExperimentalStroustrupStyle && node.BodyExpr.IsStroustrupStyleExpr then
+        if (isStroustrupExpr ctx.Config node.BodyExpr) then
```
nojaf, line 1923: Little nit but you don't need the parentheses here.

nojaf:
This looks good! The only thing I'm really missing is a documentation entry.
I've changed the script to use a local `Fantomas.Core` instead of grabbing it from NuGet.
Could you add a section for the new setting, please?

nojaf (approved):
Great work! I'm very happy you pulled this off with minimal changes.

PR #2747, opened 2023-01-26
Fix handling of AppExpr with a single stroustrup record

josh-degraw:
I'm going through the code in https://github.com/dotnet/docs/pull/33171 to try and get fantomas to match the ideal styles there.

This changes the output of the following code from this:

```fsharp
let newState = {
    Foo =
        Some
            {
                F1 = 0
                F2 = ""
            }
}
```
to this

```fsharp
let newState = {
    Foo =
        Some {
            F1 = 0
            F2 = ""
        }
}
```
Which lets fantomas format code to better match some of the examples there.

But this also brings up a point we've discussed before:

For lowercase function applications with a single record node, with Stroustrup enabled, is there a problem with allowing this style? Because this will also format code like this

```fsharp
let newState =
    notAComputationExpr {
        F1 = 0
        F2 = ""
    }
```

I argue that it's fine, and it's is my personal preference, but @nojaf has expressed valid concerns with it, and I thought we'd came to this agreement but I wanted to confirm before doing too much with it that this was one of the settings we wanted to add for Stroustrup: E.g. `fsharp_disallow_stroustrup_for_lowercase_invocations` (because I think we agree its fine for uppercase applications e.g. discriminated unions like above) or something better named? I'd argue that this output makes sense as a default with Stroustrup enabled, but I definitely see the value of allowing people to opt out.

If we do agree that we want to have a separate setting for this, I'm happy to add that change to this PR.

### DIFF
--- a/src/Fantomas.Core.Tests/Stroustrup/SynBindingValueExpressionTests.fs
+++ b/src/Fantomas.Core.Tests/Stroustrup/SynBindingValueExpressionTests.fs
@@ -654,6 +654,88 @@ let newState =
 """
 
 [<Test>]
+let ``uppercase app node with single record arg with explicit setting`` () =
+    formatSourceString
+        false
+        """
+let newState =
+    someFunc
+        {
+            F1 = 0
+            F2 = ""
+        }
+
+let newState =
+    Some
+        {
+            F1 = 0
+            F2 = ""
+        }
+"""
+        { config with
+            MultilineBracketStyle = Aligned
+            StroustrupForMultilineRecordAsUppercaseInvocationFinalArg = true
+            RecordMultilineFormatter = NumberOfItems }
+    |> prepend newline
+    |> should
+        equal
+        """
+let newState =
+    someFunc
+        {
+            F1 = 0
+            F2 = ""
+        }
+
+let newState =
+    Some {
+        F1 = 0
+        F2 = ""
+    }
+"""
+
+[<Test>]
+let ``lowercase app node with single record arg with explicit setting`` () =
+    formatSourceString
+        false
+        """
+let newState =
+    someFunc
+        {
+            F1 = 0
+            F2 = ""
+        }
+
+let newState =
+    Some
+        {
+            F1 = 0
+            F2 = ""
+        }
+"""
+        { config with
+            MultilineBracketStyle = Aligned
+            StroustrupForMultilineRecordAsLowercaseInvocationFinalArg = true
+            RecordMultilineFormatter = NumberOfItems }
+    |> prepend newline
+    |> should
+        equal
+        """
+let newState =
+    someFunc {
+        F1 = 0
+        F2 = ""
+    }
+
+let newState =
+    Some
+        {
+            F1 = 0
+            F2 = ""
+        }
+"""
+
+[<Test>]
 let ``lowercase app node with single record arg`` () =
     formatSourceString
         false
--- a/src/Fantomas.Core.Tests/Stroustrup/SynBindingValueExpressionTests.fs
+++ b/src/Fantomas.Core.Tests/Stroustrup/SynBindingValueExpressionTests.fs
@@ -598,3 +598,82 @@ let newState = {
         }
 }
 """
+
+[<Test>]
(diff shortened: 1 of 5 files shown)


src/Fantomas.Core.Tests/Stroustrup/SynBindingValueExpressionTests.fs
```diff
@@ -592,3 +592,111 @@ let myRecord = {
     }
 }
 """
+
```
nojaf, line 595: I think this should also work for a function that takes in more parameters:
```fsharp
myFn a b c {
    D = d
}
```

src/Fantomas.Core/CodePrinter.fs
```diff
@@ -1348,7 +1348,9 @@ let genExpr (e: Expr) =
 
                 let longExpression =
                     genExpr node.FunctionExpr
-                    +> indentSepNlnUnindent (col sepNln node.Arguments genExpr)
```
nojaf, line 1351: I actually refactored this code a while ago so you could add a new active pattern for this instead. Something very similar to `(|EndsWithSingleListApp|_|)` but it looks if the last expression is (anon) record and the new setting is on.

nojaf (changes requested):
Hi Josh, good stuff here.

I would like to have a setting for this. I'm not sure casing should be a factor in this. I don't think that is a problem, you either want to enable this or not.

Similar, to how we want the Elmish settings to be separate from the main `Stroustrup`, I would go for a new setting here as well.

src/Fantomas.Core/CodePrinter.fs
```diff
@@ -2172,20 +2172,25 @@ let (|EndsWithSingleListApp|_|) (config: FormatConfig) (appNode: ExprAppNode) =
...
+            visit args
 
+    if config.ExperimentalStroustrupStyle then
         visit appNode.Arguments
+    else
+        match appNode.FunctionExpr with
+        | Expr.Constant _ -> None
+        | ParenExpr _ -> None
+        | UppercaseExpr when config.StroustrupForMultilineRecordAsUppercaseInvocationFinalArg -> visit appNode.Arguments
+        | LowercaseExpr when config.StroustrupForMultilineRecordAsLowercaseInvocationFinalArg -> visit appNode.Arguments
```
nojaf, line 2192: I'm still not convinced about bringing casing into this mix.
Does it really matter when the function name starts with a lower or capital?

josh-degraw: I guess not a ton, the distinction I was picturing was like if you wanted it to apply with Discriminated Union cases e.g. `Some`, where it's clear that it's not a computation expression, but didn't want to use it if it's lowercase to avoid ambiguity between application and computation expressions.

nojaf: The casing won't always tell you that much. 
Consider

```fsharp
open type System.Console

Write {|
    X = 2
|}

let printfObj = printfn "%A"

printfObj {|
    X = 2
|}

type VoidBuilder<'T>() =
    member x.Run (_ : 'T) = ()
    member x.Zero () = ()
    
let VOID = VoidBuilder()

VOID {
    ()
}
```

You cannot truly say what `appNode.FunctionExpr` will be, and if you want to use this style, I think you should go all in.

Perhaps this should not even be its own setting. For `EndsWithDualListApp`, `EndsWithSingleListApp`, I see a use-case (HTML DSL code) to use this outside of Stroustrup.

For `EndsWithSingleRecordApp`, it might just be a part of Stroustrup.

Would it be ok for you to discuss this further during our office hours?

nojaf (changes requested):
Let's discuss this in more detail online.

src/Fantomas.Core/CodePrinter.fs
josh-degraw: Fair points. Honestly I'm fine keeping it one setting. I may have been trying prematurely to optimize against a nonexistant argument haha.

nojaf (approved):
Looks good, thank you!

PR #2748, opened 2023-01-26
Update output for copy-and-update expression for Stroustrup

josh-degraw:
This is to update to match what's currently in the docs proposal.

### DIFF
--- a/src/Fantomas.Core.Tests/AlignedMultilineBracketStyleTests.fs
+++ b/src/Fantomas.Core.Tests/AlignedMultilineBracketStyleTests.fs
@@ -1,4 +1,4 @@
-module Fantomas.Core.Tests.MultilineBlockBracketsOnSameColumnRecordTests
+module Fantomas.Core.Tests.AlignedMultilineBracketStyleTests
 
 open NUnit.Framework
 open FsUnit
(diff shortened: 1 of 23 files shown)


josh-degraw:
It also brings up another point: should `Aligned` brackets be respected for update expressions? Currently it doesn't look like they are but I can't see any technical reason why they couldn't be. Currently my implementation formats some expressions like this if they're on their own line:

```fsharp
{
    astContext with
        IsInsideMatchClausePattern = true
}
```

Which seems like exactly what I would expect if I'd enabled `Aligned`, but fantomas currently formats them like this: 
```fsharp
{ astContext with
    IsInsideMatchClausePattern = true
}
```

I'd be happy to implement that for `Aligned` in a different PR if we want to adopt that.

nojaf (approved):
Hi Josh, great work here!

> It also brings up another point: should Aligned brackets be respected for update expressions?

Yes, this is very deliberate actually. `Aligned` follows the `G-Research`, see https://github.com/G-Research/fsharp-formatting-conventions#formatting-copy-and-update-record-expressions.

I'm ok with your changes, but `Aligned` should remain to work as it now does.
Could you perhaps add a unit test somewhere that highlights this behaviour, that it is to be expected? Maybe two unit tests where both behaviours are captured and some comments on why this is what it is.

nojaf (approved):
Tweaked those last tests a little bit to highlight the difference better.
This is good PR, thank you for these changes!

PR #2749, opened 2023-01-27
Change CodeFormatter.MakePosition return type to pos option

DedSec256:
Since this public method is intended for non-trivial situations, such as use in reflection-based calls

[Rider example](https://github.com/JetBrains/resharper-fsharp/blob/8cb4658df4471c87acb87be998ad6bbbe9e060a3/ReSharper.FSharp/src/FSharp.Fantomas.Host/src/FantomasCodeFormatter.cs#L176-L180)

and reflection calls do not support automatic conversion from `Position` to `FSharpOption<Position>`, it would be more convenient to have `FSharpOptions<Position>` as the return type.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## [Unreleased]
+
+### Changed
+* Rename public API `CodeFormatter.MakePosition -> CodeFormatter.MakeSomePosition` and change its return type to `pos option`. [#2749](https://github.com/fsprojects/fantomas/pull/2749)
+
 ## [6.0.0-alpha-001] - 2023-01-24
 
 ### Changed
--- a/src/Fantomas.Core.Tests/CursorTests.fs
+++ b/src/Fantomas.Core.Tests/CursorTests.fs
@@ -17,7 +17,7 @@ let a =
 """
 
     let formattedResult =
-        CodeFormatter.FormatDocumentAsync(false, source, cursor = CodeFormatter.MakePosition(3, 8))
+        CodeFormatter.FormatDocumentAsync(false, source, ?cursor = CodeFormatter.MakeSomePosition(3, 8))
         |> Async.RunSynchronously
 
     // After formatting the let binding will be on one line
@@ -35,7 +35,7 @@ let a =
 """
 
     let formattedResult =
-        CodeFormatter.FormatDocumentAsync(false, source, cursor = CodeFormatter.MakePosition(3, 7))
+        CodeFormatter.FormatDocumentAsync(false, source, ?cursor = CodeFormatter.MakeSomePosition(3, 7))
         |> Async.RunSynchronously
 
     match formattedResult.Cursor with
--- a/src/Fantomas.Core/CodeFormatter.fs
+++ b/src/Fantomas.Core/CodeFormatter.fs
@@ -32,4 +32,4 @@ type CodeFormatter =
     static member MakeRange(fileName, startLine, startCol, endLine, endCol) =
         Range.mkRange fileName (Position.mkPos startLine startCol) (Position.mkPos endLine endCol)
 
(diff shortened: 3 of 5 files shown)


nojaf:
Hi Alex,

Your example doesn't really illustrate the pain point you are having.
I don't really see why

```csharp
using System.Reflection;

var assembly = Assembly.LoadFile(@"C:\Users\nojaf\Projects\fantomas\src\Fantomas\bin\Debug\net6.0\FSharp.Core.dll");
var optType = assembly.DefinedTypes.First(t => t.Name == "FSharpOption`1");
var optType2 = optType.MakeGenericType(typeof(string)); // imagine Position instead of string ;)
var someMember = optType2.GetMethod("Some", BindingFlags.Public | BindingFlags.Static);
var optPos = someMember.Invoke(null, new[] {"somePosition"});
```
wouldn't work for you. Yes, it involves a little bit of extra work, but that is to be expected when going the reflection route.

`MakeRange` and `MakePosition` make sense to have in our public, as consumers don't need to know about `Fantomas.FCS` and the more confusing `FSharp.Compiler.Text` namespace that actually contains our input types.

`MakeSomePosition` is quite on the noise that one consumer has a more questionable limitation. So I'd like to know more about whether this is really a problem or more some friction on your side. If this really is a problem, I'd be more open to an overload or separate `FormatDocumentWithCursor` if we need to go there.

DedSec256:
> var optType2 = optType.MakeGenericType(typeof(string)); // imagine Position instead of string ;)``

This is where the complexity lies. We don't want to access `Position` directly (since we support the use of Fantomas since version 3.2+), and through reflection we also need to get the `FSharp.Compiler.Text.Position` type from `Fantomas.FCS`.
If it is necessary to explicitly access `Position` for `FSharpOption<Position>` creation, then the `MakePosition` method becomes, possibly, unnecessary for us, since we can get `Position` type through reflection and call its constructor/setters.

> MakeSomePosition is quite on the noise that one consumer has a more questionable limitation.

You are right, we do have stricter limitations. It seemed to me that using `MakeSomePosition` method would not create any problems for other consumers, but it would make life much easier for us (:

nojaf:
I've been giving this some more thought and I believe we should design the `API` centred towards consumption from F# and C#. The current `?cursor` indeed leads to `FSharpOption<Position>` which is not really nice for C# consumption.

Using `[<Optional; DefaultParameterValue<(...)>]` doesn't quite work out for the `FormatConfig` type, so I think we should go with overloads in this case.

```fsharp
    /// <summary>
    /// Format a source string using an optional config.
    /// </summary>
    /// <param name="isSignature">Determines whether the F# parser will process the source as signature file.</param>
    /// <param name="source">F# source code</param>
    static member FormatDocumentAsync: isSignature: bool * source: string -> Async<FormatResult>

    /// <summary>
    /// Format a source string using an optional config.
    /// </summary>
    /// <param name="isSignature">Determines whether the F# parser will process the source as signature file.</param>
    /// <param name="source">F# source code</param>
    /// <param name="config">Fantomas configuration</param>
    static member FormatDocumentAsync: isSignature: bool * source: string * config: FormatConfig -> Async<FormatResult>

    /// <summary>
    /// Format a source string using an optional config.
    /// </summary>
    /// <param name="isSignature">Determines whether the F# parser will process the source as signature file.</param>
    /// <param name="source">F# source code</param>
    /// <param name="config">Fantomas configuration</param>
    /// <param name="cursor">The location of a cursor, zero-based.</param>
    static member FormatDocumentAsync:
        isSignature: bool * source: string * config: FormatConfig * cursor: pos -> Async<FormatResult>
```

`CodeFormatter.MakePosition` does the job there if you need that overload.

DedSec256:
Yes, this is also one of the good solutions. I can close this PR and create a new one to add the overload.

PR #2750, opened 2023-01-27
Consolidate and share logic for Records and AnonymousRecords

josh-degraw:
It's happened several times that something that gets fixed with records that didn't get fixed for anonymous records. Since they share so much in the style guide, I wanted to see if I could refactor `genExpr` to handle them both at the same time so they could share as much logic as possible. I refactored both nodes to use the same field node type, and added an active pattern to join them together and then only handle differences explicitly as needed. It seems to work perfectly to me and I actually found an existing bug that was fixed for records but not anonymous records in the process 🙃.

### DIFF
--- a/src/Fantomas.Core/CodePrinter.fs
+++ b/src/Fantomas.Core/CodePrinter.fs
@@ -465,7 +465,7 @@ let genExpr (e: Expr) =
 
         let genCrampedFields targetColumn =
             match node.CopyInfo with
-            | Some we -> genRecordCopyExpr (genMultilineRecordFieldsExpr node) we
+            | Some we -> genMultilineRecordCopyExpr (genMultilineRecordFieldsExpr node) we
             | None ->
                 fun (ctx: Context) ->
                     col
@@ -1614,8 +1614,14 @@ let genQuoteExpr (node: ExprQuoteNode) =
     +> genSingleTextNode node.CloseToken
     |> genNode node
 
-let genRecordCopyExpr fieldsExpr ci =
-    atCurrentColumnIndent (genExpr ci)
+/// <summary>
+/// Prints the inside of an update record expression.
+/// This function does not print the opening and closing braces.
+/// </summary>
+/// <param name="fieldsExpr">Record fields.</param>
+/// <param name="copyExpr">Expression before the `with` keyword.</param>
+let genMultilineRecordCopyExpr fieldsExpr copyExpr =
+    atCurrentColumnIndent (genExpr copyExpr)
     +> !- " with"
     +> indent
     +> whenShortIndent indent
@@ -1634,6 +1640,11 @@ let genRecordFieldName (node: RecordFieldNode) =
 let genMultilineRecordFieldsExpr (node: ExprRecordBaseNode) =
     col sepNln node.Fields genRecordFieldName
 
+/// <summary>
+/// Print a (anonymous) record with additional information as a single line.
+/// </summary>
+/// <param name="genExtra">Either the `expr with` or `inherit T`.</param>
+/// <param name="node">ExprRecordBaseNode</param>
 let genSmallRecordBaseExpr genExtra (node: ExprRecordBaseNode) =
     genSingleTextNode node.OpeningBrace
     +> addSpaceIfSpaceAroundDelimiter
@@ -1649,17 +1660,36 @@ let genSmallRecordNode (node: ExprRecordNode) =
          | None -> sepNone)
         node
 
+/// <summary>
+/// Print a multiline ExprRecordNode.
+/// <para>
+/// Depending on the actual record, either regular or anonymous,
+/// a different strategy needs to be provided to deal with the record fields in the Cramped style.
+/// This is too avoid offset errors when using a smaller `indent_size`.
+/// </para>
+/// </summary>
+/// <param name="genCrampedFields">
+///<para>
+/// Takes a targetColumn that indicates the column
+/// after the opening brace `{ ` with respect to the `SpaceAroundDelimiter` setting.
+/// </para>
+/// <para>
+/// In `Cramped` style we try to ensure that all record fields are starting at that column.
+/// </para>
+/// </param>
+/// <param name="node">The ExprRecordNode</param>
+/// <param name="ctx">Context</param>
 let genMultilineRecord genCrampedFields (node: ExprRecordNode) (ctx: Context) =
     let expressionStartColumn = ctx.Column
 
     let targetColumn =
-        let openBracketLength = node.OpeningBrace.Text.Length
+        let openBraceLength = node.OpeningBrace.Text.Length
 
         expressionStartColumn
         + (if ctx.Config.SpaceAroundDelimiter then
-               openBracketLength + 1
+               openBraceLength + 1
            else
-               openBracketLength)
+               openBraceLength)
 
     let genMultilineAlignBrackets =
         let fieldsExpr = genMultilineRecordFieldsExpr node
@@ -1671,7 +1701,7 @@ let genMultilineRecord genCrampedFields (node: ExprRecordNode) (ctx: Context) =
                 (fun ctx -> ctx.Config.IsStroustrupStyle)
                 (indent +> sepNln)
                 sepNlnWhenWriteBeforeNewlineNotEmpty // comment after curly brace
-            +> genRecordCopyExpr fieldsExpr ci
+            +> genMultilineRecordCopyExpr fieldsExpr ci
             +> onlyIfCtx (fun ctx -> ctx.Config.IsStroustrupStyle) unindent
             +> sepNln
             +> genSingleTextNode node.ClosingBrace
--- a/src/Fantomas.Core/CodePrinter.fs
+++ b/src/Fantomas.Core/CodePrinter.fs
@@ -1649,7 +1649,6 @@ let genSmallRecordNode (node: ExprRecordNode) =
          | None -> sepNone)
         node
 
-
 let genMultilineRecord genCrampedFields (node: ExprRecordNode) (ctx: Context) =
     let expressionStartColumn = ctx.Column
 
--- a/src/Fantomas.Core/CodePrinter.fs
+++ b/src/Fantomas.Core/CodePrinter.fs
@@ -461,184 +461,36 @@ let genExpr (e: Expr) =
                     isSmallExpression size smallExpression multilineExpression ctx
             |> genNode node
     | Expr.Record node ->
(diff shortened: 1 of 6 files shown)


src/Fantomas.Core/CodePrinter.fs
```diff
@@ -308,6 +308,31 @@ let isSynExprLambdaOrIfThenElse (e: Expr) =
     | Expr.IfThenElse _ -> true
     | _ -> false
 
+let (|RecordOrAnonRecord|_|) (expr: Expr) =
```
nojaf, line 311: I'm not really a fan of this active pattern. It somewhat ruins the magic that you access an exact node type from `SyntaxOak` in `CodePrinter`.
The anonymous records used are also not consistent with anything else in this codebase.
If you were going to make this split, it should be in `SyntaxOak` where we have a shared `Node` for the two when possible.

src/Fantomas.Core/CodePrinter.fs
```diff
@@ -308,6 +308,31 @@ let isSynExprLambdaOrIfThenElse (e: Expr) =
...
+    match expr with
+    | Expr.Record record ->
+        Some
+            {| OpeningBrace = record.OpeningBrace
+               Extra = record.Extra
+               Fields = record.Fields
+               ClosingBrace = record.ClosingBrace
+               Range = record.Range
+               AnonRecord = None
```
nojaf, line 320: This exact field is a code smell on why these types cannot be combined.
There are subtle differences but in essence, one cannot be completely the other.
The code that needs to deal with this type will inevitably need to place `if/else`.

The anonymous record cannot have `inherit` and the regular record cannot have `struct`. That isn't very obvious from this refactor.

nojaf (changes requested):
Hi Josh, I see where you are going with this, but I find it to be somewhat questionable.

I think extracting helper functions like `genCopyExpr` is more interesting than the active pattern.

As mentioned, the types differ enough that combining them no longer makes that obvious.

One alternative I do see is to split the types into different levels.
Having `Expr` like `Record`, `InheritRecord` and `StructAnonRecord`.
Where the `Record` case shares everything that `SynExpr.Record` and `SynExpr.AnonRecord` actually share. And `InheritRecord` and `StructAnonRecord` highlight where they actually differ.

src/Fantomas.Core/CodePrinter.fs
josh-degraw: That makes a lot of sense. My original plan was essentially doing what my active pattern did but with a single shared Node, but I had issues because of something about trivia which led me to try the active pattern instead. But consolidating via Node inheritance makes a ton of sense. My goal was mainly to reduce duplication and make it easier for one change to fix a problem in both, but I also don't want to sacrifice readability or make anything harder to follow. Also didn't love having to have additional cases that failed because the exhaustive check failed. I'll try and refactor to your approach here instead.

src/Fantomas.Core/SyntaxOak.fs
```diff
@@ -766,49 +766,64 @@ type RecordFieldNode(fieldName: IdentListNode, equals: SingleTextNode, expr: Exp
...
+
+    member val OpeningBrace = openingBrace
+    member val Fields = fields
+    member val ClosingBrace = closingBrace
+    member x.HasFields = List.isNotEmpty x.Fields
+
+type ExprCopyableRecordNode
     (
         openingBrace: SingleTextNode,
-        extra: RecordNodeExtra,
```
nojaf, line 772: `RecordNodeExtra` is no longer being used, so please remove that type all together.

src/Fantomas.Core/SyntaxOak.fs
```diff
@@ -766,49 +766,64 @@ type RecordFieldNode(fieldName: IdentListNode, equals: SingleTextNode, expr: Exp
...
+[<AbstractClass>]
+type ExprRecordNode(openingBrace: SingleTextNode, fields: RecordFieldNode list, closingBrace: SingleTextNode, range) =
+    inherit NodeBase(range)
+
+    member val OpeningBrace = openingBrace
+    member val Fields = fields
+    member val ClosingBrace = closingBrace
+    member x.HasFields = List.isNotEmpty x.Fields
+
+type ExprCopyableRecordNode
```
nojaf, line 778: I'm not sure `Copyable` really should be reflected in this name.
All records are copyable, so I really think you can call this one `ExprRegularRecordNode`.
The copy expression is not always there, so I wouldn't let it be a part of the name.

src/Fantomas.Core/SyntaxOak.fs
```diff
@@ -766,49 +766,64 @@ type RecordFieldNode(fieldName: IdentListNode, equals: SingleTextNode, expr: Exp
...
     override val Children: Node array =
         [| yield openingBrace
-           yield! noa (Option.map Expr.Node copyInfo)
+           yield InheritConstructor.Node inheritConstructor
            yield! nodes fields
            yield closingBrace |]
 
+type ExprAnonRecordNode
+    (
+        isStruct: bool,
```
nojaf, line 818: You don't really need the boolean if you were to rename this to `ExprAnonRecordStructNode`.
A non-struct anonymous record would fit perfectly in `ExprCopyableRecordNode`, so you only really need a new type for when it has the `struct` in it.

src/Fantomas.Core/CodePrinter.fs
```diff
@@ -1695,6 +1557,157 @@ let genQuoteExpr (node: ExprQuoteNode) =
...
+                    // See unit test "multiline string before closing brace"
+                    let brace =
+                        addFixedSpaces expressionStartColumn +> genSingleTextNode node.ClosingBrace
+
+                    ifElseCtx lastWriteEventIsNewline brace (addSpaceIfSpaceAroundDelimiter +> brace) ctx)
+            )
+
+    ifAlignOrStroustrupBrackets (genMultilineRecordAlignBrackets node) genMultilineCramped ctx
+
+let genRecord smallRecordExpr multilineRecordExpr (node: ExprRecordNode) =
```
nojaf, line 1706: Add `(ctx: Context)` as a last parameter for `genRecord` instead of returning a lambda.

src/Fantomas.Core/CodePrinter.fs
```diff
@@ -485,135 +460,24 @@ let genExpr (e: Expr) =
...
-                    |> genNode (InheritConstructor.Node ie))
-                   +> onlyIf hasFields sepSemi
-               | RecordNodeExtra.With we -> genExpr we +> !- " with "
-               | RecordNodeExtra.None -> sepNone
-            +> col sepSemi node.Fields genRecordFieldName
-            +> addSpaceIfSpaceAroundDelimiter
-            +> genSingleTextNode node.ClosingBrace
+    | Expr.Record node ->
+        let smallRecordExpr = genSmallCopyableRecordNode node
+        let genCrampedFields = genMultilineStandardRecordExprCrampedFields node
```
nojaf, line 465: `genMultilineStandardRecordExprCrampedFields` is only used in one place. If that is the case, I prefer it to be inside the match clause body. Only re-usable functions should be outside of it.

src/Fantomas.Core/CodePrinter.fs
```diff
@@ -1695,6 +1557,157 @@ let genQuoteExpr (node: ExprQuoteNode) =
     +> genSingleTextNode node.CloseToken
     |> genNode node
 
+let genFieldsExpr (node: ExprRecordNode) =
+    col sepNln node.Fields genRecordFieldName
+
+let private genSmallRecordExpr genExtra (node: ExprRecordNode) =
```
nojaf, line 1563: Because we are using a signature file, every function that is not listed inside the signature file is already private. There is no need to specify it again.

src/Fantomas.Core/CodePrinter.fs
```diff
@@ -485,135 +460,24 @@ let genExpr (e: Expr) =
...
-            onlyIf isStruct !- "struct "
-            +> ifAlignOrStroustrupBrackets genMultilineRecordInstanceAlignBrackets genMultilineRecordInstance
+            genStructPrefix +> genMultilineRecord genCrampedFields node
 
-        fun ctx ->
-            let size = getRecordSize ctx node.Fields
-            genNode node.Node (isSmallExpression size smallRecordExpr multilineRecordExpr) ctx
+        genRecord smallRecordExpr multilineRecordExpr node
+    | Expr.InheritRecord node ->
+        let smallRecordExpr = genSmallInheritRecordExpr node
```
nojaf, line 478: `genSmallInheritRecordExpr` is only used here. Please move the definition up here.

src/Fantomas.Core/CodePrinter.fs
```diff
@@ -485,135 +460,24 @@ let genExpr (e: Expr) =
...
+        let multilineRecordExpr = genMultilineInheritRecordInstance node
```
nojaf, line 479: `genMultilineInheritRecordInstance` is only used here.

src/Fantomas.Core/CodePrinter.fs
```diff
@@ -485,135 +460,24 @@ let genExpr (e: Expr) =
...
-            +> genSingleTextNode node.ClosingBrace
+    | Expr.Record node ->
+        let smallRecordExpr = genSmallCopyableRecordNode node
+        let genCrampedFields = genMultilineStandardRecordExprCrampedFields node
+        let multilineRecordExpr = genMultilineRecord genCrampedFields node
+        genRecord smallRecordExpr multilineRecordExpr node
+    | Expr.AnonRecord node ->
+        let genStructPrefix = onlyIf node.IsStruct !- "struct "
+        let smallRecordExpr = genStructPrefix +> genSmallCopyableRecordNode node
+        let genCrampedFields = genMultilineAnonRecordCrampedFields node
```
nojaf, line 471: `genMultilineAnonRecordCrampedFields` is only used here. Please move it here.

nojaf (changes requested):
A few more nits, but I really like this solution.
Great job!

src/Fantomas.Core/SyntaxOak.fs
josh-degraw: There are still a few differences in non-struct records that I've noticed, see `genMultilineAnonRecordCrampedFields` (which I'll also inline like the other one-offs you mentioned). If those can be consolidated, that would be great, but I wasn't able to get everything to work without those changes

nojaf (approved):
Thanks, this looks good! I've added a couple of XML documents as it can get challenging to find out what each helper function does at the point of invocation.

I do believe we can address the cramped logic even further. I might try it out in a separate PR. Thanks again for all your help!

PR #2755, opened 2023-01-31
Drop `Experimental` prefix from `ExperimentalStroustrup`.

josh-degraw:
﻿Drop `Experimental` prefix from `ExperimentalStroustrup`.

I also updated the configuration docs script to reference the local version to stay in sync.


nojaf (approved):
I like it. I removed `experimental_stroustrup` altogether. No way back.

PR #2756, opened 2023-01-31
Move isStroustrupStyleExpr and isStroustrupStyleType functions

josh-degraw:
Per @nojaf's advice in different PR, I refactored `isStroustrupStyleExpr` and `isStroustrupStyleType` to functions in `Context.fs`.

Those changes felt like a distinct enough change to extract into its own PR.


nojaf (approved):
Looks good! Thanks!

PR #2763, opened 2023-02-03
Add separate CodeFormatter.FormatDocumentAsync overloads with cursor and config

DedSec256:
https://github.com/fsprojects/fantomas/pull/2749#issuecomment-1410435446

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,5 @@
 # Changelog
 
-## [Unreleased]
-
-### Changed
-* Add separate `CodeFormatter.FormatDocumentAsync` overloads with cursor and config. [#2763](https://github.com/fsprojects/fantomas/pull/2763)
-
 ## [6.0.0-alpha-002] - 2023-02-01
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## [Unreleased]
+
+### Changed
+* Add separate `CodeFormatter.FormatDocumentAsync` overloads with cursor and config. [#2763](https://github.com/fsprojects/fantomas/pull/2763)
+
 ## [6.0.0-alpha-002] - 2023-02-01
 
 ### Changed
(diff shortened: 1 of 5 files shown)


src/Fantomas.Core/CodeFormatter.fsi
```diff
@@ -12,15 +12,41 @@ type CodeFormatter =
...
+
+    /// <summary>
+    /// Format a source string using an optional config.
+    /// </summary>
+    /// <param name="isSignature">Determines whether the F# parser will process the source as signature file.</param>
+    /// <param name="source">F# source code</param>
     /// <param name="cursor">The location of a cursor, zero-based.</param>
     static member FormatDocumentAsync:
-        isSignature: bool * source: string * ?config: FormatConfig * ?cursor: pos -> Async<FormatResult>
+        isSignature: bool * source: string * cursor: pos -> Async<FormatResult>
```
nojaf, line 39: This overload seems excessive and I don't see it being used in practice.
If you want to use this then you should use the next one and pass the default config.
Please remove it.

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Changelog
 
+## [Unreleased]
```
nojaf, line 3: There was no way you could have known this but for the non-main branch, I prefer that you don't add a changelog entry. This makes rebasing the v6 branch with the main branch easier. (fewer conflicts).

src/Fantomas/Daemon.fs
```diff
@@ -64,12 +64,15 @@ type FantomasDaemon(sender: Stream, reader: Stream) as this =
 
                 try
                     let! formatResponse =
-                        CodeFormatter.FormatDocumentAsync(
-                            request.IsSignatureFile,
-                            request.SourceCode,
-                            config,
-                            ?cursor = cursor
-                        )
+                        match cursor with
```
nojaf, line 67: Could you swap the order, please?
In this case, I prefer to have the shortest case first.

nojaf (changes requested):

nojaf (approved):
Thanks!

nojaf:
@DedSec256 available in https://www.nuget.org/packages/Fantomas.Core/6.0.0-alpha-003

Issue #2751, opened 2023-01-30
Revisit `--profile` flag


nojaf:
In light of the new logging from https://github.com/fsprojects/fantomas/pull/2693 I think we need to revisit what happens when you `--profile` as an argument.

Currently, the output is full with:

```
[16:04:28 INF] Line count: 204
[16:04:28 INF] Time taken: 00:00:00.3194135 s
[16:04:28 INF] Line count: 120
[16:04:28 INF] Time taken: 00:00:00.0386711 s
[16:04:28 INF] Line count: 1536
[16:04:28 INF] Time taken: 00:00:00.0620313 s
[16:04:28 INF] Line count: 266
[16:04:28 INF] Time taken: 00:00:00.0184182 s
[16:04:28 INF] Line count: 1222
[16:04:28 INF] Time taken: 00:00:00.0366623 s
[16:04:28 INF] Line count: 368
```
without mentioning the name of the file.
Maybe after https://github.com/fsprojects/fantomas/pull/2717 is merged, we can check if we can incorporate the information better.

//cc @dawedawe

dawedawe:
How about adding the following to Logging.fs
```fsharp
/// log a message if given value = true or the verbosity level is >= Detailed
let logIfGivenOrGrEqDetailed b s =
    if b then logger.Information(s) else logger.Debug(s)
```

and calling it in Program.fs like
```fsharp
logIfGivenOrGrEqDetailed profile $"Processing %s{inFile}"
```

nojaf:
I was thinking of maybe using `Spectre.Console` instead.
https://spectreconsole.net/widgets/table caught my eye. I haven't used this NuGet package, but it might be nice to give a table overview of all the processed files.

nojaf:
I think moving the profile logic to `Format.fs` might make sense.
And just add the results to the `FormatResult` somehow. (Or wrap the result in a new type and add the profiling data or something). Then in process the result in `Program.fs` using spectr.

PR #2765, opened 2023-02-05
Add Spectr.Console

nojaf:
Adding Spectr to the project, I think this would be relevant later for https://github.com/fsprojects/fantomas/issues/2751.

image

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -8,7 +8,7 @@ env:
 
 jobs:
   build:
-
+    continue-on-error: true
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest, macOS-latest]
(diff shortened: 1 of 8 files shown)


dawedawe:
Mmh, GitHub Actions seems to have some problem with the color coding:
`3852mFormatted0m  0  Ignored  0  38512mUnchanged0m  0  3859mErrored0m  0`

Besides from that, the PR looks good to me.

But I wonder if it's a good idea to have another output technique besides Serilog?
It does feel somewhat strange to me but I also can't think of a scenario in which it might cause some real issues 🤷‍♂️

nojaf:
> But I wonder if it's a good idea to have another output technique besides Serilog?

We (ab)use Serilog right now predominantly for the difference in log levels.
It would be straightforward to get rid of that and log everything using Spectr.Console.

dawedawe:
Ok, makes sense.
I'll approve as soon as we find the right steps to dance with the tests in github actions...

nojaf:
I'm also having some second thoughts about what we should show when you target a single file.
See this with only 1 file is kinda overkill:

dawedawe (approved):

PR #2769, opened 2023-02-11
Update GeneratingCode.fsx to use the Oak model.

nojaf:
I've updated https://fsprojects.github.io/fantomas/docs/end-users/GeneratingCode.html to work with the `Oak` model rather than the untyped syntax tree.

You might find this interesting @edgarfgp.


docs/docs/end-users/GeneratingCode.fsx
```diff
@@ -32,83 +32,63 @@ In simple scenarios this can work out, but in the long run it doesn't scale well
...
+- [ModuleDecl.TopLevelBinding](../../https://fsprojects.github.io/fantomas/reference/fantomas-core-syntaxoak-moduledecl.html#TopLevelBinding) takes a [BindingNode ](../../reference/fantomas-core-syntaxoak-bindingnode.html).  
 
-  You would have multiple bindings in case of a recursive function.  
-- The `headPat` of binding contains the name and the parameters. 
-- The `expr` ([SynExpr](../../reference/fsharp-compiler-syntax-synexpr.html)) represents the F# syntax expression.
+- The `functionName ` of binding contains the name or is a pattern. 
+- The `expr` ([Expr](../../reference/fantomas-core-syntaxoak-expr.html)) represents the F# syntax expression.
 - Because there is no actual source code, all ranges will be `Range.Zero`.  
 
 The more you interact with AST, the easier you pick up which node represents what.
```
dawedawe, line 87:
```suggestion
The more you interact with AST/Oak, the easier you pick up which node represents what.
```

docs/docs/end-users/GeneratingCode.fsx
```diff
@@ -137,7 +120,7 @@ Parse.parseFile false (SourceText.ofString "let a = 1") []
 The syntax tree can have an overwhelming type hierarchy.  
 We wholeheartedly recommend to use our **[online tool](https://fsprojects.github.io/fantomas-tools/#/ast)** when working with AST.
 
-![F# AST Viewer](../../images/ast-viewer.png)
+![F# AST Viewer](../../images/oak-viewer.png)
 
 This shows you what AST nodes the parser created for a given input text.
```
dawedawe, line 125:
```suggestion
This shows you what Oak nodes the parser created for a given input text.  
```

dawedawe (approved):
I think depending on context it's fine to use both "Oak" and "AST", but I feel like here a some situations in which we should use "Oak" instead of "AST".

nojaf:
Thanks for the review!

docs/docs/end-users/GeneratingCode.fsx
```diff
@@ -32,83 +32,63 @@ In simple scenarios this can work out, but in the long run it doesn't scale well
...
+        [],
+        [ ModuleOrNamespaceNode(
+              None,
+              [ BindingNode(
+                    None,
+                    None,
+                    MultipleTextsNode([ SingleTextNode("let", Range.Zero) ], Range.Zero),
+                    false,
+                    None,
+                    None,
```
josh-degraw, line 52: I know this was already merged and it's totally fine, but I think it would be helpful to use explicit param names here, the waterfall of `None` could feel really confusing for some (myself included lol) and also serve to document what the params actually represent

nojaf: I wouldn't mind accepting a PR for that if you feel strongly about it.
For better or worse you can see the signature when you hover over it:
image

josh-degraw: Ooh I did not know that, that's super useful! I think it would still be helpful to have param names just to illustrate what the parameters are in examples but that's good to know!

nojaf: As I said 😉

> I wouldn't mind accepting a PR

PR #2770, opened 2023-02-13
Move profile logic to Format.fs and use Spectre

### LINKED ISSUE #2751: Revisit `--profile` flag

dawedawe:
fixes #2751

- Move profile logic to `Format.fs`
- Add profiling data to `FormatResult`
- Use Spectre to render the results

image

### DIFF
--- a/src/Fantomas.Core/CodeFormatterTypes.fs
+++ b/src/Fantomas.Core/CodeFormatterTypes.fs
@@ -1,6 +1,9 @@
 ﻿namespace Fantomas.Core
 
 open FSharp.Compiler.Text
+open System
+
+type ProfileInfos = { LineCount: int; TimeTaken: TimeSpan }
 
 type FormatResult =
     {
@@ -9,4 +12,5 @@ type FormatResult =
         /// New position of the input cursor.
         /// This can be None when no cursor was passed as input or no position was resolved.
         Cursor: pos option
+        ProfileInfos: ProfileInfos option
     }
--- a/src/Fantomas.Core/CodeFormatterTypes.fs
+++ b/src/Fantomas.Core/CodeFormatterTypes.fs
@@ -1,9 +1,6 @@
 ﻿namespace Fantomas.Core
 
 open FSharp.Compiler.Text
-open System
-
-type ProfileInfos = { LineCount: int; TimeTaken: TimeSpan }
 
 type FormatResult =
     {
@@ -12,5 +9,4 @@ type FormatResult =
         /// New position of the input cursor.
         /// This can be None when no cursor was passed as input or no position was resolved.
         Cursor: pos option
-        ProfileInfos: ProfileInfos option
     }
--- a/src/Fantomas/Format.fs
+++ b/src/Fantomas/Format.fs
@@ -5,8 +5,8 @@ open System.IO
 open Fantomas.Core
 
 type FormatResult =
-    | Formatted of filename: string * formattedContent: string
-    | Unchanged of filename: string
+    | Formatted of filename: string * formattedContent: string * profileInfos: ProfileInfos option
+    | Unchanged of filename: string * profileInfos: ProfileInfos option
     | InvalidCode of filename: string * formattedContent: string
     | Error of filename: string * formattingError: Exception
     | IgnoredFile of filename: string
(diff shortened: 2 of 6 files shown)


src/Fantomas.Core/CodeFormatterTypes.fs
```diff
@@ -1,6 +1,9 @@
 ﻿namespace Fantomas.Core
 
 open FSharp.Compiler.Text
+open System
+
+type ProfileInfos = { LineCount: int; TimeTaken: TimeSpan }
```
nojaf, line 6: No changes should be more to Fantomas.Core for this feature.
It was always a command line-only thing and I would like to keep it that way for now.

src/Fantomas/Format.fs
```diff
@@ -5,15 +5,16 @@ open System.IO
 open Fantomas.Core
 
 type FormatResult =
-    | Formatted of filename: string * formattedContent: string
-    | Unchanged of filename: string
+    | Formatted of filename: string * formattedContent: string * profileInfos: ProfileInfos option
```
nojaf, line 8: I don't think `ProfileInfos` would be plural here. `ProfileInfo` seems more correct.

src/Fantomas/Format.fs
```diff
@@ -42,16 +62,17 @@ let private formatContentInternalAsync
...
-                    return Unchanged(filename = file)
+                    return Unchanged(filename = file, profileInfos = profileInfos)
             with ex ->
                 return Error(file, ex)
         }
 
 let formatContentAsync = formatContentInternalAsync false
 
-let private formatFileInternalAsync (compareWithoutLineEndings: bool) (file: string) =
+let private formatFileInternalAsync (compareWithoutLineEndings: bool) (profile: bool) (file: string) =
```
nojaf, line 75: We might want to consider bundling all the parameters into a single record.
It is slowly becoming a boolean fest.

src/Fantomas/Program.fs
```diff
@@ -411,14 +388,40 @@ let main argv =
 
             elog $"Failed to format file: {file}{message}"
 
+        let reportProfileInfos results =
```
nojaf, line 391: You should pass the `oks` and `unchanged` from `let oks, ignored, unchanged, errored = partitionResults results` here.

src/Fantomas/Program.fs
```diff
@@ -64,9 +55,9 @@ type OutputPath =
 
 [<RequireQualifiedAccess>]
 type ProcessResult =
-    | Formatted of string
+    | Formatted of string * ProfileInfos option
```
nojaf, line 58: Shower thought: do we need this difference between `FormatResult` and `ProcessResult`?
Perhaps we can just use `FormatResult` instead?

nojaf (changes requested):
The result looks great but I would structure things a bit differently.

src/Fantomas.Core/CodeFormatterTypes.fs
dawedawe: Oops, yeah, changed that.

src/Fantomas/Format.fs
dawedawe: Done

src/Fantomas/Program.fs
dawedawe: Yep, seems like we can.

src/Fantomas/Program.fs
dawedawe: Done, as you already saw.

src/Fantomas/Format.fs
dawedawe: I tried to come up with something and hope it makes somewhat sense.

PR #2774, opened 2023-02-17
Cursor with defines

nojaf:
This is related to https://github.com/fsprojects/fantomas/issues/2727 and tries to tackle the scenario where we format the code multiple times and merge each back into a single result.

For example:
```fsharp
let v =
  #if A
    'A'
  #else
    #if B
      'B'
    #else
      'C'
    #endif 
  #endif
```

Fantomas will format this three times: 
- `[]`
- `[ "A" ]`
- `[ "B" ]`

In this PR, I'm trying to merge all the results in one go. In the `main` branch we are reducing the results instead, so I think for large files this will be quite beneficial in terms of performance.

@dawedawe and @josh-degraw could you please carefully review this PR? The goal was to make everything a bit more understandable than how it works today. I'm not quite sure I've hit the mark just yet.

There are some TODOs left, please let me know what you would do in those cases. I'm not quite sure about everything just yet.


src/Fantomas.Core/MultipleDefineCombinations.fs
```diff
@@ -0,0 +1,284 @@
...
+open System.Text.RegularExpressions
+open Microsoft.FSharp.Core.CompilerServices
+open FSharp.Compiler.Text
+
+/// A CodeFragment represents a chunk of code that is either
+///     a single conditional hash directive line,
+///     non existing content (for a specific combination of defines) or
+///     active content.
+///
+/// When the code need to be compared, a CustomComparison is used to determine which fragment we are interested in.
```
dawedawe, line 15:
```suggestion
/// When the code needs to be compared, a CustomComparison is used to determine which fragment we are interested in.
```

src/Fantomas.Core/MultipleDefineCombinations.fs
```diff
@@ -0,0 +1,284 @@
...
+[<RequireQualifiedAccess>]
+[<CustomEquality; CustomComparison>]
+type CodeFragment =
+    /// Any line that starts with `#if`, `#else` or `#endif`
+    | HashLine of line: string * defines: DefineCombination
+    /// Content found between two HashLines
+    | Content of code: string * lineCount: int * defines: DefineCombination
+    /// When two HashLine follow each other without any content in between.
```
dawedawe, line 23:
```suggestion
    /// When two HashLines follow each other without any content in between.
```

src/Fantomas.Core/MultipleDefineCombinations.fs
```diff
@@ -0,0 +1,284 @@
...
+
+    interface IComparable<CodeFragment> with
+        member x.CompareTo y =
+            match x, y with
+            // When comparing the different results of each format result, the single constant is that all hash lines
+            // should exactly match.
+            | CodeFragment.HashLine(line = lineX), CodeFragment.HashLine(line = lineY) ->
+                assert (lineX = lineY)
+                0
+            // Pick the other fragment is it has content you don't
```
dawedawe, line 55:
```suggestion
            // Pick the other fragment if it has content you don't
```

src/Fantomas.Core/MultipleDefineCombinations.fs
```diff
@@ -0,0 +1,284 @@
...
+            | CodeFragment.NoContent _, CodeFragment.Content _ -> -1
+            // Pick our fragment is the other fragment has no code
```
dawedawe, line 57:
```suggestion
            // Pick our fragment if the other fragment has no code
```

src/Fantomas.Core/MultipleDefineCombinations.fs
```diff
@@ -0,0 +1,284 @@
...
+            | CodeFragment.Content _, CodeFragment.NoContent _ -> 1
+            // If both fragments are empty they are equivalent.
+            // Keep in mind that we could be comparing more the two fragments at the same time in `traverseFragments`
```
dawedawe, line 60:
```suggestion
            // Keep in mind that we could be comparing more than the two fragments at the same time in `traverseFragments`
```

src/Fantomas.Core/MultipleDefineCombinations.fs
```diff
@@ -0,0 +1,284 @@
...
+                let hasOwnContent = not (String.IsNullOrWhiteSpace ownContent)
+                let hasOtherContent = not (String.IsNullOrWhiteSpace otherContent)
+
+                if ownLineCount > otherLineCount then
+                    1
+                elif ownLineCount < otherLineCount then
+                    -1
+                elif
+                    String.IsNullOrWhiteSpace ownContent
+                    && not (String.IsNullOrWhiteSpace otherContent)
```
dawedawe, line 74: Isn't this the same as line 79?: `elif not hasOwnContent && hasOtherContent then`

src/Fantomas.Core/MultipleDefineCombinations.fs
```diff
@@ -0,0 +1,284 @@
...
+                    let ownLength = ownContent.Length
+                    let otherLength = otherContent.Length
+
+                    if ownLength < otherLength then 1
+                    elif ownLength > otherLength then -1
+                    else 0
+            // This is an unexpected situation.
+            // You should never enter the case where you need to compare a hash line with something other than a hash line.
+            | x, other ->
+                // TODO: throw custom exception?
```
dawedawe, line 94: Mmh, as of now I don't see that much of an advantage with a custom exception for this, so I wouldn't bother.

src/Fantomas.Core/MultipleDefineCombinations.fs
```diff
@@ -0,0 +1,284 @@
...
+
+/// Split the given `source` into the matching `CodeFragments`.
+let splitWhenHash (defines: DefineCombination) (newline: string) (source: string) : CodeFragment list =
+    let lines = source.Split([| newline |], options = StringSplitOptions.None)
+    let mutable fragmentsBuilder = ListCollector<CodeFragment>()
+
+    let closeState (acc: SplitHashState) =
+        if acc.LastLineInfo = LastLineInfo.Content then
+            let lastFragment = acc.CurrentBuilder.ToString()
+            // The last fragment could be a newline after the the last #endif
```
dawedawe, line 137:
```suggestion
            // The last fragment could be a newline after the last #endif
```

src/Fantomas.Core/MultipleDefineCombinations.fs
```diff
@@ -0,0 +1,284 @@
...
+    if not allHaveSameFragmentCount then
+        let chunkReport =
+            allInFragments
+            |> List.map (fun result ->
+                sprintf "[%s] has %i fragments" (String.concat ", " result.Defines.Value) result.Fragments.Length)
+            |> String.concat config.EndOfLine.NewLineString
+
+        raise (
+            FormatException(
+                $"""Fantomas is trying to format the input multiple times due to the detect of multiple defines.
```
dawedawe, line 200:
```suggestion
                $"""Fantomas is trying to format the input multiple times due to the detection of multiple defines.
```

dawedawe (changes requested):
From my perspective it's definitely going in the right direction 👍

src/Fantomas.Core/CodeFormatterImpl.fs
```diff
@@ -4,7 +4,8 @@ module internal Fantomas.Core.CodeFormatterImpl
 open FSharp.Compiler.Diagnostics
 open FSharp.Compiler.Syntax
 open FSharp.Compiler.Text
-open Fantomas.Core.SyntaxOak
+open SyntaxOak
```
dawedawe, line 7: open Statement not needed
```suggestion
```

dawedawe (approved):

src/Fantomas.Core/MultipleDefineCombinations.fs
```diff
@@ -35,8 +35,28 @@ type CodeFragment =
...
+        match other with
+        | :? CodeFragment as other ->
+            match x, other with
+            | HashLine _ as x, (HashLine _ as other) -> x = other
+            | Content _ as x, (Content _ as other) -> x = other
+            | NoContent _ as x, (NoContent _ as other) -> x = other
+            | _ -> false
+        | _ -> false
+
+    override x.GetHashCode() =
```
nojaf, line 48: @dawedawe I didn't really know how to best deal with this one.
https://learn.microsoft.com/en-us/dotnet/api/system.hashcode.combine?view=net-7.0 doesn't seem to be a `netstandard2.0` thing.

Let me know if you have any other suggestions.

dawedawe: No, that seems really good enough for me 👍

Issue #2768, opened 2023-02-11
[Feature request] Invert --recursive flag


nojaf:
I propose we change the current `-r` (`--recursive`) flag to be enabled by default.
Change it to `--no-recurse` if people want to disable this.

The existing way of Fantomas deals with this problem is you always need to pass `-r`, while I find myself rarely in the position of leaving that flag out.

## Pros and Cons 

The advantages of making this adjustment to Fantomas are that it might feel more natural to users and would be slightly more ergonomic.

The disadvantage of making this adjustment to Fantomas is that well-known default behaviour is changed.

## Examples

Right now I have `dotnet fantomas src -r`.
After this feature, it would be `dotnet fantomas src`.

If I don't want to format the folder recursively: `dotnet fantomas src --no-recurse`

## Extra information

Estimated cost (XS, S, M, L, XL, XXL): S

Related suggestions: (put links to related suggestions here)

## Affidavit (please submit!)

Please tick this by placing a cross in the box:
* [x] I have read the [Contribution Guidelines](https://github.com/fsprojects/fantomas/blob/main/CONTRIBUTING.md).
* [x] I have [searched both open and closed suggestions on this site](https://github.com/fsprojects/fantomas/issues) and believe this is not a duplicate

Please tick all that apply:
* [ ] This is not a breaking change to Fantomas
* [x] I or my company would be willing to help implement and/or test this
* [ ] This suggestion is part of the [Microsoft style guide](https://docs.microsoft.com/en-us/dotnet/fsharp/style-guide/formatting) (please add a link to section if so)
* [ ] This suggestion is part of the [G-Research style guide](https://github.com/G-Research/fsharp-formatting-conventions) (please add a link to section if so)

dawedawe:
I'd like the change a lot. And maybe we can even go a step further.

Is there even a real demand to have --no-recurse? To me, it seems a bit strange to call Fantomas on a folder and only wanting the files in the folder to be processed but not the folders below. Maybe having `.fantomasignore` is enough and there's no need for more options in that area 🤷

nojaf:
True, would there be any scenario where you couldn't get away with a `.fantomasignore` entry?

nojaf:
Available in [v6](https://www.nuget.org/packages/fantomas/6.0.0).

PR #2775, opened 2023-02-17
Always process folder recursive.

nojaf:
See #2768


dawedawe:
Nice!
I think, we can also remove the flag from `regressions.fsx` and `IgnoreFilesTests.fs`.

dawedawe (approved):

PR #2781, opened 2023-02-25
Extract Stroustrup for final list argument into a separate setting.

nojaf:
This carves out the existing Stroustrup style for the last (two) list/array argument(s).

### DIFF
--- a/docs/docs/end-users/Configuration.fsx
+++ b/docs/docs/end-users/Configuration.fsx
@@ -869,6 +869,45 @@ formatCode
 (*** include-output ***)
 
 (**
+<fantomas-setting name="fsharp_stroustrup_final_list_arguments" orange></fantomas-setting>
+
+Applies the Stroustrup style to the final (two) array or list argument(s) in a function application.
+
+Default = false
+*)
+
+formatCode
+    """ 
+let dualList =
+    div
+        []
+        [
+            h1 [] [ str "Some title" ]
+            ul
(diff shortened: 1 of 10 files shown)


dawedawe (approved):

nojaf:
Hi @josh-degraw, I'd like to hear your thoughts on this one as well.
Could I get a review?

josh-degraw:
> Hi @josh-degraw, I'd like to hear your thoughts on this one as well. Could I get a review?

Sorry, just saw this today.

One thing I'm not sure of is that this always applies to the final 2 arguments. Personally (having not used elmish at all), I would really only want to apply this style for the last argument only, if it applies. Would it be possible to modify the setting to specify if you want stroustrup for the last 1, 2, or 0 lists? 
e.g. 

```fsharp

//...
StroustrupFinalListArguments: bool
```

Also this doesn't change anything about record applications, right?

nojaf:
> Also this doesn't change anything about record applications, right?

Nope, just arrays.

> Would it be possible to modify the setting to specify if you want stroustrup for the last 1, 2, or 0 lists?

The two popular DSL, in this case, are `Fable.React`(uses two lists) and `Feliz` (uses one list) and you would frequently mix these anyway.

I'm against having a DU for this as nobody is asking for it. Premature optimization and all that.

josh-degraw (approved):

PR #2790, opened 2023-03-07
Use Stroustrup for list applications if enabled globally or with specific setting

### LINKED ISSUE #2789: List/array applications should use stroustrup if global stroustrup setting is enabled
I was just trying to update my codebase to the latest alpha and it took me much longer than I'd like to admit to track down the source of this problem. I think it makes sense to have it be a separate setting, but I think it should absolutely be the default style if Stroustrup is enabled.

Simplified code example 
```f#
module internal Plan =
    type Record = 
        {
            //
            Foo: string
            Bar: string
        }
    let canDoSomething =
        Validation.any
            [
                Validation.someValidationLogicHere
                Validation.someMoreValidationLogicHere
                Validation.evenMoreLomethingOrOther
            ]
```

```f#
module internal Plan =
    type Record = {
        //
        Foo: string
        Bar: string
    }

    let canDoSomething =
        Validation.any
            [
                Validation.someValidationLogicHere
                Validation.someMoreValidationLogicHere
                Validation.evenMoreLomethingOrOther
            ]
```

Note that the multiline bracket style is still respected if `Cramped` is specified:

```f#
module internal Plan =
    type Record =
        {
          //
          Foo: string
          Bar: string }

    let canDoSomething =
        Validation.any
            [ Validation.someValidationLogicHere
              Validation.someMoreValidationLogicHere
              Validation.evenMoreLomethingOrOther ]
```

I think how this should work is that if you have _either_ Stroustrup enabled globally _or_ you have the function application specific setting, it should use Stroustrup.

We could also have it so that if you have Stroustrup enabled globally and `fsharp_stroustrup_final_list_arguments` set to false, then it doesn't use it there, although that would require a little more tweaking, it's certainly possible.

josh-degraw:
Closes #2789


nojaf:
Superseded by https://github.com/fsprojects/fantomas/pull/2795

PR #2795, opened 2023-03-17
EndsWithSingleListApp & EndsWithDualListApp are enable by IsStroustrupStyle as well.

nojaf:
As discussed with @josh-degraw.

### DIFF
--- a/src/Fantomas.Core.Tests/Stroustrup/ExperimentalElmish.fs
+++ b/src/Fantomas.Core.Tests/Stroustrup/ExperimentalElmish.fs
@@ -1,4 +1,4 @@
-﻿module Fantomas.Core.Tests.Stroustrup.FinalListArgumentsTests
+﻿module Fantomas.Core.Tests.Stroustrup.ExperimentalElmish
 
 open NUnit.Framework
 open FsUnit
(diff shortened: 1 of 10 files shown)


src/Fantomas.Core.Tests/Stroustrup/ExperimentalElmish.fs
josh-degraw, line 1: Should we not name this file `ExperimentalElmishTests.fs`?

src/Fantomas.Core/FormatConfig.fs
```diff
@@ -227,7 +227,7 @@ type FormatConfig =
 
       [<Category("Convention")>]
       [<DisplayName("Applies the Stroustrup style to the final (two) array or list argument(s) in a function application")>]
```
josh-degraw, line 229: Not sure either way, but should we change anything about this displayName to mention elmish or do you think it's fine to stay as is??

nojaf: I think this is fine for now. We can revisit this later.
Little secret, but somebody is actually doing anything with these attributes anyway.
We expose them via `Fantomas.Client` so IDEs can request them but none of them has implemented this.

src/Fantomas.Core.Tests/Stroustrup/ExperimentalElmish.fs
nojaf: Good catch

josh-degraw (approved):

PR #2799, opened 2023-03-25
Return string from FormatASTAsync api.

nojaf:
I think it makes sense to align the return type of `FormatASTAsync` with `FormatOakAsync`.
As the cursor is never passed into these methods, it would always be empty in the result anyway.


dawedawe (approved):
Makes a lot of sense.

PR #2813, opened 2023-03-29
update the dotnet sdk to 7.0.202

### LINKED ISSUE #2812: Update SDK to 7.0.200
We should update the dotnet SDK to `7.0.20X`.
Be aware of https://learn.microsoft.com/en-us/dotnet/core/compatibility/sdk/7.0/solution-level-output-no-longer-valid.

dawedawe:
Fixes #2812 
As Chet pointed out, it's fine to just update to 7.0.202 without changing the way we pack.


nojaf (approved):

nojaf:
Please also follow-up on fantomas-tools for this. Best to bump there as well.

dawedawe:
If I understand things right, we should also go back to the single version in regressions.yml, don't we?

nojaf:
The would be a maybe, I'm not quite sure. You can try running the regression workflow from this branch, see if it sticks.

dawedawe:
Turns out, we need the older version for the regression test on Fantomas itself. At least for the time being with the configured git hash. With that, the regression tests all pass: https://github.com/dawedawe/fantomas/actions/runs/4556893232

PR #2852, opened 2023-04-19
Fix issues with multiline generic type parameters

### LINKED ISSUE #2706: Aligned bracket style in anonymous record is not respected
Issue created from [fantomas-online](https://fsprojects.github.io/fantomas-tools/#/fantomas/main?data=N4KABGBEDGD2AmBTSAuKAbRAXMAHATgJYBuAhlomKQM4BS1sAdmABSn4C2aAkgIKcAlRAwCu%2BaIgCUYALwAdfHOYQqnAHT0mAWQSJ0CpSrAAfAHxg4jYonxYAKrAA8wYweVGIAa0KN4aalhEjADmbh4qBLC4NliEwmguYAHkhAGE0NQAooykAEaYfmC5sLDoJgC%2BYSrG5aaQADTgUNTYsSHUqGAA2k0QoOGQACRYAJ7RnZA%2BWA29KkNk6CLIaD3hEAAMjWtQ3L6IjFgAyoQAXshbawAssxAAurPlF303UMNjy1BTMwODC0udqzWAEYnh5IFpSAAPAAyPkQ0P2wSwAAtvsCAEzrF73IyPWb9MFvcZoSD7eAAeQAZrDGIhDqNMGijPNSIsPoDwujQczshTqXCmWD0JTINiHqCCcyiR9IMVSoKICy2QCXhAAMzcua7Fq2ABiPlZADlEAB3dACzUQQJLMW4iUvIajYlQOXoBWvP7s1VgS6WqCHXCkCQAIUQlNg%2BEQAAV2KQONgbO6rfgbeEcSo8UZJXNpRNXe6lf8Vt6AKx%2ByABoOIUPhyPQ2AmmzQGiIXbEWDN2JMJNga2IW0Z%2B0-J0y-Oawte7YANnLlZDYYjiAAqrhouIW22Oylu37KayWgOIJmVNnFbmSWOHb9WUXut6AOyzwPz2uIADC6Bo1DfTACKegWARj2e7oAeabiviV4jnmJRuuO17KsW2wABxPlWNaLloiAcLkia7vu-bgXakHDu8MHyvBnoqtsACcaEvouP7oDu3ogWBHjpkeQ6EtBF6wQWCG3hyHhAps3oVs%2BiC8JSFD4D%2BHAcKQPZ9oeYDHs8pHOrK-GUTek7AiC4lztWC6Roc2HpKULHbGxhEcRBWZQWRfEUVeVFIRi9FSTJNjmRwlnMYwykpnZRicWp3FSrxLo6W5enUcCGpGZJ-CwCIvgACJ6IQ-mycFqb2cRjmaTKXy6Yhd7bECvriRCkLcJSdjIvshzIhGWAAOqEPAKI9liRGDiRPHOZ8BwCe5lXAmWtVQg1TX7JkoF0m1thdT1qJ%2BlO-WFYNxXDVpZVxRVwlGECM4zfVjCUoQkLkmu5ARpkkIEMI1CENZazIdtYUOSeTkHWN5VCd6QKPhdQhwPg8Brb1fqXN9KjhepYCnq80WTIDR3A1VqHg4gkPwIaIg4TYVLcBQHAdH6QKqcjqOOiNkAcCI6CxOatK6hGilYLJdhkUD%2BnhECdHiRDEbwFoLNs3CnOcOQeXltAyKxgBNgAPomt1vW05FObo4dPwTSdKiYuWdX8PgpAjOS%2BCwgEMMbd6X060NUWMwbhJG966KGds4JQhbVs23bWBEyT%2BBkxTVMgy7e1u1pzOs4Q7OILL3O8-zWOCx46JcuJgfW7bqRYJLScp2n8t4eJSsq7JGta6isd-SVEwe1KXvbOiSV%2B3VABqenBj48A%2BMEDs9s7A1ca7evu5jhvxR5nI1T3UK6ulAHvYwg%2B%2BCPY9wwjdy-Rp%2B2lXPnsL5NnLTSvkIZbAWAAOLYE9L3UG9TB707B9gEjutnujl557HW9udP2xozRwlDFgRs%2Bw%2BbRCytdRghAuyMF4L4LC4do7bBUpPCK09-6M0AefYBncwZ%2B14OaYIjA16MA3kwY4VDyBiEQA4XYSADjbiCvhJaTdj7x1HLFIB2M1jolxuQ1mNgcixGsPWEIGDcL4AQT4ZBm8sFrFsrwlG-0BGuSEdnIw6IRY9yloQGk8I4y5HgKQD8sA3ohHASnYCBFNH03PDFXRxDhHhDVGJP2L8bA5X2FgVkABpRAiBcBsKCbsYMltaGOxss43BdNtHkTglnBK3jfZrEgMGT8jBPBmOoKldKhNhAUAliYlOL9Ixv1UflUKiMj5aJbi5dJejMkeDVHnP2wZ2AYUjBlVI0AiD%2BSkYgeAS4kFMCytAT8lsUFOJ4ckv%2BaNGaJ2lrSWJQZPDYHpCMRkAtOlGDVN3HJpdNnVkttAXZRwGTnHEqyQgVCJmihWfgtZANphHMXl05eOSwkRLqmHBRVI8mkAKUUnsolv6-w%2BQzLSRD24X2Nuqa%2BOT6REAAjoJASz2I-VxE0W4FxJjUF1G9ToGjyhAA)

#### Code

```fsharp
let private asJson (arm: IArmResource) =
    arm.JsonModel
    |> convertTo<{|
        kind: string
        properties: {| statisticsEnabled: bool |}
    |}>
```
            

#### Result

```fsharp
let private asJson (arm: IArmResource) =
    arm.JsonModel
    |> convertTo<{|
        kind: string
        properties: {| statisticsEnabled: bool |}
    |} >

```
            
#### Problem description

I'm not entirely sure this is due to recent changes. I wouldn't claim this ever worked.
I'm guessing the result should be something like:
```fsharp
    arm.JsonModel
    |> convertTo<
           {|
               kind: string
               properties: {| statisticsEnabled: bool |}
           |} 
       >
```

#### Extra information

- [ ] The formatted result breaks my code.

#### Options

Fantomas main branch at 2023-01-12T20:47:35Z - b6f7ecfd364a4db5a67888acc3248c37f1424c37

```fsharp
    { config with
                MultilineBracketStyle = aligned }
```

<sub>Did you know that you can ignore files when formatting from fantomas-tool or the FAKE targets by using a [.fantomasignore file](https://fsprojects.github.io/fantomas/docs/end-users/IgnoreFiles.html)?</sub>

josh-degraw:
Closes #2706, from Amplifying F# session

### DIFF
--- a/src/Fantomas.Core.Tests/DallasTests.fs
+++ b/src/Fantomas.Core.Tests/DallasTests.fs
@@ -1515,11 +1515,14 @@ let autoCompleteItems: cmap<DeclName, DeclarationListItem * Position * string<Lo
         equal
         """
 let autoCompleteItems
-    : cmap<DeclName, DeclarationListItem *
-      Position *
-      string<LocalPath> *
-      (Position -> option<string>) *
-      FSharp.Compiler.Syntax.ParsedInput> =
+    : cmap<
+          DeclName,
+          DeclarationListItem *
+          Position *
+          string<LocalPath> *
+          (Position -> option<string>) *
+          FSharp.Compiler.Syntax.ParsedInput
+       > =
     cmap ()
 """
 
--- a/src/Fantomas.Core.Tests/TypeAnnotationTests.fs
+++ b/src/Fantomas.Core.Tests/TypeAnnotationTests.fs
@@ -262,15 +262,18 @@ let private asJson (arm: IArmResource) =
         >
 """
 
-[<Test>]
-let ``type application including nested multiline function type`` () =
+let alignedMaxLine30 =
     { config with
         MaxLineLength = 30
         MultilineBracketStyle = Aligned }
-    |> formatSourceString
+
+[<Test>]
+let ``type application including nested multiline function type`` () =
+    formatSourceString
         """
 let bv = unbox<Foo<'innerContextLongLongLong, 'bb -> 'b>> bf
-        """
+"""
+        alignedMaxLine30
     |> prepend newline
     |> should
         equal
@@ -287,19 +290,7 @@ let bv =
 
 [<Test>]
 let ``multiline type argument with AppLongIdentAndSingleParenArg`` () =
-    { config with
-        MaxLineLength = 30
-        SpaceBeforeUppercaseInvocation = true
-        SpaceBeforeClassConstructor = true
-        SpaceBeforeMember = true
-        SpaceBeforeColon = true
-        SpaceBeforeSemicolon = true
-        MultilineBracketStyle = Aligned
-        AlignFunctionSignatureToIndentation = true
-        AlternativeLongMemberDefinitions = true
-        MultiLineLambdaClosingNewline = true
-        NewlineBetweenTypeDefinitionAndMembers = false }
-    |> formatSourceString
+    formatSourceString
         """
 path.Replace<
         Foo<
@@ -308,6 +299,18 @@ path.Replace<
          >
      >("../../../", "....")
 """
+        { config with
+            MaxLineLength = 30
+            SpaceBeforeUppercaseInvocation = true
+            SpaceBeforeClassConstructor = true
+            SpaceBeforeMember = true
+            SpaceBeforeColon = true
+            SpaceBeforeSemicolon = true
+            MultilineBracketStyle = Aligned
+            AlignFunctionSignatureToIndentation = true
+            AlternativeLongMemberDefinitions = true
+            MultiLineLambdaClosingNewline = true
+            NewlineBetweenTypeDefinitionAndMembers = false }
     |> prepend newline
     |> should
         equal
@@ -325,11 +328,7 @@ path.Replace<
 
 [<Test>]
 let ``multiline type argument with AppLongIdentAndSingleParenArg 2`` () =
-    { config with
-        MaxLineLength = 30
-        SpaceBeforeClassConstructor = true
-        MultilineBracketStyle = Aligned }
-    |> formatSourceString
+    formatSourceString
         """
 path.Replace<
         Foo<
@@ -338,6 +337,8 @@ path.Replace<
          >
      >("../../../", "....")
 """
+        { alignedMaxLine30 with
+            SpaceBeforeClassConstructor = true }
     |> prepend newline
     |> should
         equal
@@ -355,11 +356,7 @@ path.Replace<
 
 [<Test>]
 let ``multiline type argument with AppLongIdentAndSingleParenArg 3`` () =
-    { config with
-        MaxLineLength = 30
-        SpaceBeforeClassConstructor = false
-        MultilineBracketStyle = Aligned }
-    |> formatSourceString
+    formatSourceString
         """
 path.Replace<
         Foo<
@@ -368,6 +365,7 @@ path.Replace<
          >
      >("../../../", "....")
 """
+        alignedMaxLine30
     |> prepend newline
     |> should
         equal
@@ -385,10 +383,7 @@ path.Replace<
 
 [<Test>]
 let ``multiline type argument with AppSingleParenArg`` () =
-    { config with
-        MaxLineLength = 30
-        MultilineBracketStyle = Aligned }
-    |> formatSourceString
+    formatSourceString
         """
 someFunc<
         Foo<
@@ -397,6 +392,7 @@ someFunc<
          >
      >(a,b)
 """
+        alignedMaxLine30
     |> prepend newline
     |> should
         equal
@@ -414,10 +410,7 @@ someFunc<
 
 [<Test>]
 let ``multiline type argument with AppWithLambda`` () =
-    { config with
-        MaxLineLength = 30
-        MultilineBracketStyle = Aligned }
-    |> formatSourceString
+    formatSourceString
         """
 someFunc<
     Bar<
@@ -426,6 +419,7 @@ someFunc<
      >
  > (fun x -> x)
 """
+        alignedMaxLine30
     |> prepend newline
     |> should
         equal
@@ -441,10 +435,7 @@ someFunc<
 
 [<Test>]
 let ``multiline type argument with NestedIndexWithoutDot`` () =
-    { config with
-        MaxLineLength = 30
-        MultilineBracketStyle = Aligned }
-    |> formatSourceString
+    formatSourceString
         """
 something<
         Foo<
@@ -453,6 +444,7 @@ something<
          >
      >["thing"][8](a,b)
 """
+        alignedMaxLine30
     |> prepend newline
     |> should
         equal
@@ -467,10 +459,7 @@ something<
 
 [<Test>]
 let ``multiline type argument with EndsWithDualListApp`` () =
-    { config with
-        MaxLineLength = 30
-        MultilineBracketStyle = Aligned }
-    |> formatSourceString
+    formatSourceString
         """
 div<
     Bar<
@@ -479,6 +468,7 @@ div<
      >
  > [ ClassName "container" ] [ str "meh" ]
 """
+        alignedMaxLine30
     |> prepend newline
     |> should
         equal
@@ -495,11 +485,7 @@ div<
 
 [<Test>]
 let ``multiline type argument with elmish EndsWithDualListApp`` () =
-    { config with
-        MaxLineLength = 30
-        MultilineBracketStyle = Aligned
-        ExperimentalElmish = true }
-    |> formatSourceString
+    formatSourceString
         """
 div<
     Bar<
@@ -508,6 +494,8 @@ div<
      >
  > [ ClassName "container" ] [ str "meh" ]
 """
+        { alignedMaxLine30 with
+            ExperimentalElmish = true }
     |> prepend newline
     |> should
         equal
@@ -525,10 +513,7 @@ div<
 
 [<Test>]
 let ``multiline type argument with EndsWithSingleListApp`` () =
-    { config with
-        MaxLineLength = 30
-        MultilineBracketStyle = Aligned }
-    |> formatSourceString
+    formatSourceString
         """
 input<
     Bar<
@@ -537,6 +522,7 @@ input<
      >
  > [ Type "text" ]
 """
+        alignedMaxLine30
     |> prepend newline
     |> should
         equal
--- a/src/Fantomas.Core.Tests/TypeAnnotationTests.fs
+++ b/src/Fantomas.Core.Tests/TypeAnnotationTests.fs
@@ -135,7 +135,7 @@ type CancellableTaskResultBuilderBase with
 """
 
 [<Test>]
-let `` Aligned bracket style in anonymous record is respected, #2706`` () =
+let ``Aligned bracket style in anonymous record is respected, #2706`` () =
     formatSourceString
         false
         """
@@ -163,7 +163,7 @@ let private asJson (arm: IArmResource) =
 """
 
 [<Test>]
-let `` Aligned bracket style in anonymous record is respected for multiple types, #2706`` () =
+let ``Aligned bracket style in anonymous record is respected for multiple types, #2706`` () =
     formatSourceString
         false
         """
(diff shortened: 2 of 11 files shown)


nojaf:
Thanks for following up on this. I need to play with this some more and we may want to have the conversation over at the style guide as well.
Once all is settled, we should target 6.1 for this.

src/Fantomas.Core/CodePrinter.fs
```diff
@@ -3037,7 +3045,8 @@ let genType (t: Type) =
         +> optSingle genIdentListNodeWithDot node.PostIdentifier
         +> genSingleTextNode node.LessThen
         +> addExtraSpace
-        +> col sepComma node.Arguments genType
+        +> leadingExpressionIsMultiline (colGenericTypeParameters node.Arguments) (fun isMultiline ->
```
nojaf, line 3048: We should add some tests where the `<` and `>` contain trivia.
Cases like:
```fsharp
type X =
    Teq< //
        int
     //
     >
```
won't be covered right now.

src/Fantomas.Core/CodePrinter.fs
```diff
@@ -3037,7 +3045,8 @@ let genType (t: Type) =
...
+            if isMultiline then !- " " else sepNone)
```
nojaf, line 3049: Use helper `onlyIf` here.

src/Fantomas.Core/CodePrinter.fs
```diff
@@ -2191,13 +2191,21 @@ let genKeepIdent (startNode: Node) (e: Expr) ctx =
         indentSepNlnUnindent (genExpr e) ctx
 
 let colGenericTypeParameters typeParameters =
-    coli sepComma typeParameters (fun idx t ->
-        let leadingSpace =
-            match t with
-            | Type.Var n when idx = 0 && n.Text.StartsWith("^") -> sepSpace
-            | _ -> sepNone
+    let short sep =
```
nojaf, line 2194: Maybe rename to `genParameters` instead of `short`.
We could have a separate `let short = genParameters sepComma`.

src/Fantomas.Core/CodePrinter.fs
nojaf: Also please add a code comment on why we are doing this in the first place.

src/Fantomas.Core.Tests/TypeAnnotationTests.fs
```diff
@@ -122,11 +127,79 @@ type CancellableTaskResultBuilderBase with
...
+    |> convertTo<
+        {|
+            kind: string
+            properties: {| statisticsEnabled: bool |}
+        |}
+       >
+"""
+
+[<Test>]
+let `` Aligned bracket style in anonymous record is respected for multiple types, #2706`` () =
```
nojaf, line 172: We should have the same test for the other styles.

nojaf (changes requested):
Hi @josh-degraw, many thanks again for taking the lead in this during https://amplifying-fsharp.github.io/sessions/2023/04/14/

I think the change is sensible but this does change the style and so I would like to see this covered by https://learn.microsoft.com/en-us/dotnet/fsharp/style-guide/formatting#formatting-explicit-generic-type-arguments-and-constraints

And also to have a link for people who will open an issue claiming the additional space is a bug. (Trust me, it will be reported 🙈).

Could you open an issue in https://github.com/fsharp/fslang-design/issues so we can further discuss these implications?

And please also retarget this branch to `v6.1`.

Thanks!

josh-degraw:
Sorry it's taken me so long to finally get back around to trying to finish this up haha.

I'm not sure if something has changed on the parser level since I originally worked on this, but a test that previously succeeded is now failing and I'm having trouble figuring out the best way to address it.

I extracted the relevant bits into its own test but essentially this kind of expression (which is how my PR originally formatted this in the updated tests here):
```fs
let bv =
    unbox<
        Foo<
            'innerContextLongLongLong,
            'bb -> 'b
            >
    >
        bf
```

Is no longer an idempotent operation. It seems like if I increase the indent of the final `>` when it's output in `genExpr` under `Expr.TypeApp` so it looks like this:
```fs
let bv =
    unbox<
        Foo<
            'innerContextLongLongLong,
            'bb -> 'b
            >
     >
        bf
```

it becomes idempotent again, but since this worked before I rebased this PR I'm not sure if this is a bug in the AST parsing layer or just in my code here.

nojaf:
@josh-degraw thanks for picking this up again! Much appreciated!

I'm afraid this one kinda sucks:

image

The `unbox<...> bf` no longer is a function application after formatting.

So, instead of `a b` it becomes `a ; b` conceptually.

I'm not sure what we need to do to keep this as a function application.

josh-degraw:
> @josh-degraw thanks for picking this up again! Much appreciated!
> 
> I'm afraid this one kinda sucks:
> 
> image
> 
> The `unbox<...> bf` no longer is a function application after formatting.
> 
> So, instead of `a b` it becomes `a ; b` conceptually.
> 
> I'm not sure what we need to do to keep this as a function application.

Yeah like I mentioned it seems like just increasing the indent for the closing angle bracket at least fixes the parser issue, but that's still valid code with the additional indent right? If not I guess we'd have to carve out an exemption for this scenario or wrap in parentheses or something

nojaf:
Ok, consider this example:

```fsharp
foo<
    'bar
        -> int
>
    barry
```

This looks like it is a `SynExpr.App(foo<...>, barry)` but actually it is already is two top level `SynModuleDecl.Expr`.

Adding a space before `>` makes it into a `SynModuleDecl.Expr (SynExpr.App ...)`.
So, I think our hand is forced here to just move it one space further.

nojaf:
Added a workaround of `let startColumn = ctx.Column + 1`

josh-degraw:
I tried that but that messes with a bunch of other tests, wouldn't we want to only do this in this specific situation?

nojaf:
Hmm, I think the safest thing to do would be to have it all the time.
We essentially do the same thing with `SynType` I think.

Digging into the exact case of `SynExpr.App(SynExpr.TypeApp, ...)` might be quite challenging and not really worth it.

josh-degraw:
Fair enough. Looks like most of the tests that fail with that change were added in this PR anyway haha. I'll update the tests to reflect this

nojaf:
@josh-degraw took a quick peek if it is possible. It seems less intrusive at first glance to remove the space.

This of course can still go wrong in things like `Expr.AppLongIdentAndSingleParenArg`, `Expr.AppSingleParenArg`, `Expr.AppWithLambda`, `Expr.NestedIndexWithoutDot`, `EndsWithDualListApp`, `EndsWithSingleListApp`, ...

We should check what happens in those scenarios. If we need to update all these individually I don't think it is worth it.

src/Fantomas.Core.Tests/DallasTests.fs
```diff
@@ -1596,11 +1596,14 @@ let autoCompleteItems: cmap<DeclName, DeclarationListItem * Position * string<Lo
         equal
         """
 let autoCompleteItems
-    : cmap<DeclName, DeclarationListItem *
-      Position *
-      string<LocalPath> *
-      (Position -> option<string>) *
-      FSharp.Compiler.Syntax.ParsedInput> =
+    : cmap<
+          DeclName,
```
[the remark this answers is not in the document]
josh-degraw: Do you still consider this a problem or are we okay with this indent level?

josh-degraw:
> This of course can still go wrong in things like Expr.AppLongIdentAndSingleParenArg, Expr.AppSingleParenArg, Expr.AppWithLambda, Expr.NestedIndexWithoutDot, EndsWithDualListApp, EndsWithSingleListApp, ...

Can you maybe help me know in what situations these nodes would apply here? I'm not quite sure exactly how to test for these just going off of the node names themselves (not quite there yet haha..)

nojaf:
> Can you maybe help me know in what situations these nodes would apply here?

Sure thing.

[SynExpr.App](https://fsprojects.github.io/fantomas/reference/fantomas-fcs-syntax-synexpr.html#App) has historically been a tricky one to format. Because it really depends on what expressions it has before we decide how we can format it.

Some [examples](https://fsprojects.github.io/fantomas-tools/#/oak?data=N4KABGBEDGD2AmBTSAuKB6dYCiAPADgE4CC%2B%2BAygJYB2A5gDaIAKAhoYtcYbQHIKIAdQgOoAxWLAAUQkRAiYZ1OWABGbRQEpFizDgIkyAdUoAXABYAZFgFsV8FnySK1hSQDMArktxgAtAD4wXC1hah0sPCJSfEdBUJYwAG0wAAl2NzABSABnWGtEX3oaAGssgF0kxWVsk0JMmFhqEw4TLMUy7VDdSMIASWo3Slxo2MzQn18wAE9OkW79chtEfsHhsmzYxR8AKmmwXYAvSAAacCgkQepEbNQoE7PISmzRbMpbtxZ6bMRTiEfsgDihBY%2BDMADVKIgAO7vT7fEAAXyAA):
```fsharp
// ExprAppSingleParenArgNode
Foo(
    //
    bar
)

// ExprAppWithLambdaNode
bar(fun x -> x)

// ExprAppNode
a [ Href "some-link"] [
    str "content"
]

// ExprInfixAppNode 
x - y

// ExprSameInfixAppsNode
x * y * z
```

image

All of these expressions are `SynExpr.App`, but we transform them into different `Oak` shapes in `ASTTransformer`.

This is sometimes very convenient but at the same has its drawbacks.

A clear pro: sometimes you really need to deal with these shapes very carefully and it helps to have separation.

A clear con: the is no single entry `genApp` that deals with all types of applications. It can be a bit scattered.

A lot of what we have today is somewhat historical as well. Infix applications are a good example of that. Almost all of the code pre-dates the style guide and it is just what we have today.

We split the applications into different Oak nodes in:
https://github.com/fsprojects/fantomas/blob/21bae08e87b63b2f9824ec19d4aa326d16309075/src/Fantomas.Core/ASTTransformer.fs#L1184-L1315

So, to answer your question, that would be the list you need to check.

On a side note, some of these nodes probably shouldn't exist because they are too focused on what we do with them inside `CodePrinter`. For this reason, you also have `EndsWithDualListApp` as an active pattern instead of a separate node. It is hard to really draw the line there.

Anyway, you should check what happens if any of these nodes can be used with `ExprTypeAppNode` and what happens if they do with a specific case.

I hope this helps a little.

src/Fantomas.Core.Tests/TypeAnnotationTests.fs
```diff
@@ -266,25 +262,27 @@ let private asJson (arm: IArmResource) =
        >
 """
 
+let forcedLongDefnConfig =
```
nojaf, line 265: I'm not a fan of this configuration.
The only thing you are trying to achieve is to spread things over multiple lines, using `MaxLineLength = 30`. All the other settings are irrelevant and misleading to readers.

src/Fantomas.Core.Tests/TypeAnnotationTests.fs
```diff
@@ -299,82 +297,211 @@ let bv =
...
-        """
-let private asJson (arm: IArmResource) =
-    arm.JsonModel
-    |> convertTo<{|
-        kind: string
-        properties: {| statisticsEnabled: bool |}
-    |}>
-"""     
-        config
+let ``Multiline type argument with AppLongIdentAndSingleParenArg`` () =
```
nojaf, line 301: Nit: lowercase unit test names.

src/Fantomas.Core.Tests/TypeAnnotationTests.fs
josh-degraw: Yeah, fully agree, this latest commit was mainly a WIP

src/Fantomas.Core.Tests/TypeAnnotationTests.fs
```diff
@@ -262,23 +262,11 @@ let private asJson (arm: IArmResource) =
...
-        MultilineBracketStyle = Aligned
-        AlignFunctionSignatureToIndentation = true
-        // AlternativeLongMemberDefinitions = true
-        MultiLineLambdaClosingNewline = true
-        NewlineBetweenTypeDefinitionAndMembers = false }
-
 [<Test>]
 let ``type application including nested multiline function type`` () =
-    forcedLongDefnConfig
+    { config with
```
nojaf, line 267: Nitpick but this doesn't comply with the [pull request ground rules](https://fsprojects.github.io/fantomas/docs/contributors/Pull%20request%20ground%20rules.html).

I don't really see a compelling reason to deviate from the standard unit template.
All our tests are the same, I'm not a fan of mixing that up in this PR.

nojaf:
Be aware that in its current form this is not working out for every `SynExpr.App` combination.
I did a quick test with:

```fsharp
[<Test>]
let ``dual list application`` () =
    formatSourceString
        """
unbox<Foo<'innerContextLongLongLong,'bb -> 'b>> [
    ClassName "meh"
] [
    str "foo"
]
"""
        { config with
            MaxLineLength = 12
            MultilineBracketStyle = Stroustrup }
    |> prepend newline
    |> should
        equal
        """
unbox<
    Foo<
        'innerContextLongLongLong,
        'bb
            -> 'b
     >
> [
    ClassName 
        "meh"
] [
    str 
        "foo"
]
"""
```

josh-degraw:
> Be aware that in its current form this is not working out for every `SynExpr.App` combination.

Yeah I know, I'm still having a little trouble pinpointing the spot to fix some of these

josh-degraw:
> Don't hesitate to reach out if you wanna pair on this sometime!

That might be helpful. I felt like a few I found right where I thought I needed to make changes to fix some of these issues but it didn't work haha

src/Fantomas.Core.Tests/DallasTests.fs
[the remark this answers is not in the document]
nojaf: Hmm, I think I can live with this.
This sample screams "use type alias" anyway.

nojaf (approved):
Alright, I think this is worth merging in and release as an alpha.
Thank you for your endurance here @josh-degraw!

PR #2859, opened 2023-04-21
Fix documented defaults

cmeeren:
Some values did not match those in https://github.com/fsprojects/fantomas/blob/main/src/Fantomas.Core/FormatConfig.fs.

For `MaxDotGetExpressionWidth` and `NewlineBetweenTypeDefinitionAndMembers`, I had to update the examples so they demonstrate an actual change.

You may consider moving the `NewlineBetweenTypeDefinitionAndMembers` from the G-Research style section, because (as I understand it) the G-Research style is now the default. (This is not done in this PR.)


dawedawe:
Thanks @cmeeren 
I pushed a change which should help avoiding such things in the future.

dawedawe (approved):

PR #2868, opened 2023-04-30
Expose TransformAST api.

nojaf:
I have a use-case in Telplin to transform a ParsedInput (created with a specific set of defines) to an Oak. This would need a `6.1` branch.

@dawedawe you on board with having this?


nojaf:
This would also be slightly more convenient for the Oak viewer.

dawedawe (approved):
Sure, if there's use for it, let's have it.

PR #2872, opened 2023-05-04
Automate releases

nojaf:
This would automate the release process.

### DIFF
--- a/build.fsx
+++ b/build.fsx
@@ -57,10 +57,6 @@ let pushPackage nupkg =
         return result.ExitCode
     }
 
-let getNuGetPackages () =
-    Directory.EnumerateFiles("bin", "*.nupkg", SearchOption.TopDirectoryOnly)
-    |> Seq.filter (fun nupkg -> not (nupkg.Contains("Fantomas.Client")))
-
 pipeline "Build" {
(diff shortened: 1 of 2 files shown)


build.fsx
```diff
@@ -313,3 +312,153 @@ pipeline "Init" {
...
+          Draft = draft }
+
+let getReleaseNotes currentRelease lastRelease =
+    let date = lastRelease.Date.ToString("yyyy-MM-dd")
+    let authorMsg =
+        let authors =
+            Cli
+                .Wrap("gh")
+                .WithArguments(
+                    $"pr list -S \"state:closed base:main closed:>{date} -author:app/robot\" --json author --jq \".[].author.login\""
```
dawedawe, line 377: Mmh, I think that could miss authors with PRs closed after the last release but still on the same day as the last release.
Not really critical but something to be aware of.

nojaf: Yeah, this is a bit of a trade-off. When using `closed:>={date}` authors would be included twice if we release on the same day.

dawedawe: Yes, it's okay for me. Just wanted to point out that we should be careful if we have such a day.

dawedawe (approved):

PR #2891, opened 2023-06-02
Add regression test for cons pattern in let binding (#1996)

clkbug:
I wrote a regression test case for cons pattern in let binding (#1996).
Is it OK?


nojaf (approved):
Hi there, thanks for this PR!
There was something a little off in the test and I tweaked it slightly.

PR #2895, opened 2023-06-02
Update editorconfig package

nojaf:
This uses a cache for parsing editor config files.
See https://github.com/editorconfig/editorconfig-core-net/pull/23.
I tested this for Fantomas itself (200 files) and could see a slight improvement (50ms faster give or take). This would have more effect when with larger editor configs and nested scenarios. 
Nonetheless worth having I think.


dawedawe (approved):
Very nice!

PR #2901, opened 2023-06-14
[WIP] Smarter ignore

nojaf:
Poking around with some ideas:
- Use a lot more `System.IO.Abstractions` so we can unit test the tool better.
- Extract all code that finds files and folders to process into a new module.
- Don't care about ignored files in `Check` and `Format` code. That should already have been done.


nojaf:
Meh

PR #2919, opened 2023-06-28
Update FCS to 'Correct range of SynInterfaceImpl'

nojaf:
Due to https://github.com/fsharp/fslang-suggestions/issues/1273, some warnings are now errors and I had to update the input of some of the unit tests.


dawedawe (approved):
I very much like the stricter rules.

PR #2922, opened 2023-07-10
replacing "vanity alignment" with "whitespace alignment"

smoothdeveloper:
This is just replacing the adjective which is unfortunate with something which is neutral, related to : https://github.com/dotnet/docs/pull/36143

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -380,7 +380,7 @@
 
 ### Fixed
 * Comments in SynArgPats.NamePatPairs are lost. [#2541](https://github.com/fsprojects/fantomas/issues/2541)
-* Vanity alignment used inside base ctor call. [#2111](https://github.com/fsprojects/fantomas/issues/2111)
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -380,7 +380,7 @@
 
 ### Fixed
 * Comments in SynArgPats.NamePatPairs are lost. [#2541](https://github.com/fsprojects/fantomas/issues/2541)
-* Vanity alignment used inside base ctor call. [#2111](https://github.com/fsprojects/fantomas/issues/2111)
+* Whitespace alignment used inside base ctor call. [#2111](https://github.com/fsprojects/fantomas/issues/2111)
```
dawedawe, line 383: I don't really like changing such a thing in the changelog file.
It feels too much like rewriting history to me.

smoothdeveloper: Mmh, I see, maybe we can leave the changelog if that's contentious and the other part of the change seems beneficial?

PR #2923, opened 2023-07-11
Brace yourself, warnings are coming

dawedawe:
As the good early adopters we claim to be, let's silence the warnings from lang preview.


nojaf (approved):
Thanks, good thinking!
Wanna go all-in and start using dotnet 8 previews with `<LangVersion>Preview</LangVersion>` as well?

dawedawe:
> Thanks, good thinking! Wanna go all-in and start using dotnet 8 previews with `<LangVersion>Preview</LangVersion>` as well?

Let's try. I'll do a separate PR for that.

PR #2961, opened 2023-10-06
Analyzers integration

dawedawe:


src/Fantomas.Analyzers/Fantomas.Analyzers.fsproj
```diff
@@ -0,0 +1,15 @@
+<Project Sdk="Microsoft.Build.NoTargets/1.0.80"> <!-- This is not a project we want to build. -->
```
nojaf, line 1: Not a fan of putting this in the source.
We should have this somewhere in the repo root.
Eventually, we will start writing Fantomas-specific analyzers which should be in this project.
Nit: the filename can just be `analyzers.proj`

build.fsx
```diff
@@ -75,6 +77,7 @@ pipeline "Build" {
         )
     }
     stage "CheckFormat" { run "dotnet fantomas src docs build.fsx --check" }
+    stage "GetAnalyzers" { run "dotnet build ./src/Fantomas.Analyzers/Fantomas.Analyzers.fsproj" }
```
nojaf: A `dotnet restore` should be enough here.

build.fsx
```diff
@@ -64,6 +64,8 @@ pipeline "Build" {
         run (
             cleanFolders
                 [| "bin"
+                   "src/Fantomas.Analyzers/bin/Release"
```
nojaf, line 67: There probably is no need for this really.

build.fsx
```diff
@@ -57,13 +57,17 @@ let pushPackage nupkg =
...
+    $"dotnet fsharp-analyzers --project {projectPath} --analyzers-path ./.analyzerpackages/g-research.fsharp.analyzers/ --verbose"
+
 pipeline "Build" {
     workingDir __SOURCE_DIRECTORY__
     stage "RestoreTools" { run "dotnet tool restore" }
     stage "Clean" {
         run (
             cleanFolders
-                [| "bin"
+                [| ".analyzerpackages/g-research.fsharp.analyzers"
```
dawedawe, line 69: Old versions of analyzer DLLs that are still on disk might be picked up and cause issues.
We could extend the analyzer path but that would mean we have to maintain the version in two places, so let's remove them.

nojaf: I'm a little less in favour of this. Can't we read the `proj` file and target the correct folder that way? I want this to work during a flight on the aeroplane 😏

build.fsx
```diff
@@ -66,8 +66,8 @@ let analyzersVersion =
     let matches = regex.Match(s)
     matches.Groups[1].Value
 
-let analyzeProject projectPath =
-    $"dotnet fsharp-analyzers --project {projectPath} --analyzers-path ./.analyzerpackages/g-research.fsharp.analyzers/{analyzersVersion} --verbose"
+let analyzeProjects (projectPaths: string seq) =
+    $"dotnet fsharp-analyzers --project {String.Join(' ', projectPaths)} --analyzers-path ./.analyzerpackages/g-research.fsharp.analyzers/{analyzersVersion} --verbose"
```
nojaf, line 70: Nitpick: the cool kids use `String.concat " " projectPaths`

dawedawe:
I think this is good to go in now.

nojaf (approved):
Great work!

PR #2964, opened 2023-10-18
use __SOURCE_DIRECTORY__ for the analyzersProjectPath

dawedawe:
With this PR, you can do a simple `dotnet fsi build.fsx` after a `git clean -xdf`.
Before, you had to do a `dotnet fsi build.fsx -p Init` prior to the `Build` pipeline.

### DIFF
--- a/build.fsx
+++ b/build.fsx
@@ -58,7 +58,7 @@ let pushPackage nupkg =
         return result.ExitCode
     }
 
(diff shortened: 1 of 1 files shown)


nojaf (approved):
Ah yes, I think I have seen that before as well. So that was the culprit.

PR #2967, opened 2023-10-19
Update configuration page with editor config setting copy button.

nojaf:
This fixes the configuration page (icons were broken due to CDN no longer working).
Add copy/past icons next to the settings.


dawedawe (approved):
Nice!

PR #2968, opened 2023-10-26
Integrate analyzers via msbuild

dawedawe:
The sarif file upload might not work yet. Let's try with a wildcard first and then go to a directory to collect all sarif files and upload that.


Directory.Build.props
```diff
@@ -58,4 +58,10 @@ Some common use cases include:
         <PackageReference Include="Ionide.KeepAChangelog.Tasks" Version="0.1.8" PrivateAssets="all" />
         <PackageReference Include="DotNet.ReproducibleBuilds" Version="1.1.1" PrivateAssets="all" />
     </ItemGroup>
+   <ItemGroup>
+        <PackageReference Include="G-Research.FSharp.Analyzers" Version="0.1.5">
```
nojaf, line 62: Bump to 0.16

Directory.Build.targets
```diff
@@ -0,0 +1,8 @@
+<Project>
+
+    <Target Name="AnalyzeProject">
+        <Message Importance="High" Text="$(MSBuildProjectFile)" />
```
nojaf, line 4: Maybe update this to `Start analyzing $(MSBuildProjectFile)`

Directory.Solution.targets
```diff
@@ -0,0 +1,16 @@
+<Project>
+
+    <ItemGroup>
+        <ProjectsToAnalyze Include="src\**\*.fsproj" Exclude="src\Fantomas.FCS\Fantomas.FCS.fsproj"/>
+    </ItemGroup>
+
+    <Target Name="AnalyzeSolution">
+        <Message Importance="High" Text="@(ProjectsToAnalyze)" />
```
nojaf, line 8: Remove message

Directory.Solution.props
```diff
@@ -0,0 +1,3 @@
+<Project>
+    <Import Project="Directory.Build.props" />
```
nojaf, line 2: We don't actually need this file, please remove.

build.fsx
```diff
@@ -94,23 +76,8 @@ pipeline "Build" {
...
-                [ "./src/Fantomas/Fantomas.fsproj"
-                  "./src/Fantomas.Benchmarks/Fantomas.Benchmarks.fsproj"
-                  "./src/Fantomas.Client/Fantomas.Client.fsproj"
-                  "./src/Fantomas.Client.Tests/Fantomas.Client.Tests.fsproj"
-                  "./src/Fantomas.Core/Fantomas.Core.fsproj"
-                  "./src/Fantomas.Core.Tests/Fantomas.Core.Tests.fsproj"
-                  "./src/Fantomas.Tests/Fantomas.Tests.fsproj" ]
-        )
-    }
+    stage "Analyze" { run "dotnet msbuild /t:AnalyzeSolution" }
```
nojaf: The exit code of this right now is still 0 even if there were problems.

nojaf (commented):
A good start!

build.fsx
```diff
@@ -82,6 +82,9 @@ pipeline "Build" {
     stage "Build" { run "dotnet build -c Release" }
     stage "Analyze" {
         run (fun _ -> async { System.IO.Directory.CreateDirectory(analysisReportsDir) |> ignore })
+        envVars
```
nojaf, line 85: I would put this closer to the tool invocation.
Can we do this right before the `<Exec Command`?

nojaf: Probably using https://learn.microsoft.com/en-us/visualstudio/msbuild/setenv-task?view=vs-2022

Directory.Build.targets
```diff
@@ -0,0 +1,18 @@
...
+
+    <Target
+        Name="AnalyzeProject" 
+        DependsOnTargets="Restore;ResolveAssemblyReferencesDesignTime;ResolveProjectReferencesDesignTime;ResolvePackageDependenciesDesignTime;FindReferenceAssembliesForReferences;_GenerateCompileDependencyCache;_ComputeNonExistentFileProperty;BeforeBuild;BeforeCompile;CoreCompile">
+        
+        <Message Importance="normal" Text="fsc arguments: @(FscCommandLineArgs)" />
+        <Message Importance="High" Text="Analyzing $(MSBuildProjectFile)"/>
+        <Exec
+            ContinueOnError="true"
+            Command="dotnet fsharp-analyzers --fsc-args &quot;@(FscCommandLineArgs)&quot; --analyzers-path &quot;$(PkgG-Research_FSharp_Analyzers)\analyzers\dotnet\fs&quot; --exclude-analyzer PartialAppAnalyzer --fail-on-warnings GRA-STRING-001 --verbose --report &quot;../../$(MSBuildProjectName)-analysis.sarif&quot;">
```
nojaf, line 11: Do `--fail-on-warnings` for all codes please.

nojaf (approved):
One nit left but man this is good stuff!

nojaf:
`--report &quot;..\..\analysisreports\$(MSBuildProjectName)-analysis.sarif&quot;`

nojaf:
@dawedawe The reason I want to set `DOTNET_ROLL_FORWARD_TO_PRERELEASE` and `DOTNET_ROLL_FORWARD` in the target is that doing this allows for `dotnet msbuild /t:AnalyzeSolution` to just work from the command line.

Issue #2973, opened 2023-11-01
ExperimentalKeepIndentInBranch should fire when first branch was all on one line


Smaug123:
[...]

PR #2975, opened 2023-11-02
Also accept alignment from if keyword in genKeepIdentIfThenElse.

nojaf:
See #2973.

@Smaug123 your desired result doesn't have a blank line after `else`:

```
module Program =
    let main _ =
        if false then
            1
        else
        printfn "hi!"
        0
``` 

We currently always add a blank line there. Do you want to change that behaviour as well?

### DIFF
--- a/src/Fantomas.Core.Tests/KeepIndentInBranchTests.fs
+++ b/src/Fantomas.Core.Tests/KeepIndentInBranchTests.fs
@@ -2183,3 +2183,55 @@ module Foo =
 
         failwith ""
 """
+
+[<Test>]
+let ``if keyword and elseBranch align, 2973`` () =
+    formatSourceString
+        false
+        """
+module Program =
+    let main _ =
+        if false then 1 else
+        printfn "hi!"
+        0
+"""
+        config
+    |> prepend newline
+    |> should
+        equal
+        """
+module Program =
+    let main _ =
+        if false then
+            1
+        else
+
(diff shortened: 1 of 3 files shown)


Smaug123:
I'm very happy with adding the blank line. I was toying with the question; without the blank line it does look a bit cramped.

src/Fantomas.Core.Tests/KeepIndentInBranchTests.fs
```diff
@@ -2183,3 +2183,55 @@ module Foo =
...
+        if false then
+            1
+        else
+
+        printfn "hi!"
+        0
+"""
+
+[<Test>]
+let ``elif keyword and elseBranch align`` () =
```
Smaug123, line 2214: For completeness, would you mind adding a test for the following too?

```fsharp
async {
    if not (proc.Start ()) then return Error "failed to start" else
    use stdout = proc.StandardOutput
    let! ct = Async.CancellationToken
    return! Async.AwaitTask (stdout.ReadToEnd ct)
}
```

nojaf: Sure thing!

Smaug123 (approved):
I can't speak for the implementation, but the tests look like they're doing the right thing to me! Thanks.

dawedawe (approved):

Issue #2972, opened 2023-11-01
Reformatting a list expression with if/else changes the list expression


gubser:
[...]

PR #2980, opened 2023-11-11
Dev container

nojaf:
As a response to @gubser poor experience in https://github.com/fsprojects/fantomas/issues/2972, I tried to set up a devcontainer.

While doing this, I did encounter some problems when only `8rc2` SDK was installed.
I needed `RollForward` for the test projects and of course, had to workaround https://github.com/dotnet/sdk/issues/35989.

During my experiment, I updated `fsdocs` to the latest alpha. Even though we use a custom theme, I had to change some things for the search.
<strike>
Ionide was not starting in my dev container. @Martin521 would you have any ideas what I'm still missing?
</strike>

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,38 @@
+// For format details, see https://aka.ms/devcontainer.json. For config options, see the
+// README at: https://github.com/devcontainers/templates/tree/main/src/dotnet-fsharp
+{
+	"name": "Fantomas",
+	// Or use a Dockerfile or Docker Compose file. More info: https://containers.dev/guide/dockerfile
+	"image": "mcr.microsoft.com/devcontainers/dotnet:1-6.0-bullseye",
+
+	// Features to add to the dev container. More info: https://containers.dev/features.
+	// "features": {},
+
+	// Configure tool-specific properties.
+	"customizations": {
+		// Configure properties specific to VS Code.
+		"vscode": {		
+			// Add the IDs of extensions you want installed when the container is created.
+			"extensions": [
+				"Ionide.Ionide-fsharp",
+				"ms-dotnettools.csharp",
+				"tintoy.msbuild-project-tools"
+			]
+		}
+	},
+	"features": {
+		"ghcr.io/devcontainers/features/github-cli:1": {},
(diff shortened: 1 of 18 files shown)


nojaf:
image

nojaf:
Turns out, that I was trying to use a local FSAC inside the container.
Works now.

dawedawe:
Mmh, so far my success rate with the container is 2 out of 10 but if it can unblock others, sure, let's have it.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,38 @@
...
+	// "features": {},
+
+	// Configure tool-specific properties.
+	"customizations": {
+		// Configure properties specific to VS Code.
+		"vscode": {		
+			// Add the IDs of extensions you want installed when the container is created.
+			"extensions": [
+				"Ionide.Ionide-fsharp",
+				"ms-dotnettools.csharp",
```
dawedawe, line 18: Not sure if it helps, but maybe have the csharp extension before Ionide, as Ionide depends on it.

Martin521:
This is how dotnet/fsharp does it:

```
    "customizations": {
        "vscode": {
            // Add the IDs of extensions you want installed when the container is created.
            "extensions": [
                "ms-dotnettools.csharp",
                "Ionide.Ionide-fsharp",
                "tintoy.msbuild-project-tools"
            ]
        }
    },
```

.devcontainer/devcontainer.json
nojaf: Yeah, the order doesn't seem to matter.

nojaf:
Yup, stole that part from `dotnet/fsharp` indeed.

Martin521:
And wouldn't it make sense to use a .net 8 image?

nojaf:
> And wouldn't it make sense to use a .net 8 image?

Yes, probably, I selected the `F#` option from the UI.
Might make sense to use a Dockerfile instead.

nojaf:
You mean in Windows or in the container?

dawedawe:
In the container

nojaf:
Ok, I now have it in Windows. This is a checkout-as line ending shenannigan.
Forcing `lf` in the test now.

dawedawe (approved):
Much more stable for me now 👍

PR #2981, opened 2023-11-13
Remove sass from homepage

nojaf:
I refactor the homepage to no longer use Bootstrap and Sass.
This should visually be more of less the same.
The next step would be to refactor our custom `_template.html` and tweak the modern theme instead of rolling with our own.

Good stuff.


dawedawe (approved):

nojaf:
Yeah, I'm aware of that one. All good

PR #2985, opened 2023-11-15
Use customized modern theme for docs!

nojaf:
@dawedawe please play with this on your local machine.
If nothing out of the ordinary pops up we can just ship it!


dawedawe (approved):
Love it ❤️

Arrow7000:
Looks like there was some sloppy find & replace action here, because some text in [Configuration.fsx](https://github.com/fsprojects/fantomas/blob/main/docs/docs/end-users/Configuration.fsx) has been broken. 

If you search for `required`, `measured`, and `considered` those have been broken. `required` has been replaced with `requicolor="red"` and similar for the others.

https://github.com/fsprojects/fantomas/commit/51992ee149b722cedb5ab53dee819e116425cfee#diff-8c398a95c800a0f1ec078d842b43babe102dd037851a4c87255381d559b6baa0

PR #2987, opened 2023-11-15
UseArtifactsOutput

nojaf:
I'm enabling `<UseArtifactsOutput>`.


nojaf:
😔 I don't think `proj-info` knows how to deal with `UseArtifactsOutput`.
Will check if that is the case.

nojaf:
Currently bumping into https://github.com/dotnet/BenchmarkDotNet/issues/2466

dawedawe (approved):

PR #2992, opened 2023-11-18
Refactor test helpers

josh-degraw:
A first attempt / POC to simplify test code by reducing duplication and improving readability.

There are now two helper functions:

```fsharp
let formatSignatureString = formatFSharpString true
let formatSourceString = formatFSharpString false
```

Since most code used `formatSourceString false`, I kept that name the same and just changed it to pass `false` to the implementation.


josh-degraw:
Two remaining suggestions I'd make for test refactoring would be:

Now we could pipe `config` into `formatSourceString ""`, which I personally feel like would make things a lot more readable by eliminating the currently necessary hanging indent when you need to pass `config` after the source string. Consider:

<details>
<summary>Long example collapsed</summary>

```fsharp
[<Test>]
let ``multiline downcast expression, `` () =
    formatSourceString
        """
longMethodName
    longArgument
    longArgument2
:?> List<bool>
"""
        { config with
            MaxLineLength = 30
            SpaceBeforeUppercaseInvocation = true
            SpaceBeforeColon = true
            SpaceBeforeSemicolon = true
            AlternativeLongMemberDefinitions = true }
    |> prepend newline
    |> should
        equal
        """
longMethodName
    longArgument
    longArgument2
:?> List<bool>
```
would change to:

```fsharp
[<Test>]
let ``multiline downcast expression, `` () =
    { config with
        MaxLineLength = 30
        SpaceBeforeUppercaseInvocation = true
        SpaceBeforeColon = true
        SpaceBeforeSemicolon = true
        AlternativeLongMemberDefinitions = true }
    |> formatSourceString
        """
longMethodName
    longArgument
    longArgument2
:?> List<bool>
"""
    |> prepend newline
    |> should
        equal
        """
longMethodName
    longArgument
    longArgument2
:?> List<bool>
```

</details>

The second suggestion I'd make is to modify how `prepend string` is currently necessary almost everywhere. For example, even for one-line code tests such as this, we often need to add a newline, either explicitly or via `prepend newline`:
```fsharp
[<Test>]
let ``should keep parens around active patterns`` () =
    formatSourceString
        """let (|Boolean|_|) = Boolean.parse
    """
        config
    |> should
        equal
        """let (|Boolean|_|) = Boolean.parse
"""
```

We could add a helper like this to eliminate the need to call `prepend newline` or add an explicit newline in most places:
```fsharp
let shouldEqualTrimmed (y: string) (s: string) =
    let x = s.Replace("\r\n", "\n").Trim()
    y.Trim()
    |> should equal x
```

src/Fantomas.Core.Tests/ListTests.fs
```diff
@@ -430,8 +409,7 @@ let ``line comment inside array`` () =
 #if RELEASE
 [<Test>]
 let ``long array sequence`` () =
-    formatSourceString
-        false
+    formatImplSourceString
```
dawedawe, line 412:
```suggestion
    formatSourceString
```

dawedawe (approved):
Thanks Josh, I like it a lot.
I think we should keep the scope of this PR as it currently is and discuss/review the other two ideas separately.

src/Fantomas.Core.Tests/TestHelpers.fsi
```diff
@@ -9,7 +9,10 @@ module String =
 
 val config: FormatConfig
 val newline: string
-val formatSourceString: isFsiFile: bool -> s: string -> config: FormatConfig -> string
+
+val formatSignatureString: (string -> FormatConfig -> string)
+val formatSourceString: (string -> FormatConfig -> string)
```
nojaf, line 14: Nit: maybe a `///` to mention that the code will be an implementation file, and users should use `formatSignatureString` if they want to format a signature.

nojaf (approved):
Well, I don't think I'll emotionally recover from this 😅.

<details><summary>Details</summary>
<p>
I don't like change 
</p>
</details> 

But, I'm ok with these first changes.

PR #2994, opened 2023-11-20
Remove ExprAppSingleParenArgNode from Oak model.

nojaf:
This is an example of something I mentioned in [comment](https://github.com/fsprojects/fantomas/pull/2852#issuecomment-1818870225).

I'm not quite sure if this improves things and we should merge this.

But my reasoning is that `SyntakOak` should "strictly" not have any knowledge of how `CodePrinter` deals with the details of the captured nodes.

Right now: `f(g)` is represented as `ExprAppSingleParenArgNode` while it could have easily been an `ExprAppNode`. We make the distinction based on what we are about to do with it in `CodePrinter`.

I'm not sure that this is a good reason to do it. It covers the representation of the typed code just as well. The significance of arguments being a single `paren expr` should be kept inside of CodePrinter, where it actually matters.

This PR also isn't a plea to move everything back into `ExprAppNode`. I still think it makes a lot of sense to have `ExprInfixApp` and others. As a rule of thumb, I think we should try and keep **style** out of `SyntaxOak`. So groupings like `Expr.AppLongIdentAndSingleParenArg` and `Expr.AppWithLambda` should be ported to `CodePrinter` instead.

//cc @dawedawe @josh-degraw

### DIFF
--- a/src/Fantomas.Core/CodePrinter.fs
+++ b/src/Fantomas.Core/CodePrinter.fs
@@ -752,11 +752,6 @@ let genExpr (e: Expr) =
         | Expr.Constant _
         | Expr.InterpolatedStringExpr _ when not (String.startsWithOrdinal "%" node.Operator.Text) ->
             genSingleTextNode node.Operator +> sepSpace +> genExpr node.Expr
-        | Expr.AppSingleParenArg appNode ->
-            genSingleTextNode node.Operator
-            +> sepSpace
-            +> genExpr appNode.FunctionExpr
-            +> genExpr appNode.ArgExpr
         | Expr.AppLongIdentAndSingleParenArg appNode ->
             let mOptVarNode = appNode.FunctionName.Range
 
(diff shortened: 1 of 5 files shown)


src/Fantomas.Core/CodePrinter.fs
```diff
@@ -2309,48 +2314,52 @@ let genFunctionNameWithMultilineLids (trailing: Context -> Context) (longIdent:
...
+    /// Example: a {| X = y |}
+    | EndsWithSingleRecord of otherArguments: Expr list * record: Expr
+    /// Example: Foo(a,b)
+    | SingleParenArgument of appNode: ExprAppNode * parenthesesArgument: Expr
+    /// Example: path.Replace("../../../", "....")
+    | LongIdentAndSingleParenArgument of longIdent: ExprOptVarNode * parenthesesArgument: Expr
+    /// Example: myFunc a b c
+    | Regular of appNode: ExprAppNode
+
+// TODO: memoize function based on the range of appNode
```
josh-degraw, line 2332: Is memoizing something you'd like specific to this task, or is that a general improvement you'd like to make at some point?

nojaf: Specifically in this case.
During the pipeline of `CodePrinter` on multiple occasions, we need to transform `ExprAppNode` to `AppNode`. This transformation is always going to be the same (in a single run), so it should not be recalculated each time we invoke the function.

nojaf:
Meh

PR #2997, opened 2023-11-23
Update F# analyzers

nojaf:
Good stuff

### DIFF
--- a/src/Fantomas/Program.fs
+++ b/src/Fantomas/Program.fs
@@ -350,7 +350,7 @@ let main argv =
     let check = results.Contains <@ Arguments.Check @>
     let isDaemon = results.Contains <@ Arguments.Daemon @>
 
-    let partitionResults (results: #seq<FormatResult>) =
+    let partitionResults (results: #(FormatResult seq)) =
         (([], [], [], []), results)
         ||> Seq.fold (fun (oks, ignores, unchanged, errors) next ->
             match next with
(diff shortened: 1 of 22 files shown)


dawedawe (approved):

src/Fantomas/Program.fs
```diff
@@ -350,7 +350,7 @@ let main argv =
     let check = results.Contains <@ Arguments.Check @>
     let isDaemon = results.Contains <@ Arguments.Daemon @>
 
-    let partitionResults (results: #seq<FormatResult>) =
+    let partitionResults (results: #(FormatResult seq)) =
```
josh-degraw, line 353: I know this PR was already merged and it's an objectively minor thing but are we sure we want to use this naming style? Feels very clunky and unintuitive at a glance. Which analyzer is this from? I'd like to have a word with them 👊🏻

nojaf: https://learn.microsoft.com/en-us/dotnet/fsharp/style-guide/formatting#formatting-types-and-type-annotations

josh-degraw: Yeah I like that in general, I'm just talking about when using the flexible type syntax. `#seq<int>` is just so much more clear than `#(int seq)`. Ideally I think `int #seq` would be the best of both worlds but I don't think it's possible rn. 

Also if using multiple levels (e.g. `int list array`) it can get confusing. Neither of these situations are in the style guide, from what I have seen. Maybe I'll take the fight over there haha

PR #3024, opened 2023-12-21
Fix 2854, accidental treatment of old flags as folders

### LINKED ISSUE #2854: "fantomas --recurse" creates a "--recurse" directory
Just updated from `fantomas-tool 4.7.9` to `fantomas 6.0.1` and it creates `--recurse` directories in my modules.

Steps to reproduce:
```
witoldsz@678db1ca1a61:/app/modules/event-logger2$ ls
bin  event-logger2.sln	main  Makefile	test  VERSION  w3module.yaml
witoldsz@678db1ca1a61:/app/modules/event-logger2$ dotnet fantomas --recurse .
                                                             
  Formatted │ 0 │ Ignored │ 6 │ Unchanged │ 8 │ Errored │ 0  
                                                             
witoldsz@678db1ca1a61:/app/modules/event-logger2$ ls
bin  event-logger2.sln	main  Makefile	--recurse  test  VERSION  w3module.yaml
witoldsz@678db1ca1a61:/app/modules/event-logger2$ 
```

As you can see, the `--recurse` directory appeared.

I can see, the `fantomas 6.x` does not require `--recurse` anymore, but my project upgrading to fantomas 6 is in progress and the `--recurse` flag is still required by other modules (using older fantomas version) and there is one build script for them all.

dawedawe:
fixes #2854 

It's a bit complicated, but 
first we treated unknown flags as input folders
these input folders weren't found and then we treated these folders as output folders
these output folders didn't exist, so we created them, hence the famous `--recurse` folder came into life.

### DIFF
--- a/src/Fantomas/Program.fs
+++ b/src/Fantomas/Program.fs
@@ -221,11 +221,7 @@ let main argv =
         | None -> OutputPath.NotKnown
 
     let inputPath =
-        let maybeInput =
-            results.TryGetResult <@ Arguments.Input @>
-            |> Option.map (fun input ->
-                input
-                |> List.filter (fun s -> not (s.StartsWith("--", StringComparison.Ordinal)))) // don't accidentally treat old flags as folders
+        let maybeInput = results.TryGetResult <@ Arguments.Input @>
 
         match maybeInput with
         | Some [ input ] ->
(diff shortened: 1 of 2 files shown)


src/Fantomas/Program.fs
```diff
@@ -221,7 +221,11 @@ let main argv =
         | None -> OutputPath.NotKnown
 
     let inputPath =
-        let maybeInput = results.TryGetResult <@ Arguments.Input @>
+        let maybeInput =
+            results.TryGetResult <@ Arguments.Input @>
+            |> Option.map (fun input ->
+                input
+                |> List.filter (fun s -> not (s.StartsWith("--", StringComparison.Ordinal)))) // don't accidentally treat old flags as folders
```
nojaf, line 228: Is there not any Argu configuration we are missing here? Don't parse invalid input or something?

MangelMaxime:
Hello,

For me, the problem reported in #2854 is not limited to flags.

For example, if I run `dotnet fantomas dir1 dir2` where `dir1` and `dir2` doesn't exist.

Then fantomas creates these folders.

> these input folders weren't found and then we treated these folders as output folders

This behaviour feels strange to me, this means if the user makes a typo then the command has a different behaviour than what he would expect.

If this is for legacy reason, that's unfortunate. 

Otherwise, I think it would be nice to treat input folders as input only.

For example, if I run `cd dir1` when `dir1` doesn't exist, `cd` doesn't try to create that folder. It report that the folder doesn't exist so I can fix my command invocation.

dawedawe:
I pushed a change that should handle old flags and non-existing folders as @MangelMaxime described.

nojaf (approved):
Thanks!

dawedawe:
@MangelMaxime Please test our latest [alpha](https://www.nuget.org/packages/fantomas/6.3.0-alpha-005)
It should handle your issues more gracefully.

MangelMaxime:
I confirm that the alpha doesn't create the folder if it doesn't exist:

```bash
➜  Glutinum.Converter git:(main) ✗ dotnet fantomas di1 dir2 
Input path 'di1' not found.
➜  Glutinum.Converter git:(main) ✗ dotnet fantomas src dir2
Input path 'dir2' not found.
# src folder exist
```

Thank you

PR #3025, opened 2023-12-21
Fix output filename

dawedawe:
Uh, there really seems to be no one using this :(
I'll add tests later.

image

### DIFF
--- a/src/Fantomas.Tests/Integration/WriteTests.fs
+++ b/src/Fantomas.Tests/Integration/WriteTests.fs
@@ -1,5 +1,6 @@
 module Fantomas.Tests.Integration.WriteTests
 
+open System.IO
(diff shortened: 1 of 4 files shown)


src/Fantomas.Tests/Integration/WriteTests.fs
```diff
@@ -42,3 +42,34 @@ let ``incorrectly formatted file should be written`` () =
     exitCode |> should equal 0
 
     output |> should contain "has been written"
+
+[<Test>]
+let ``file should be written to out folder when input folder has trailing slash`` () =
```
dawedawe, line 47: This one fails in main because `A.fs` becomes `.fs` in the output folder

nojaf (approved):
Thanks!

PR #3027, opened 2023-12-26
[Ignore] Top level types

nojaf:
Part of the experiment in https://github.com/dotnet/fsharp/discussions/16450.
I have no intention to actually merge this.

The `tl;dr` of this is that I've removed all signature files and added type annotations to everything top-level so the compiler can extract signature files from the implementation files.


nojaf:
No longer pursuing this.

PR #3034, opened 2024-01-10
Update FCS to 'Parser: parse primary ctor params as normal patterns'

nojaf:
Update FCS to https://github.com/dotnet/fsharp/pull/16425


dawedawe (approved):
Wow, awesome!

PR #3039, opened 2024-01-30
POC double indentation.

nojaf:
Quick implementation of having double indentation for long function signature parameters.

//cc @auduchinok @dawedawe @josh-degraw @Smaug123 

Note that the equals sign rule didn't change. The [guide](https://learn.microsoft.com/en-us/dotnet/fsharp/style-guide/formatting) is pretty clear on that. It goes on its own line. 
Unless the last parameter is a long tuple.

For the sake of source control, I wouldn't change that:

image

image

### DIFF
--- a/src/Fantomas.Core.Tests/ExperimentalDoubleIdentParametersTests.fs
+++ b/src/Fantomas.Core.Tests/ExperimentalDoubleIdentParametersTests.fs
@@ -1,0 +1,30 @@
+﻿module Fantomas.Core.Tests.ExperimentalDoubleIdentParametersTests
+
+open NUnit.Framework
+open FsUnit
+open Fantomas.Core.Tests.TestHelpers
+
+let config =
+    { config with
+        ExperimentalDoubleIndentParameters = true
+        MaxLineLength = 10 }
+
+[<Test>]
+let ``initial sample`` () =
+    formatSourceString
+        """
+let sillyfuncWithParams parameterName1 ignoredParameterName2 ignoredParameterName3 =
+    longBody
+"""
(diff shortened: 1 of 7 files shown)


Smaug123:
@auduchinok I'd be interested to hear more about what this change lets you do.

src/Fantomas.Core.Tests/ExperimentalDoubleIdentParametersTests.fs
```diff
@@ -0,0 +1,64 @@
...
+        // Long body goes here...
+        longBody
+"""
+        config
+    |> prepend newline
+    |> should
+        equal
+        """
+type X() =
+    member x.Y
```
nojaf, line 48: @auduchinok this does the double indent but has the multiline tuple parameter.
According to the existing lore, we indent each entry of the tuple.
Does this meet your expectations?

nojaf:
I talked a bit more with @auduchinok on this.

It could be interesting to follow the approach of putting parameters on the current line until the max line length is crossed:

Screenshot 2024-03-13 at 15 00 15
Screenshot 2024-03-13 at 14 54 45
Screenshot 2024-03-13 at 14 56 06
Screenshot 2024-03-13 at 15 00 08
Screenshot 2024-03-13 at 14 54 41
Screenshot 2024-03-13 at 14 18 05
Screenshot 2024-03-13 at 14 18 01

It keeps things compact and having the double indent helps to differentiate the parameters with the body.

nojaf:
@auduchinok I think I'm going to close this as the subject lost steam.

PR #3046, opened 2024-02-12
Update analyzers to 0.24.0 SDK

nojaf:
I'm not sure if we really need those `treat-as-warnings` anymore.


dawedawe:
> I'm not sure if we really need those `treat-as-warnings` anymore.

I'm leaning more on the side to keep them, yes.

nojaf:
Don't they all produce warnings?

dawedawe:
No, some of the Ionide ones are producing Hints and Infos. But yeah, not a hill to die on.

dawedawe (approved):

Issue #3079, opened 2024-04-15
Regression: An empty line or comment at the end of a list breaks Stroustrup formatting


Linschlager:
[...]

PR #3080, opened 2024-04-15
Check Hash Directive more strictly inside of lists or arrays

Linschlager:
Resolves https://github.com/fsprojects/fantomas/issues/3079

Instead of giving up on Stroustrup-Formatting whenever there is Trivia on the closing brackets, only do that for HashDirectives and allow any other kind of Trivia such as comments or NewLines

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## 6.3.4 - 2024-04-16
+
+### Fixed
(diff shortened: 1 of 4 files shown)


nojaf (approved):
Thanks, this looks good.
Do you want a release for this?

Linschlager:
> Thanks, this looks good. Do you want a release for this?

That would be awesome as it's currently blocking us from updating

PR #3085, opened 2024-05-17
[WIP] Add basic support for "Consistent" Pattern Matching

isaacabraham:
Just an initial WIP PR.

### DIFF
--- a/src/Fantomas.Core/FormatConfig.fs
+++ b/src/Fantomas.Core/FormatConfig.fs
@@ -26,6 +26,19 @@ type MultilineFormatterType =
         | "number_of_items" -> Some MultilineFormatterType.NumberOfItems
         | _ -> None
 
+type PatternMatchStyle =
+    | LineSpecific
+    | Consistent
+    static member ToConfigString(cfg: PatternMatchStyle) =
+        match cfg with
+        | LineSpecific -> "line_specific"
+        | Consistent -> "consistent"
+    static member OfConfigString(cfgString: string) =
+        match cfgString with
+        | "line_specific" -> Some LineSpecific
+        | "consistent" -> Some Consistent
+        | _ -> None
+
 type MultilineBracketStyle =
     | Cramped
     | Aligned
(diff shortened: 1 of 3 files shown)


src/Fantomas.Core/FormatConfig.fs
```diff
@@ -227,7 +240,13 @@ type FormatConfig =
 
       [<Category("Convention")>]
       [<DisplayName("Applies the Stroustrup style to the final (two) array or list argument(s) in a function application")>]
-      ExperimentalElmish: bool }
+      ExperimentalElmish: bool
+
+      [<Category("Indentation")>]
+      [<DisplayName("How to format pattern match expression")>]
+      [<Description("Possible options include each line judged separately (default), or consistent (either all single line or all multiline)")>]
+      ExperimentalPatternMatchStyle : PatternMatchStyle
```
nojaf, line 248: I would prefer a boolean setting for this. This would be much more consistent with most other Fantomas settings.

Some reasons:
- Typo's are less likely in editorconfig. 
- Online tool doesn't need to be update.
- There are only two values here, most other DU's have three.

isaacabraham: Yes, I had thought about that. The reason I put it as an enum was because the original feature request had 4 cases (these two, plus "always single" and "always multiline"). Happy to move to a bool, although it would be a little less readable i.e. instead of mode, it would be e.g. `ConsistentPatternMatching = true / false` with the false case being "implicitly" the standard / existing behaviour.

isaacabraham: Nonetheless, it would be helpful to know why the current unit test is broken - any ideas?

nojaf: I appreciate your enthusiasm, but at the moment, we don't have enough support to justify having four different cases for this. Right now, we're considering just a boolean setting as an alternative approach. The other options aren't really worth pursuing at this stage.

> it would be helpful to know why the current unit test is broken - any ideas?

To give you a better answer, I'd need to see the test running on CI. My initial guess is that there might be an issue with your DU.

Thanks for understanding!

isaacabraham: Not really enthusiasm - just was thinking of the right thing to do. I've no problem to convert to a boolean and will do.

I've fixed the formatting so that should get the CI working anyway so that we can see the test fail on CI (hopefully) - that might enable you to explain why the test is failing, which would be good to know in order to improve my understanding of the Fantomas codebase, enabling me to become a more active participant in the future!

nojaf: I noticed you did not update https://github.com/fsprojects/fantomas/blob/873d9d70d755f2c058f49b4712f3d22abb28dce4/src/Fantomas/Daemon.fs#L125-L173 with your new config thing.

The build errors I noticed on the CI, are likely a result of `8.0.300` being installed on the agent.

isaacabraham: Great, thanks for this - that's the bit I couldn't find.

nojaf:
Hi @isaacabraham , just checking in to see if you still wish to pursue this?
If not, I'd like to close this.

PR #3087, opened 2024-05-24
rolForward to latestPatch to let build succeed on systems with 8.0.30…

dawedawe:
I might be missing something, but on my system LangVersion Preview and 8.0.300 have the effect that the argu attributes are reported as errors. So while argu has no compatible version out yet, let's stay with latestPatch


dawedawe:
Yes, but 6.2.4 is not released yet. Can you push the right button?

nojaf:
My bad, I pushed the button. https://www.nuget.org/packages/Argu/6.2.4

nojaf (approved):

PR #3100, opened 2024-07-09
Bump SDK to 8.0.300

nojaf:


Numpsy:
Are the ```Failed to inline the value 'getLambdaBodyExpr' marked 'inline'``` warnings in the build a result of https://github.com/dotnet/fsharp/issues/17161 or something else?

nojaf:
No clue, feel free to investigate.

baronfel:
It's a bug in the 300 release for F# - wait for this month's patch release and then go right to that one, by setting it as the min bar in your global.json.

nojaf:
As it is only a warning, I can temporary live with this.
Feel free to disagree @dawedawe, @josh-degraw.

dawedawe (approved):
Yeah, I won't loose sleep about that temporary warning.

PR #3118, opened 2024-09-13
Deal with F# 9

nojaf:
So, F# 9 is coming in November!

There are some new syntax features that we don't support yet:
- https://github.com/dotnet/fsharp/pull/15181
- https://github.com/dotnet/fsharp/pull/16861

In this PR, I'm updating the compiler (syncing the source dependency) and making the bare minimum adjustments to support the new syntax. This means I'm not focusing on style and will be winging it based on best effort.

To-do:
- [x] Pass all existing tests
- [x] Nullness tests
- [x] Access modifiers for auto-properties' getters and setters tests
- [x] Sanity check: format dotnet/fsharp to ensure everything is okay

Open questions:
- I would like to bump `FSharp.Core` and `StreamJsonRpc`, but I'm unsure of the consequences. I'm most concerned about https://github.com/fsprojects/fantomas-for-vs/issues/55.
- We should probably aim for a major version bump here and increase the TFM of the tool to `net8.0`.
- Should we reconsider anything else if we proceed with a major version bump?

Follow up:
- [x] https://github.com/dotnet/fsharp/issues/17720
- [x] https://github.com/dotnet/fsharp/issues/17721

//cc @dawedawe @josh-degraw

### DIFF
--- a/src/Fantomas.Core/ASTTransformer.fs
+++ b/src/Fantomas.Core/ASTTransformer.fs
@@ -2178,8 +2178,9 @@ let mkSynTypeConstraint (creationAide: CreationAide) (tc: SynTypeConstraint) : T
         TypeConstraintEnumOrDelegateNode(mkSynTypar tp, "delegate", List.map (mkType creationAide) ts, m)
         |> TypeConstraint.EnumOrDelegate
     | SynTypeConstraint.WhereSelfConstrained(t, _) -> mkType creationAide t |> TypeConstraint.WhereSelfConstrained
-    | SynTypeConstraint.WhereTyparNotSupportsNull(genericName, range) ->
-        failwithf "todo WhereTyparNotSupportsNull : %A" (genericName, range)
+    | SynTypeConstraint.WhereTyparNotSupportsNull(typar, EndRange 4 (mNull, m)) ->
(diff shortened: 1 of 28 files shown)


nojaf:
Not sure why, but the logger was no longer printing anything to the stdout.
No clue why, but using the static logger works.

src/Fantomas.Core/ASTTransformer.fs
```diff
@@ -78,6 +78,16 @@ let mkSynAccess (vis: SynAccess option) =
...
+let (|AccessSynValSigAccess|) (valSigAccess: SynValSigAccess) =
+    match valSigAccess with
+    | SynValSigAccess.GetSet(accessibility = vis)
+    | SynValSigAccess.Single(accessibility = vis) -> vis
+
+let (|SynValSigAccessAll|) (valSigAccess: SynValSigAccess) =
+    match valSigAccess with
+    | SynValSigAccess.Single(vis) -> vis, None, None
+    | SynValSigAccess.GetSet(vis, g, s) -> vis, g, s
+
```
dawedawe, line 90: Nit:
As they are so close to each other, I'd go with the same order in the APs: `GetSet`, then `Single`

dawedawe (approved):
In the name of all Fantomas users: Thank you!

Issue #3082, opened 2024-04-22
Non needed parentheses are added around lambda call from tuple/members


MangelMaxime:
[...]

PR #3119, opened 2024-09-14
Fixed unnecessary parentheses in last tuple item (fsprojects#3082)

johnW-ret:
1. fix for #3082 in CodePrinter.fs
2. 1 test in TupleTests.fs
3. Alter test to new _intended_ behavior in OperatorTests.fs

❗CHANGELOG.md change missing (I didn't push it because it was breaking my builds)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## 6.3.15 - 2024-09-14
+
+### Fixed
+* Non needed parentheses are added around lambda call from tuple/members [#3082](https://github.com/fsprojects/fantomas/issues/3082)
+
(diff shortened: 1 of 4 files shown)


dawedawe:
Looks good to me :)
Regarding the CHANGELOG.md, does it break your build if you just add this:
```
## [Unreleased]

### Fixed
* Non needed parentheses are added around lambda call from tuple/members [#3082](https://github.com/fsprojects/fantomas/issues/3082)
```

nojaf (approved):
Looks great! Thank you!  
I'm publishing this as a separate revision just in case it leads to invalid code in a scenario we haven't discovered yet.

PR #3123, opened 2024-09-16
V7.0

nojaf:
- Support F# 9
- Bump tfm to net8.0 for tool
- Release after dotnet 9 was released.
- Don't put a space before a prefix operator, unless we have to. (#3134)


dawedawe (approved):
Wonderful work :rocket:

PR #3127, opened 2024-09-20
Why operators are difficult to format

nojaf:
Done during https://amplifyingfsharp.io/sessions/2024/09/20/
I might close this in the future, unsure


nojaf:
@dawedawe to see minor degree curious how Fantomas would look like without any `atCurrentColumn` call.
Not sure if possible or even visually pleasing but could we at all cost maintain indent flow?

nojaf:
So, we had a [session](https://www.youtube.com/watch?v=PlqKbYF_QCE) on this.
And coding-wise I don't wish to pursue anything here.

Issue #3112, opened 2024-08-30
Project Update


nojaf:
Hi folks,

We want to give you an update on the latest regarding Fantomas. Personally, I believe Fantomas is in a good place. It's used by many developers on a daily basis, and it has come a long way. Over the years, we've made improvements in performance, accuracy (remember the disappearing code comments 🙈), and style (through the style guide process).

The current maintainers are @josh-degraw, @dawedawe, and myself. While we're happy to continue in this role, we must admit that life is taking us in different directions, and we're not as active as we used to be. There are no plans for any new features in Fantomas; we're more focused on reviewing pull requests as they come in.

That being said, this feels like a good moment to find some new blood. It's possible that our involvement may eventually phase out completely. So, it would be great to find some new people to whom we can pass on our knowledge and eventually hand over the torch. If you're interested, feel free to reach out here, on the F# Foundation Slack, or on Discord. (We don't want this project to fade away, but without any succession, that might someday happen.)

Feel free to share your thoughts in this issue. We'd be happy to answer any questions you might have.

Warm regards,

David, Josh & Florian

PR #3138, opened 2024-11-26
Remove MaxDotGetExpressionWidth

nojaf:
As we prepare for a major version update to support F# 9, I would like to propose the removal of the `MaxDotGetExpressionWidth` setting.

### Description
This setting controls the maximum width for which nested [SynExpr.DotGet](https://fsharp.github.io/fsharp-compiler-docs/reference/fsharp-compiler-syntax-synexpr.html#DotGet) expressions can remain on a single line.

### Key Reasons for Removal:
- Since the introduction of the [Chain AST node](https://github.com/fsprojects/fantomas/pull/2696), this setting has become somewhat awkward and inconsistent.
- I am unaware of any users who actively utilize this setting; most people do not even know of its existence.
- In hindsight, I regret adding this as a configurable option, as there was no significant demand for it. The community's focus was on resolving [the original problem](https://github.com/fsprojects/fantomas/issues/501), which has been addressed and remains so after the removal of this setting.
- This is one less setting to maintain, little things like this [keep the lights on a bit longer](https://github.com/fsprojects/fantomas/issues/3112).

@dawedawe, @josh-degraw & @Smaug123 (or @pleaseletmesearch) please let me know if you are on board with this.

### DIFF
--- a/docs/docs/end-users/Configuration.fsx
+++ b/docs/docs/end-users/Configuration.fsx
@@ -730,28 +730,6 @@ fsharp_max_function_binding_width = 10
 (*** include-output ***)
 
 (**
-<fantomas-setting green></fantomas-setting>
-### fsharp_max_dot_get_expression_width
-<copy-to-clipboard text="fsharp_max_dot_get_expression_width = 100"></copy-to-clipboard>
-
-Control the maximum width for which (nested) [SynExpr.DotGet](https://fsharp.github.io/fsharp-compiler-docs/reference/fsharp-compiler-syntax-synexpr.html#DotGet) expressions should be in one line.
-*)
-
-(*** hide ***)
-printfn
-    $"# Default\n{toEditorConfigName (nameof FormatConfig.Default.MaxDotGetExpressionWidth)} = {FormatConfig.Default.MaxDotGetExpressionWidth}"
-(*** include-output ***)
-
-formatCode
-    """ 
-   let job = JobBuilder.UsingJobData(jobDataMap).Create<WrapperJob>().Build()
-    """
-    """
-fsharp_max_dot_get_expression_width = 60
-    """
-(*** include-output ***)
-
-(**
 <fantomas-setting green gr></fantomas-setting>
 ### fsharp_multiline_bracket_style
 <copy-to-clipboard text="fsharp_multiline_bracket_style = stroustrup"></copy-to-clipboard>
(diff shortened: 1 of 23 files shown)


dawedawe (approved):

josh-degraw (approved):

JimAlexBerger:
After upgrading to fantomas 7.0.0 we can see this affecting our formatting.
It looks like this was changed from previously having a default value of 80 to now follow the max_line_length config value.

We found it very pleasing and readable having these on separate lines. 
I know its not fantomas job to cater to everyone personal preference, but we were wondering if there has been a change in the style guideline to now change the default away from 80 and this change was intended.

Here is a sample from our code base, among many, that were affected by this change.

Before:
```
module NormalizedPersistenceId =
        let replaceNationalCharacters (text: string) =
            text
                .Replace("æ", "ae")
                .Replace("ø", "oe")
                .Replace("å", "aa")
                .Replace("Æ", "AE")
                .Replace("Ø", "OE")
                .Replace("Å", "AA")
```

After (we have a bit of an increased max_line_length of 160):
```
module NormalizedPersistenceId =
    let replaceNationalCharacters (text: string) =
        text.Replace("æ", "ae").Replace("ø", "oe").Replace("å", "aa").Replace("Æ", "AE").Replace("Ø", "OE").Replace("Å", "AA")

```

nojaf:
> I know its not fantomas job to cater to everyone personal preference, but we were wondering if there has been a change in the style guideline to now change the default away from 80 and this change was intended.

This setting predates the style guide, so it was never compliant with the style guide to begin with.

The change in behavior is a logical consequence of the removal of this setting.

While this may not be the answer you're looking for, you could add `//` after the first `)` to enable multiline mode.

guythetechie:
@JimAlexBerger and others who miss the old setting: a workaround I'm using is the [simple lambda syntax](https://devblogs.microsoft.com/dotnet/announcing-fsharp-8/#_.property-shorthand-for-(fun-x-->-x.property)) introduced in F# 8. The documentation shows how to access instance properties, but looks like it also works with instance methods.

Here's an example:
```fsharp
let a =
    "abc"
    |> _.Replace("æ", "ae")
    |> _.Replace("ø", "oe")
    |> _.Replace("å", "aa")
    |> _.Replace("Æ", "AE")
```

It's not as clean as the old setting or @nojaf 's suggestion to use `//` to enable multiline mode. That said, it avoids having to explain to other maintainers that `//` is intentional and not accidentally left over from a previous comment.

PR #3149, opened 2025-01-14
Fix a link

psfinaki:
Dead link :) I guess this is what's supposed to be there...


nojaf:
Thanks!

psfinaki:
@nojaf also I noticed that the preview tool returns a server error - is this a known issue or temporal?
image

PR #3161, opened 2025-04-05
Print version in detailed verbosity

### LINKED ISSUE #3148: [Feature request] Print version at the beginning of the execution
I propose Fantomas prints its version at the very beginning of its execution.

The existing way to deal with this is to just separately ask for the tool version using `--version`.

## Pros and Cons 

The advantages of making this adjustment to Fantomas are:
- sometimes multiple versions of the tool are installed on the machine (e.g. locally and globally) and having the version printed decreases cognitive load in these situations
- this would also help maintainers since the screenshots with problems would also include the version number, reducing a need to communicate this
- this is common behavior for many dotnet tools (at least in my practice)

The disadvantages of making this adjustment to Fantomas are - cannot think of any.

## Extra information

Estimated cost (XS, S, M, L, XL, XXL): XS. Or XS/S if this is configurable.

Related suggestions: (put links to related suggestions here) - didn't find any.

## Affidavit (please submit!)

* [x] I have read the [Contribution Guidelines](https://github.com/fsprojects/fantomas/blob/main/CONTRIBUTING.md).

Please tick all that apply:
* [x] I or my company would be willing to help implement and/or test this

dawedawe:
should go in after #3160 

fixes #3148 

One could argue it should only be reported when there are format results but I think it makes sense to have it also with all the error cases like `InputPath.NoFSharpFile` etc.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Changed
 
-* Pring version at beginning of execution when detailed verbosity is set. [3148](https://github.com/fsprojects/fantomas/issues/3148)
+* Print version at beginning of execution when detailed verbosity is set. [3148](https://github.com/fsprojects/fantomas/issues/3148)
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Changed
+
+* Pring version at beginning of execution when detailed verbosity is set. [3148](https://github.com/fsprojects/fantomas/issues/3148)
+
 ### Fixed
 
 * Idempotency problem with comments in applications on lambda expressions. [#3128](https://github.com/fsprojects/fantomas/issues/3128)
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Changed
+
+* Pring version at beginning of execution when detailed verbosity is set. [3148](https://github.com/fsprojects/fantomas/issues/3148)
```
nojaf, line 7:
```suggestion
* Print version at beginning of execution when detailed verbosity is set. [3148](https://github.com/fsprojects/fantomas/issues/3148)
```

nojaf (approved):

nojaf:
Thanks!

src/Fantomas/Program.fs
```diff
@@ -462,9 +462,15 @@ Join our Discord community:      https://discord.gg/Cpq9vf8BJH
 
     let asyncRunner = Async.Parallel >> Async.RunSynchronously
 
-    if Option.isSome version then
+    let versionLog =
         let version = CodeFormatter.GetVersion()
-        stdlog $"Fantomas v%s{version}"
+        $"Fantomas v%s{version}"
+
+    if Option.isNone version then
```
nojaf, line 469: Nitpick but maybe go for a pattern match here.

dawedawe: Mmh, na, I think it reads nicer that way. Also with line 472.

psfinaki:
Thanks :)

PR #3164, opened 2025-05-19
Use FSharp.Core.Extended

nojaf:
Before

```
| Method | Mean     | Error    | StdDev   | Rank | Gen0     | Gen1     | Allocated |
|------- |---------:|---------:|---------:|-----:|---------:|---------:|----------:|
| Format | 60.39 ms | 1.102 ms | 1.312 ms |    1 | 888.8889 | 333.3333 | 155.13 MB |
```

After

```
| Method | Mean     | Error    | StdDev   | Rank | Gen0     | Gen1     | Allocated |
|------- |---------:|---------:|---------:|-----:|---------:|---------:|----------:|
| Format | 60.51 ms | 0.489 ms | 0.458 ms |    1 | 888.8889 | 333.3333 | 155.13 MB |
```

It didn't really do anything for this repo.
Maybe I missed something in setting this up, although we probably don't use any of the optimized APIs.

//cc @vzarytovskii

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -4,12 +4,13 @@
         <CentralPackageTransitivePinningEnabled>true</CentralPackageTransitivePinningEnabled>
     </PropertyGroup>
     <ItemGroup>
-        <PackageVersion Include="FSharp.Core" Version="8.0.100"/>
+        <PackageVersion Include="FSharp.Core" Version="9.0.100"/>
         <PackageVersion Include="System.Collections.Immutable" Version="8.0.0" />
         <PackageVersion Include="System.Diagnostics.DiagnosticSource" Version="8.0.1" />
         <PackageVersion Include="System.Memory" Version="4.6.0" />
         <PackageVersion Include="System.Runtime" Version="4.3.1" />
         <PackageVersion Include="FsLexYacc" Version="11.3.0" />
+        <PackageVersion Include="FSharp.Core.Extended" Version="1.0.4" />
         
         <PackageVersion Include="Ionide.KeepAChangelog.Tasks" Version="0.1.8"/>
         <PackageVersion Include="DotNet.ReproducibleBuilds" Version="1.1.1"/>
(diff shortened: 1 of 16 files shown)


vzarytovskii:
Yeah, fair. I will look at the fantomas codebase and see what it can benefit from.

Numpsy:
Does it need a ```open FSharp.Core.Extended.Collections``` to use the alternate functions?

> Yeah, fair. I will look at the fantomas codebase and see what it can benefit from.

A quick attempt at profiling suggests that a version of Seq.tryHead using ValueOption might shave off some Option allocations in a few places fwiw

vzarytovskii:
> Does it need a ```open FSharp.Core.Extended.Collections``` to use the alternate functions?
> 
> 
> 
> > Yeah, fair. I will look at the fantomas codebase and see what it can benefit from.
> 
> 
> 
> A quick attempt at profiling suggests that a version of Seq.tryHead using ValueOption might shave off some Option allocations in a few places fwiw

That's on my list, however on latest runtime, reference types which do not escape the stack frame can be lifted to be allocated on stack. So this might be an improvement even without that change.

Numpsy:
On the related thought, would the ```Array.choose |> Array.tryHead``` at https://github.com/fsprojects/fantomas/blob/d21ab681da1328407aa4d607e67e1174d6214629/src/Fantomas.Core/Trivia.fs#L161 work as ```Array.tryPick``` ?

nojaf:
> On the related thought, would the `Array.choose |> Array.tryHead` at
> 
> https://github.com/fsprojects/fantomas/blob/d21ab681da1328407aa4d607e67e1174d6214629/src/Fantomas.Core/Trivia.fs#L161
> 
> work as `Array.tryPick` ?

Probably, would accept a PR, but I doubt it will have much impact.

Numpsy:
Saves 0.37MB of memory allocations in the benchmark project according to BDN, which is admitedly small compared to 155MB overall

nojaf:
Going to close this, but would accept/consider a new PR with this idea.

PR #3170, opened 2025-07-04
Check if application and idx are actually multiline.

nojaf:
Even though a stylistic bugfix, I would go for a revision here.
Not sure yet.


dawedawe (approved):
I like this a lot.

nojaf:
Going to release this as a revision.
I don't think there will be many folks impacted really.

PR #3172, opened 2025-07-29
Downgrade FSharp.Core in FCS

Smaug123:
Fantomas.FCS currently depends on FSharp.Core 8.0.100, which is a really very high version of the dependency for such a small library!

I'm making this change because I'd like to consume Fantomas.FCS from the WoofWare.Expect snapshot testing library, which is currently targeting FSharp.Core [at a *much* lower version](https://github.com/Smaug123/WoofWare.Expect/blob/ad05a9c106bdd33cfeb3961364745218f001a317/WoofWare.Expect/WoofWare.Expect.fsproj#L42).


Smaug123:
Oh Lord this looks complicated. Why does this override propagate to the other consumers? I see no reason why it should; surely they should retain their previous version bounds from Directory.Packages.props. I don't understand NuGet's semantics *at all* here.

nojaf:
The question is then why use an ancient version of FSharp.Core in WoofWare.Expect?
I mean, I get the theory, but in practice, who is not able to run latest LTS of dotnet these days?

Smaug123:
This *is* more a preference than an actual requirement on my part. I'll just vendor the relevant code like Fantomas did, I guess, since it's not looking easy to do this here.

nojaf:
I'm open to lowering, but it needs an actual use-case for my taste.

PR #3173, opened 2025-07-30
Docs - Update FAKE fantomas command

tboby:
https://fsprojects.github.io/fantomas/docs/end-users/FormattingCheck.html#FAKE

Currently still has `--recurse`, and doesn't have `DotNet.exec` format that works with FAKE 5 or 6 as far as I can tell.

This just fixes that.

Sorry for not creating an issue first, but it's a simple docs change!


nojaf (approved):
Thank you!

PR #3177, opened 2025-08-04
Merge sarif files

### LINKED ISSUE #3175: Sarif file update
Image

GitHub no longer support multiple sarif files.

My first guess here would be to merge the current SARIF files into one.

nojaf:
Fixes #3175

### DIFF
--- a/sarif.fsx
+++ b/sarif.fsx
@@ -1,0 +1,176 @@
+module Sarif
+
+open System
+open System.IO
+open System.Text.Json
+open System.Text.Json.Serialization
+open System.Threading.Tasks
+
+[<CLIMutable>]
+type Text =
+    { [<field: JsonPropertyName("text")>]
+      text: string }
+
+[<CLIMutable>]
+type Region =
+    { [<field: JsonPropertyName("startLine")>]
+      startLine: int
+      [<field: JsonPropertyName("startColumn")>]
+      startColumn: int
+      [<field: JsonPropertyName("endLine")>]
+      endLine: int
+      [<field: JsonPropertyName("endColumn")>]
+      endColumn: int }
+
+[<CLIMutable>]
+type ArtifactLocation =
+    { [<field: JsonPropertyName("uri")>]
+      uri: string }
+
+[<CLIMutable>]
+type PhysicalLocation =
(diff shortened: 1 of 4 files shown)


nojaf:
@TheAngryByrd maybe "a version" of this merging process should be available somewhere in the analyzer SDK? Perhaps a sub command in the cli tool, I don't quite know where but this would be useful.

Alternatively, if the cli tool accepts multiple fsprojs and only creates a single output file the problem is solved as well.

TheAngryByrd:
Yeah I was going to suggest adding this to analyzers cli. —merge-sarif with a list/glob would be fine enough.

Numpsy:
As an aside, how close is all that Sarif merger code to being able to write the sarif files to begin with, so it's all self contained rather than using Sarif.Sdk?
Asking because I once borrowed a bunch of the Sarif code from the Analyzers SDK to do https://github.com/fsprojects/FSharpLint/pull/702 and I wonder if a completed version of that might find the merge code useful as well.

nojaf:
@Numpsy yes, there probably is some sarif synergy possible.
Though, I'm leaning a bit more towards the analyzers cli being able to accept multiple projects. That might be the more interesting fix for this problem.

sarif.fsx
```diff
@@ -0,0 +1,295 @@
...
+                    results.AddRange(run.results)
+
+                    match run.tool.driver.rules with
+                    | None -> ()
+                    | Some rulesList -> rules.AddRange(rulesList)
+
+
+            let combined: SarifLog =
+                {
+                  // I don't know why firstSarif.schema is null
```
dawedawe, line 274: Maybe something with the `$` 🤷‍♂️ 
But good enough.

dawedawe (approved):

sarif.fsx
nojaf: Looks fine now.

Numpsy:
> Though, I'm leaning a bit more towards the analyzers cli being able to accept multiple projects. That might be the more interesting fix for this problem.

Actually I think the current versions of FSharpLint can be pointed at a single solution or solution filter and will then just generate a single report containing all the projects in the solution (though that will always do everything, you can't do a whole solution and exclude one project)

Issue #2447, opened 2022-08-23
Building Fantomas in Visual Studio for dev errors


cgravill:
I'm getting an error while trying to build this project:

```
1>------ Rebuild All started: Project: Fantomas.FCS, Configuration: Debug Any CPU ------
Restored X:\dev\forks\fantomas\src\Fantomas.Core.Tests\Fantomas.Core.Tests.fsproj (in 150 ms).
Restored X:\dev\forks\fantomas\src\Fantomas.Benchmarks\Fantomas.Benchmarks.fsproj (in 160 ms).
Restored X:\dev\forks\fantomas\src\Fantomas.FCS\Fantomas.FCS.fsproj (in 53 ms).
Restored X:\dev\forks\fantomas\src\Fantomas\Fantomas.fsproj (in 241 ms).
Restored X:\dev\forks\fantomas\src\Fantomas.Client\Fantomas.Client.fsproj (in 243 ms).
Restored X:\dev\forks\fantomas\src\Fantomas.Tests\Fantomas.Tests.fsproj (in 254 ms).
1>C:\Users\xyz\.nuget\packages\ionide.keepachangelog.tasks\0.1.8\build\Ionide.KeepAChangelog.Tasks.targets(29,9): error MSB4018: The "Ionide.KeepAChangelog.Tasks.ParseChangeLogs" task failed unexpectedly.
1>C:\Users\xyz\.nuget\packages\ionide.keepachangelog.tasks\0.1.8\build\Ionide.KeepAChangelog.Tasks.targets(29,9): error MSB4018: System.IO.FileNotFoundException: Could not load file or assembly 'FSharp.Core, Version=4.4.3.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' or one of its dependencies. The system cannot find the file specified.
1>C:\Users\xyz\.nuget\packages\ionide.keepachangelog.tasks\0.1.8\build\Ionide.KeepAChangelog.Tasks.targets(29,9): error MSB4018: File name: 'FSharp.Core, Version=4.4.3.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'
1>C:\Users\xyz\.nuget\packages\ionide.keepachangelog.tasks\0.1.8\build\Ionide.KeepAChangelog.Tasks.targets(29,9): error MSB4018:    at Ionide.KeepAChangelog.Parser.parseChangeLog(FileInfo file)
[...]

PR #3182, opened 2025-09-04
Update Ionide.KeepAChangelog.Tasks to 0.3.0

Numpsy:
refs
https://github.com/fsprojects/fantomas/issues/2447
https://github.com/ionide/KeepAChangelog/issues/8

This gets it building inside Visual Studio 2022 for me, but:

The new parser does have a lot of errors when parsing the change log file though, so there are lots of changes to get it to accept the file:

- It wants brackets around all version numbers (previously there were some versions in the change log that had brackets and some that didn't)
- It wants list entries to use '-' rather than '*'
- It doesn't like 'Miscellaneous' or 'Documentation' as section names
- It wants various uses of '#' to be escaped.
- It wants all releases to have at least one subsection heading (Changed/Fixed etc)

This does make the diff rather large compared to just updating the KeepAChangeLog package.


CHANGELOG.md
```diff
@@ -1,2136 +1,2132 @@
...
-* Update FCS to 36. [#899](https://github.com/fsprojects/fantomas/pull/899), [#961](https://github.com/fsprojects/fantomas/pull/961)
-* Replaced json configuration with .editorconfig. [#650](https://github.com/fsprojects/fantomas/issues/650)
-* Sunset setting ReorderOpenDeclaration. [#645](https://github.com/fsprojects/fantomas/issues/645)
-* Sunset setting KeepNewlineAfter. [#737](https://github.com/fsprojects/fantomas/issues/737)
-* Renamed setting IndentSpaceNum to IndentSize. [#940](https://github.com/fsprojects/fantomas/pull/940)
-* Renamed setting PageWidth to MaxLineLength. [#940](https://github.com/fsprojects/fantomas/pull/940)
-* Removed all style configuration options from CLI tool. [#704](https://github.com/fsprojects/fantomas/issues/704)
-* Added [Benchmarks](https://fsprojects.github.io/fantomas/). [#867](https://github.com/fsprojects/fantomas/issues/867)
-* Split up setting SpaceBeforeArgument to multiple settings. [#649](https://github.com/fsprojects/fantomas/pull/649)
-  - SpaceBeforeParameter
```
Numpsy, line 1547: The new changelog parser doesn't seem to like the nested lists here - it complains about there not being a dash at the start of the line - not sure what syntax needs to be used to keep that happy, whilst still rendering the child lists nicely in the Github formatted markdown view

nojaf:
Thanks, it does feel like we lost a lot of flexibility.
Was it easy to figure all these changes out?
Would you mind remove the VS section in https://github.com/fsprojects/fantomas/blob/main/docs/docs/contributors/Index.md as well.

baronfel:
I'd view this long of a list as things that need to be fixed in KAC, personally. This is kinda draconian for what's supposed to be a library _helping_ you do the right thing.

baronfel:
Logged https://github.com/ionide/KeepAChangelog/issues/37 to track fixing most of these pain points.

nojaf:
Thanks, I'm a bit torn on this. I don't mind a bit more strictness, but it'd be nice if certain things remained as flexible as they were before this PR.

Numpsy:
> Was it easy to figure all these changes out?

The error messages said where the issues were (though it seems to abort after reporting the first error), and then most of the changes are just repetitions of the same thing and could be changed with find and replace

Numpsy:
> Would you mind remove the VS section in https://github.com/fsprojects/fantomas/blob/main/docs/docs/contributors/Index.md as well.

Done.
(and then the CI failed, but I don't know how the docs change would cause that test failure)

nojaf:
Thanks, I created #3184 for that build failure. It pops up from time to time.
Anything else missing in this PR?

Numpsy:
Previous comments about the fornatting of the changelog aside, the update does seem to be working.

nojaf (commented):
Works for me, thanks a bunch for this update @Numpsy!

PR #3191, opened 2025-10-14
Update Ionide.KeepAChangelog to 0.3.1

Numpsy:
This picks up a fix to a bug in the changelog parser that required '#' to be escaped, so all the escapes that were added in the update to 0.3.0 can be removed now.


dawedawe (approved):
Thanks!

nojaf:
Thanks!

PR #3195, opened 2025-10-28
Fantomas tool path safety checks

Thorium:
Potential security issue: 
Someone could do some remote code execution attack by injecting their own program as fantomas-tool. Then execute that on part of build process via whatever new codebase repository build the victim is executing.

This PR adds some extra checks to pick the fantomas-tool.
Better than nothing, but this is not going as far as e.g. strong-name checks.


nojaf:
I'm not sure I want this, do you have an actual use-case for this?

Thorium:
This is theoretical security issue. I don't know the best solution. I'm lucky enough to not have seen any real-life usage.

nojaf:
Don't want this.

PR #3196, opened 2025-10-28
EditorConfig settings are fetched many times, so cache them instead.

Thorium:
Performance improvement:
EditorConfig settings are fetched and parsed many times (per file), so cache them per directory instead.


nojaf:
I prefer a solution that is implemented over in https://github.com/editorconfig/editorconfig-core-net

Numpsy:
> I prefer a solution that is implemented over in https://github.com/editorconfig/editorconfig-core-net

Is this something different that the caching talked about in https://github.com/editorconfig/editorconfig-core-net/pull/23 ?

nojaf:
As mentioned, don't want this here, should be solved at the https://github.com/editorconfig/editorconfig-core-net level.

PR #3199, opened 2025-11-06
Clarify how Fantomas works.

nojaf:
This surfaced in a recent conversation.
I like the word document analogy.


dawedawe (approved):
Yepp, good analogy :+1:

PR #3207, opened 2025-12-12
Change default for MultilineBracketStyle

nojaf:
See discussion over at https://github.com/fsprojects/fantomas/issues/3200

We would change the default to `aligned` as it is safer for the time being.
If the parser improves, post v8 we could consider switching to stroustrup.

Would like merge this after releasing a first alpha in #3206


dawedawe (approved):
Let's see how that plays out.

PR #3208, opened 2025-12-26
Refactor tuple pattern edge-case in genBinding

nojaf:
This refactors genBinding to move the actual workaround we have for tuples.

Concretely, we have this for:

```fsharp
        let var1withAVeryLongLongLongLongLongLongName, var2withAVeryLongLongLongLongLongLongName = // foo
            someFunc 1, someFunc 2
```

we need to wrap `(var1, var2)` into parentheses to remain valid code.

But we were repeating a bit too much code in `genBinding` to deal with this.


dawedawe (approved):
nice!

PR #3215, opened 2026-02-24
Add scripts to debug writer events and full format.

nojaf:
Supersedes #3189


nojaf:
@dawedawe, it has been a while since I abused my admin powers here. Felt nostalgic.
Went along with it, happy to address any remarks you still might have.

Issue #2665, opened 2022-12-25
Please retain --stdin & --stdout options in the cli


kaashyapan:
I propose we ... retain --stdin & --stdout options in the cli

The existing way of Fantomas deals with this problem is ...It has been removed from the CLI. It works but does not respect editorconfig.

## Pros and Cons 

These have been removed from the documentation but it still appears to work.
It is a useful option to have when using from plugins for vim, emacs etc.. For e.g mhartington/formatter.nvim.
Although fantomas is part of LSP, I sometimes get a timeout and it is useful to call the formatter while bypassing calls to the LSP server.

The disadvantages of making this adjustment to Fantomas are ...
Cant think of any.
Work maybe required to add some tests. If it is too difficult to get at the editorconfig for the project, it can also be handled on the editor side. Maybe add tool argument --config "editconfig/filepath".

## Affidavit (please submit!)

Please tick this by placing a cross in the box:
* [x] I have read the [Contribution Guidelines](https://github.com/fsprojects/fantomas/blob/main/CONTRIBUTING.md).
* [x] I have [searched both open and closed suggestions on this site](https://github.com/fsprojects/fantomas/issues) and believe this is not a duplicate

Please tick all that apply:
* [x] This is not a breaking change to Fantomas
* [x] I or my company would be willing to help implement and/or test this
* [ ] This suggestion is part of the [Microsoft style guide](https://docs.microsoft.com/en-us/dotnet/fsharp/style-guide/formatting) (please add a link to section if so)
* [ ] This suggestion is part of the [G-Research style guide](https://github.com/G-Research/fsharp-formatting-conventions) (please add a link to section if so)

nojaf:
Hello, thank you for your interest in this project.

We removed `--stdin` and `--stdout` because we had no clue if anybody was using this in the first place, and due to the maintenance aspect. Quite some code was involved for this earlier (see https://github.com/fsprojects/fantomas/commit/2ca49d0854454fa3535eead10ca90fd0cb788521) and it never worked perfectly in combination with other flags (f.ex. `--check`).

Having `--stdin` in itself is insufficient to format the code properly. We need to know a file name as well to determine if the file is a signature file and compile the `.editorconfig` settings. The way `.editorconfig` works is that it resolves from the file point of view and not from the config file point of view. We never nailed this in the past, so the removal made a lot of sense.

> Although fantomas is part of LSP, I sometimes get a timeout and it is useful to call the formatter while bypassing calls to the LSP server.

[...]

PR #3371, opened 2026-06-19
Add verbosity levels for warnings and errors

tesujimath:
These behave the same as normal in terms of output formatting, but have a minimum log level of warning and error respectively.

This is useful when formatting to stdout with e.g. `--out /dev/fd/1` to avoid the `"<file> was formatted."`
message appearing in the formatted output.

(Being able to format to stdout seems to be essential for e.g. hooking Fantomas into Emacs Aphelia.)

Relates to #2665

### DIFF
--- a/src/Fantomas/Logging.fs
+++ b/src/Fantomas/Logging.fs
@@ -6,6 +6,8 @@ open Serilog
 type VerbosityLevel =
     | Normal
     | Detailed
+    | Warnings
+    | Errors
 
 let initLogger (level: VerbosityLevel) : VerbosityLevel =
     let logger =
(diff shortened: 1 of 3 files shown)


nojaf:
Hey Simon, thanks for the PR!

Before we go the `--out`/stdout route though, I'd like to back up a bit. What are you using to format on save? An LSP like fsautocomplete? Normally that's the path that calls Fantomas and sorts out `.editorconfig` and `.fsi` handling for you, so I'd really like to understand why that isn't doing the job for you here.

Formatting only needs the untyped tree, so it should never get stuck waiting on type-checking. If it does, that smells like something on the editor/LSP side rather than something `--out` ought to fix. Can you tell me a bit more about your setup and what's actually going wrong? I'd rather get that working than add a stdout workaround.

tesujimath:
Hi Florian,

Thanks for looking at this.

Sorry, I meant to add a comment here to say I'd solved my particular problem using apheleia, which is happy with this.

```
(with-eval-after-load 'apheleia
  (setf (alist-get 'fsharp-ts-mode apheleia-mode-alist) 'fantomas)
  ;; F# autoformat using fantomas
  (setf (alist-get 'fantomas apheleia-formatters) '("fantomas" input "--out" output)))
```

I previously overlooked using `output`.

So I have no need for this personally. Merge it if you think it's useful, otherwise no problem!

Cheers,
Simon

PR #3377, opened 2026-07-11
Add fsharp_keep_single_case_union_multiline setting

philipcraig:
### What
Adds an opt-in setting `fsharp_keep_single_case_union_multiline` (default `false`, so existing output is unchanged). When enabled, a single-case discriminated union declaration is kept on its own line instead of being collapsed onto the type name when it fits:

```fsharp
// default (false) — unchanged
type MyDU = Short of int

// fsharp_keep_single_case_union_multiline = true
type MyDU =
    | Short of int
```

### Why
The goal is to let users uphold the [G-Research F# formatting conventions](https://github.com/G-Research/fsharp-formatting-conventions#formatting-discriminated-union-declarations), which list the multiline layout as the **Preferred** form for a single-case DU:

```fsharp
// Preferred
type Foo =
    | Foo of int
```

Today Fantomas collapses a short single-case DU onto one line — even with `fsharp_bar_before_discriminated_union_declaration = true` it produces `type Foo = | Foo of int` — so that Preferred multiline form only appears once the declaration is long enough to wrap. This setting lets people following those conventions get the Preferred shape consistently, and (independently) gives teams the uniform "one case per line regardless of case count" style, which also keeps diffs smaller when a second case is later added.

It is opt-in and off by default, so nothing changes for anyone who doesn't set it. Related prior discussion: fsharp/fslang-design#762.

### How
- `FormatConfig`: new field `KeepSingleCaseUnionMultiline` (+ default). The editorconfig key is derived by reflection, so parsing/serialization are automatic.
- `CodePrinter`: in the single-case branch of `genTypeDefn` (`TypeDefn.Union`), use the existing multiline layout when the flag is set instead of `expressionFitsOnRestOfLine`. Signature files share this path.
- Docs section + CHANGELOG entry.

### Tested
- New `KeepSingleCaseUnionMultilineTests.fs` (8 tests: fields / no-field / access modifier / bar-combination / multi-case unaffected / member case / signature file / idempotency).
- Full `Fantomas.Core.Tests` green (2849 passed); changed files pass `fantomas --check`.

Happy to bikeshed the setting name.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Added
 
-- New `fsharp_keep_single_case_union_multiline` setting to keep single-case discriminated union declarations on a separate line instead of collapsing them onto the type name. [#3373](https://github.com/fsprojects/fantomas/pull/3373)
+- New `fsharp_keep_single_case_union_multiline` setting to keep single-case discriminated union declarations on a separate line instead of collapsing them onto the type name. [#3377](https://github.com/fsprojects/fantomas/pull/3377)
 
 ## [8.0.0-alpha-012] - 2026-04-16
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## [Unreleased]
+
+### Added
+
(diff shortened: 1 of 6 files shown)


nojaf:
This got resolved in the G-Research style guide repository and we are not going to move forward with this.

PR #3438, opened 2026-08-28
Add analyzer for unused opens

nojaf:
Ported from the diagnostic FsAutoComplete raises as FSAC0001. Detection is a
single call to FSharp.Compiler.EditorServices.UnusedOpens, which walks every
symbol use of the file and keeps the opens a name was written through, so the
rule reads check results and is quiet in the editor without them.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -27,19 +27,10 @@ The style rules for this repository are analyzers rather than prose, so the feed
 you work instead of in review. They live in `analyzers/`, and the `Analyze` and `AnalyzeChanged`
 pipelines run them alongside the two analyzer packages.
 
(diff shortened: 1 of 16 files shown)


auduchinok:
It's an interesting direction and is very appealing thing to do, but it seems very wrong to do any semantic analysis inside a formatter that is expected to work fast because it is syntax-only. Or is it not enabled by default?

nojaf:
Oh this does not happen when you run Fantomas as end-users.
This is something extra that runs during CI for any changes we make to the Fantomas Codebase. It does not leak into Fantomas.Core or anything.
This is purely a localized thing for my team of experts 😅.

